PR #160, opened 2026-04-07
fix(run): discover frontend port from running nx serve process

adrien-oper:
## Summary

- `_discover_frontend_port` falls back to scanning port 4200, which picks up another worktree's frontend when multiple worktrees run concurrently
- Now matches the `nx serve` process whose command line contains the current worktree path and extracts `--port=XXXX`
- Aligns with the dynamic port architecture from #141 — ports are discovered from running processes, never stored in files or DB

## Test plan

- [x] Verified with `t3 company run e2e-private` — correctly detected port 4206 for `ac-main-product-409-ticket` worktree among 15+ running `nx serve` processes
- [x] Ruff check + format pass
- [x] Pre-commit hooks pass
- [x] Pytest pass (pre-push hook)

### DIFF
--- a/src/teatree/core/management/commands/run.py
+++ b/src/teatree/core/management/commands/run.py
@@ -1,4 +1,5 @@
 import os
+import re
 import socket
 import subprocess  # noqa: S404
 import urllib.request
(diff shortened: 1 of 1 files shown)


adrien-oper:
Superseded by a cleaner approach: dockerizing the frontend in the overlay's `docker-compose.override.yml`. This makes `docker compose port frontend 4200` the single discovery mechanism for ALL services — no process scanning needed. The fix lives in the t3-oper overlay, not in teatree core.

PR #313, opened 2026-04-13
fix(run): discover frontend port from running nx serve process

souliane:
## Summary

- `_discover_frontend_port` was falling back to scanning port 4200, which picks up another worktree's frontend when multiple worktrees run concurrently.
- Now matches the `nx serve` process whose command line contains the current worktree path and extracts `--port=XXXX` from the args.
- Aligns with the dynamic port architecture from #141 — ports are never stored in files or DB; they're discovered from running processes.

Supersedes #312 (stale branch with broken CI trigger).

### DIFF
--- a/src/teatree/core/management/commands/run.py
+++ b/src/teatree/core/management/commands/run.py
@@ -1,4 +1,5 @@
 import os
+import re
 import socket
 import subprocess  # noqa: S404
 import urllib.request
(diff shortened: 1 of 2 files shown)


souliane:
Closing as obsolete. The underlying  function was removed from main — the multi-worktree port collision is now solved by using `find_free_ports()` directly in the `frontend` command. The fix in this PR is no longer applicable.

PR #316, opened 2026-04-15
fix: require prek install before first commit in new worktrees

souliane:
## Summary

- Require prek (pre-commit) to be installed in fresh worktrees before the first commit
- Remove hardcoded `pass show gitlab/pat` from the bulk-publish snippet in review/SKILL.md

## Test plan

- [x] Pre-commit hooks pass in fresh worktree after setup


souliane:
All changes in this branch are already merged into main — diff vs main is empty. Closing as superseded.

PR #459, opened 2026-04-25
test(coverage): combine unit + e2e suites into single threshold

### LINKED ISSUE #458: Coverage: combine unit + e2e suites into a single report (option 2)
## Background

After the bug-hunt 2026-04-25 (#455) we added a batch of Playwright e2e tests under `e2e/test_dashboard_fixes.py`. As we keep pushing test mass into e2e (per the AGENTS test-writing doctrine), unit tests start duplicating coverage — but we have no way to see it because each run measures coverage in isolation:

- `test (3.13)` job: runs `uv run pytest`, emits `coverage.xml` for `tests/` only.
- `e2e` job: runs `t3 teatree e2e project`, emits no coverage data at all.

So we can't currently answer "is this unit test still earning its keep?" — there's no combined report to diff against.

## Goal

Single combined coverage report covering both unit + e2e runs, with one `fail_under` gate.

## Approach (option 2 — two jobs, combined data)

1. **Unit job** — switch from `--cov-report=xml` to `coverage run -p` so the data file is `.coverage.<host>.<pid>.<random>`; upload as artifact.
2. **E2E job** —
   - Set `COVERAGE_PROCESS_START=$(pwd)/pyproject.toml` in the `t3 teatree e2e project` invocation.
   - Add `coverage.process_startup()` to a new `src/teatree/_coverage_init.py` and import it from `manage.py`/`asgi.py` behind an `if os.environ.get('COVERAGE_PROCESS_START'):` guard, so the uvicorn child process captures server-side coverage without touching the production startup path.
   - Make pytest itself run with `coverage run -p` so any in-process e2e coverage also lands in a `.coverage.*` file.
   - Upload `.coverage.*` as artifact.
3. **New `coverage-combine` job** — runs after both: downloads both artifacts, `coverage combine`, `coverage report --fail-under=93`, `coverage html`, uploads `htmlcov/`.
4. **Remove `fail_under` from the unit job** so only the combined report enforces it.

## Identifying redundant unit tests

With `--cov-context=test` enabled, `coverage report --contexts=*` shows which tests cover which lines. Workflow for pruning:
1. Pick a unit test.
2. Check whether all its covered lines are also covered by an e2e test.
3. If yes, delete it. Re-run combined; coverage stays the same → safe drop.
4. Done by hand for now; can be scripted later.

## Acceptance

- Combined report visible as a CI artifact (`htmlcov/`) on every PR.
- Single `fail_under=93` gate, enforced only on the combined report.
- Unit job no longer does its own threshold check.
- E2E coverage > 0 (proves the wiring works).
- No change to local `uv run pytest --no-cov` ergonomics.

## Out of scope

- Actually pruning redundant unit tests — that's a follow-up once the combined report is in place.

souliane:
## Summary
- Combines `tests/` (unit) and `e2e/` (Playwright) coverage so every line of `src/teatree` either suite exercises counts toward the 93% gate. This unblocks deleting unit tests that duplicate dashboard flows already covered by Playwright.
- Replaces the manual `coverage.process_startup()` shim with `[tool.coverage.run] patch = ["subprocess"], sigterm = true` (coverage 7.10+) — auto-instruments the uvicorn worker Playwright drives, flushes on `proc.terminate()`.
- Each pytest invocation gets a distinct `COVERAGE_FILE` (`.coverage.unit`, `.coverage.e2e`) so parallel writes don't race; `coverage combine` merges the partials.
- Individual jobs pass `--cov-fail-under=0` — the 93% target is enforced only on the combined report.

## Local
```sh
./scripts/coverage.sh
```
Runs `pytest tests/` and `t3 teatree e2e project --no-docker` in parallel, combines, reports, writes `htmlcov/index.html`.

## CI
- `test` and `e2e` jobs upload `.coverage.*` artifacts.
- A small `coverage` job downloads, combines, reports — **no test re-runs**.

## Test plan
- [x] `uv run ruff check` clean
- [x] Unit suite alone produces `.coverage.unit` (93.41%)
- [x] E2E run produces `.coverage.e2e.<host>.<pid>.<rand>` partials (subprocess instrumentation working)
- [x] `coverage combine` merges multiple files into one
- [ ] CI green: `test`, `e2e`, `coverage` all pass; combined report > 93%

Closes #458

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,23 +70,28 @@ jobs:
           tags: teatree-test-${{ matrix.python-version }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+      # --cov-fail-under=0 disables the per-run threshold here — the 93% gate
+      # runs on the combined unit+e2e report in the `coverage` job below.
       - run: >
           docker run --rm
           -v "$PWD":/app
           -e UV_PROJECT_ENVIRONMENT=/tmp/.venv
-          -e COVERAGE_FILE=/app/.coverage
+          -e COVERAGE_FILE=/app/.coverage.unit
           teatree-test-${{ matrix.python-version }}
           uv run -p ${{ matrix.python-version }} pytest --no-header -q
           -o cache_dir=/tmp/.pytest_cache
-          --cov-report=xml:/app/coverage.xml
+          --cov-fail-under=0
       - uses: actions/upload-artifact@v7
         if: always()
         with:
-          name: coverage-${{ matrix.python-version }}
-          path: coverage.xml
+          name: coverage-data-unit
+          path: .coverage.unit*
+          include-hidden-files: true
 
   e2e:
     runs-on: ubuntu-latest
+    env:
+      COVERAGE_FILE: .coverage.e2e
     steps:
       - uses: actions/checkout@v6
       - uses: astral-sh/setup-uv@v7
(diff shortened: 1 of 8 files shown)


souliane:
Closing — going back to unit-only coverage. Subprocess instrumentation slowed the uvicorn worker enough that HTMX/SSE-driven e2e tests went flaky in CI even with 60s timeouts, and the design tradeoffs to fix that (disable SSE in e2e, or run e2e twice) weren't worth the complexity over the existing setup. Combined coverage is parked.

Issue #777, opened 2026-05-16
feat: sanctioned per-invocation-approved fresh-DEV-dump db refresh path


souliane:
## Problem

There is no sanctioned, agent-usable command to pull a fresh dump from the DEV environment for the correct tenant and restore it. The only working path today is a manual `pg_dump` followed by `t3 <overlay> db refresh --dump-path <file>` — a hand-rolled work-around, not a command.

Root causes (verified against HEAD):

- The overlay `db_import` hardcodes `allow_remote_dump=False` (overlay `environment.py`), so the remote-dump strategy is unreachable from `db refresh`.
- teatree-core `utils/django_db.py` additionally gates the remote `pg_dump` strategy behind the env var `T3_ALLOW_REMOTE_DUMP=1` with an "Agents must never set this env var" comment. This is a blanket prohibition anti-pattern: it makes the path permanently dead instead of safely gated.

The supporting pieces already work: `dump_path` threads cleanly into the config, `db_restore` auto-detects `pg_dump -Fc` archives and restores via `pg_restore`, and the explicit-dump-path branch bypasses the slow-import/DSLR auto-discovery.

## Proposal

Replace the blanket prohibition with a **per-invocation explicit-approval gate**:

- Add an opt-in `--fresh-dump` path to `t3 <overlay> db refresh` that pulls a fresh dump from the DEV env for the tenant resolved from the worktree/overlay context (never hardcoded), into the worktree `.data/` with an unambiguous fresh filename, then restores it via the deterministic `--dump-path` -> `db_restore` path.
- Require **explicit interactive user approval on every invocation**: a confirmation that states environment + tenant + target DB and refuses unless the user confirms in that run. The gate is designed so an agent cannot self-approve — it must read an interactive confirmation, and in non-interactive/agent context (no TTY) it refuses with a clear message that a human must run/approve it.
- Remove the `T3_ALLOW_REMOTE_DUMP` env-var gate and the `allow_remote_dump=False` hardcode; the approval gate is the safety mechanism.

Cross-repo (lockstep): the extension-point change (`OverlayBase.db_import` contract + core `db refresh` + the approval helper in `utils/`) is teatree; the overlay wiring is the sibling overlay project. One cohesive feature mirrored across both per the extension-point lockstep rule.

## Why now

This unblocks a downstream ticket that currently requires the manual `pg_dump` hack. Once landed, the fresh-dump runs through this command with the user's per-invocation approval.

## Acceptance

[...]

PR #783, opened 2026-05-16
feat(db): sanctioned per-invocation-approved fresh remote dump path (#777)

adrien-oper:
feat(db): sanctioned per-invocation-approved fresh remote dump path (#777)

Replaces the blanket remote-dump prohibition with an explicit
per-invocation interactive approval gate. Before: the overlay
hardcoded allow_remote_dump=False and core gated the remote pg_dump
behind a T3_ALLOW_REMOTE_DUMP=1 env var with an "agents must never set
this" comment — the path was permanently dead, so the only way to pull
a fresh DEV dump was a manual pg_dump work-around.

- New teatree.utils.approval.require_interactive_approval: refuses
  unless both stdin and stdout are real TTYs and the human types yes.
  An unattended agent has no TTY, so it cannot self-approve — this is
  the safety mechanism, not a blanket ban.
- t3 <overlay> db refresh --fresh-dump: states env + tenant + target
  DB and requires that approval every run; threads approve_remote_dump
  through the OverlayBase.db_import extension-point contract into
  django_db_import.
- Removed the T3_ALLOW_REMOTE_DUMP env gate inside _try_fetch_remote_dump
  (the approval flag, set only post-approval, is now the gate). The
  command-blocker hook still rejects the defunct env var and points at
  the sanctioned --fresh-dump path.
- Integrated the dump-mechanics hardening from the user's known-good
  import_db_from_dev_env.sh: pg_dump now uses --no-owner
  --no-privileges so the dump is portable across the remote->local
  superuser boundary and the existing ownership-reassignment post-steps
  take over cleanly. Restore stays the deterministic db_restore path.
- Tenant is resolved from worktree/overlay context, never hardcoded.
- BLUEPRINT §14 + overlay-api docs updated to the new contract; no
  credentials or connection strings embedded or echoed anywhere.

TDD anti-vacuous: breaking the TTY guard flips the agent-cannot-self-
approve test red (DID NOT RAISE); reverting the pg_dump hardening flips
the no-owner test red. Full suite 3643 green, zero regression;
ruff/ty/tach clean. Cross-repo: the overlay-side wiring is a sibling MR.

### DIFF
--- a/BLUEPRINT.md
+++ b/BLUEPRINT.md
@@ -1480,9 +1480,9 @@ django_db_import(cfg, skip_dslr=False, allow_remote_dump=False)
 ```
 
 - `skip_dslr=True` — skip DSLR snapshots (used with `--force` to get a fresh dump)
-- `allow_remote_dump=True` — enable the remote pg_dump strategy (requires explicit user approval)
+- `allow_remote_dump=True` — enable the remote pg_dump strategy. Set **only** after a successful per-invocation interactive approval (#777): `t3 <overlay> db refresh --fresh-dump` calls `teatree.utils.approval.require_interactive_approval`, which states the env + tenant + target DB and refuses unless a human confirms at a TTY. An unattended agent has no TTY and so cannot self-approve — this replaces the old blanket `T3_ALLOW_REMOTE_DUMP` env-var prohibition (a defunct, now-blocked bypass).
 
-**Overlay responsibility:** Provide the config values and decide when to set `allow_remote_dump=True` (typically gated behind `--force` or an interactive prompt).
+**Overlay responsibility:** Provide the config values and forward `approve_remote_dump` (the post-approval flag from core's `db refresh --fresh-dump` gate) into `django_db_import(..., allow_remote_dump=...)`. The overlay must not hardcode the value or invent its own gate — the single sanctioned gate is core's interactive approval.
 
 ### 14.6 DSLR Integration
 
--- a/docs/overlay-api.md
+++ b/docs/overlay-api.md
@@ -105,9 +105,11 @@ Extra environment variables to set for a worktree. Defaults to `{}`.
 
 How to import/restore a database for this worktree. Returns `None` if no DB import is needed.
 
-#### `db_import(worktree, *, force, slow_import, dslr_snapshot, dump_path) -> bool`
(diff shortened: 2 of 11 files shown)


souliane:
## Cross-overlay heads-up (architectural review, pre-merge)

This PR adds `approve_remote_dump: bool = False` to the
`OverlayBase.db_import` extension-point contract and changes the core call
site to pass it through:

```python
success = overlay.db_import(
    worktree,
    ...
    approve_remote_dump=fresh_dump,
)
```

`OverlayBase.db_import` and the in-repo test overlays are updated, and
`docs/overlay-api.md` documents the new "overlay responsibility". But a
registered out-of-tree overlay that overrides `db_import` with a fixed
keyword-only signature and **no** `**kwargs` will raise
`TypeError: db_import() got an unexpected keyword argument 'approve_remote_dump'`
the first time core invokes `db refresh --fresh-dump` against it.

This is not theoretical: the contract is *already* informally divergent —
core today calls `overlay.db_import(worktree, force=force, ...)` even though
the base signature has no `force` param, so overlays already maintain a
*superset* signature by hand. This PR widens the core→overlay call again
without a mechanism that guarantees every registered overlay accepts the
new kwarg.

### Suggested options (author's choice)

1. **Tolerant call site**: have core introspect / `**kwargs`-guard the
   overlay call, or only pass `approve_remote_dump` when the overlay's
   signature accepts it, so an overlay that hasn't adopted the kwarg keeps
   working (and simply doesn't get the fresh-dump capability until it
   opts in).
2. **Coordinated rollout**: land the new kwarg as accepted-but-ignored in
   the base, ship the call-site change in a follow-up once registered
   overlays have widened their `db_import` signature.

Either way, the structural fix is the same one that surfaced separately:
the base↔overlay `db_import` signature contract should be made mechanically
verified (an import-time / test-time check that each registered overlay's
`db_import` accepts the full documented kwarg set) so this class of drift
fails loudly at load time instead of silently at `db refresh` time.

### Process note (per the every-fix-needs-a-regression-test rule)

The PR adds good tests for the *approval gate* itself
(`test_approval.py`, the non-interactive-refusal cases). It does **not**
add a test that asserts a registered-overlay `db_import` *accepts* the new
kwarg — which is the exact regression this heads-up is about. A small
contract test over all loaded overlays would close the loop.

(No code changes proposed here — this is a read-only architectural review
flag so the maker can decide before merge.)

Issue #1044, opened 2026-05-19
feat(loop): SlackDmInboundScanner — also scan thread replies, not just top-level DMs


souliane:
## Context
The fresh `SlackDmInboundScanner` (#1039 / #1014) only polls top-level messages on each overlay-bot DM channel. The user routinely replies *in threads* on bot-posted messages — those replies are not picked up today, so they silently disappear.

## Ask
Extend the scanner so that for each new top-level message the bot has posted on the DM channel, it polls `conversations.replies` and treats every non-bot reply with `parent_user_id == bot_id` (i.e. on the bot's own thread) as an inbound user message. Stamp the parent ts on the `PendingChatInjection` row so the agent's drain can give the agent the surrounding bot message as context.

## Acceptance
- User reply to ANY bot-DM thread lands as a `PendingChatInjection` row.
- Idempotency key remains `(overlay, ts)` so over-polling is safe.
- Regression test with a synthetic `conversations.replies` payload.

souliane:
Fixed on `main` — shipped in #1046 (commit `59fd0de3`, `fix(backends/slack): stamp channel + scan threads on REST DM polls (#1043)`). The merge subject referenced #1043 only, so this issue's auto-close keyword never fired (stale-open, not unfixed).

Verification against the acceptance criteria, all met:

- **User reply to ANY bot-DM thread lands as a `PendingChatInjection` row** — `SlackMessagingBackend.fetch_dms()` (`src/teatree/backends/slack_bot.py:208`) now calls `_fetch_thread_replies()` (line 268) → `conversations.replies` for each bot-posted top-level message; it skips the parent `ts` and any bot-authored reply (`reply.get('ts') == thread_ts or _is_bot_authored(reply, bot_id)`), so only non-bot replies on the bot's own threads are returned. The docstring at line 215 literally states "`conversations.replies` so thread replies are picked up (#1044)". `SlackDmInboundScanner.scan()` consumes these events generically and records each via `PendingChatInjection.record(...)`, with the channel stamped on the row.
- **Idempotency key remains `(overlay, ts)`** — unchanged; `PendingChatInjection.record` is keyed on `(overlay, ts)`, so re-polling the same thread is safe (the scanner skips the `None` dedup return).
- **Regression test with a synthetic `conversations.replies` payload** — `tests/teatree_backends/test_messaging.py` (+111 lines in #1046) covers the thread-scan / drain path.

Closing as completed.

PR #1046, opened 2026-05-19
fix(backends/slack): stamp channel + scan threads on REST DM polls (#1043)

### LINKED ISSUE #1043: fix(loop): SlackDmInboundScanner drops every event — `channel` missing on conversations.history payload
## Symptom
After #1039 (`feat(loop): Slack DM inbound scanner + PendingChatInjection drain`) merged on 2026-05-19, zero `PendingChatInjection` rows are ever inserted. User DMs to both overlay bots produce no inbound side effect, the way they did before the scanner shipped.

## Root cause
`SlackBotBackend.fetch_dms` polls `conversations.history` and returns the events as-is. Slack's `conversations.history` payload does *not* include a `channel` field on each message — the channel id is the request parameter, not part of the response.

`SlackDmInboundScanner.scan` extracts `channel = _event_channel(event)` (always `""`), then calls `PendingChatInjection.record(channel=channel, …)`. That helper's guard

```python
if not slack_ts or not channel or not text.strip():
    return None
```

bails out on every call, so no row is ever written.

## Repro
```python
from teatree.core.backend_factory import iter_overlay_backends
for ov in iter_overlay_backends():
    if ov.messaging is None: continue
    for e in ov.messaging.fetch_dms():
        print('channel:', e.get('channel'))  # always '' for REST-fallback path
```

## Fix
`SlackBotBackend.fetch_dms` (the REST fallback path — the Socket Mode receiver path is fine because Slack stamps `channel` on socket events) should stamp the resolved DM `channel` on each event before returning. One-line:

```python
for m in messages:
    if isinstance(m, dict):
        m.setdefault('channel', channel)
```

Add a regression test that pre-loads the queue with a `conversations.history`-shaped event (no `channel` key) and asserts `record()` succeeds.

## Impact
The bridge merged at `0c2b8ed75d53ab8c304b2bd07a0816bcda8ee84e` is a no-op end-to-end. Bot replies via `notify_user` work; user → agent direction is broken. This was hidden because the scanner's pre-flight 0019 migration was also unapplied on the live DB — once the migration was applied, the bug surfaced.

souliane:
## Why

After PR #1039 (`feat(loop): Slack DM inbound scanner + PendingChatInjection drain (epic #1014)`) merged this morning at `0c2b8ed7`, the inbound bridge was a silent no-op end-to-end. User DMs to both overlay bots produced zero `PendingChatInjection` rows and zero session-side injection.

Two regressions accumulated:

1. *Migration `0019_pendingchatinjection` was never applied on the live DB* — applied manually during diagnosis. Not codified here (the loop's normal setup applies migrations automatically) but worth flagging.
2. *`SlackBotBackend.fetch_dms` did not stamp `channel` on REST-polled events.* Slack's `conversations.history` returns each message *without* a `channel` field — channel is the request parameter, not the response. The scanner extracted `channel = ""` from every event; `PendingChatInjection.record` then bailed via its `if not channel` guard. No row, no signal, no inbound to the agent.
3. *Thread replies were not polled at all.* The user routinely replies in threads on bot-posted DMs; those replies landed nowhere.

This PR closes #1043 and addresses #1044.

## What

- `SlackBotBackend.fetch_dms` now stamps the resolved DM channel on each event before returning (`msg.setdefault("channel", channel)` — preserves Socket Mode payloads that already carry the field).
- The same path now also polls `conversations.replies` for every top-level bot message and surfaces non-bot replies as part of the same DM stream. Idempotency on the downstream side is unchanged: `unique(overlay, slack_ts)`.
- Refactored into `_poll_dm_history` / `_collect_user_dms` / `_fetch_thread_replies` so the public method stays under the complexity gate.
- Three regression tests:
  - `test_slack_fetch_dms_stamps_channel_on_each_event` — bare `conversations.history` payload without `channel`.
  - `test_slack_fetch_dms_preserves_existing_channel` — Socket Mode events that already have `channel` are not overwritten.
  - `test_slack_fetch_dms_includes_thread_replies` — synthetic `conversations.replies` reply lands in the DM stream.

## Verification

Local backfill against the canonical DB after applying the fix recovered all 14 unanswered messages from today (10 t3-teatree thread replies + 3 t3-company messages + 1 stale `hello`). User DM sent via the recovery channel as `slack-dm-recovery-20260519`.

`uv run pytest tests/teatree_backends/test_messaging.py --no-cov -q` — 38 passed in 0.24s

## Out of scope

- Setting up an automatic migration run on `t3 loop start` so unapplied migrations can't break a future bridge feature again — file separately if needed.
- The two-overlay duplication for the shared teatree-bot DM (one row per `(t3-teatree, ts)` *and* `(teatree, ts)`) — the drain handler dedupes by content, but the duplicate row is wasted work. Worth a follow-up.

## Test plan

- [x] `uv run pytest tests/teatree_backends/test_messaging.py` — 38 passed.
- [x] `uv run pytest tests/teatree_backends/ tests/teatree_loop/test_slack_dm_inbound.py tests/teatree_loop/test_scanners.py tests/test_pending_chat_injection_hook.py` — 315 passed.
- [x] `uv run ruff check` clean.
- [x] Live backfill against canonical DB recovered today's missing messages.

### DIFF
--- a/src/teatree/backends/slack_bot.py
+++ b/src/teatree/backends/slack_bot.py
@@ -19,6 +19,15 @@ from teatree.types import RawAPIDict
 type SlackPayload = dict[str, object]
 
 
+def _is_bot_authored(msg: RawAPIDict, bot_id: str) -> bool:
+    return msg.get("user") == bot_id or msg.get("bot_id") == bot_id
+
+
+def _is_thread_root(msg: RawAPIDict) -> bool:
+    thread_ts = msg.get("thread_ts")
+    return isinstance(thread_ts, str) and bool(thread_ts) and thread_ts == msg.get("ts")
+
+
 class SlackBotBackend:
     """MessagingBackend backed by a Slack bot token.
 
@@ -96,12 +105,22 @@ class SlackBotBackend:
         return events
 
     def fetch_dms(self, *, since: str = "") -> list[RawAPIDict]:
-        """Return new DMs from the user.
+        """Return new DMs from the user, including thread replies.
 
         Drains the Socket Mode queue first (populated by a running receiver).
-        When the queue is empty, falls back to polling ``conversations.history``
-        on the bot's DM channel with the configured user. Only messages FROM
-        the user are returned (bot's own messages are filtered out).
+        When the queue is empty, falls back to polling
+        ``conversations.history`` on the bot's DM channel with the
+        configured user, then for every top-level bot message also polls
+        ``conversations.replies`` so thread replies are picked up (#1044).
+
+        Slack's ``conversations.history`` and ``conversations.replies``
+        do not stamp the ``channel`` field on each message — it is the
+        request parameter, not part of the response. We stamp it here so
+        downstream consumers (``SlackDmInboundScanner`` →
+        ``PendingChatInjection.record``) have it (#1043).
+
+        Only messages FROM the user are returned (bot's own messages are
+        filtered out).
         """
         if self._dms:
             events, self._dms = self._dms, []
@@ -111,6 +130,12 @@ class SlackBotBackend:
         channel = self.open_dm(self._user_id)
         if not channel:
             return []
+        messages = self._poll_dm_history(channel=channel, since=since)
+        bot_id = self._resolve_bot_id()
+        return self._collect_user_dms(channel=channel, messages=messages, bot_id=bot_id)
+
+    def _poll_dm_history(self, *, channel: str, since: str) -> list[RawAPIDict]:
+        """Return the ``conversations.history`` messages list (or empty)."""
         params: dict[str, str | int] = {"channel": channel, "limit": 20}
         if since:
             params["oldest"] = since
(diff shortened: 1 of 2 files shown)


souliane:
Independent reviewer pass — verdict: **APPROVE**.

CI: 5/5 green (`lint`, `docs-drift`, `uv-audit`, `sbom`, `test (3.13)`) at head `aa1aa1ca`.

Correctness:

- Channel stamping fix is correct: `msg.setdefault("channel", channel)` matches the [`#1043`](https://github.com/souliane/teatree/issues/1043) repro and the protocol on `PendingChatInjection.record` (the `if not slack_ts or not channel or not text.strip()` guard now sees a populated channel).
- Thread reply polling fix is correct: `_is_thread_root` correctly identifies thread roots (`thread_ts == ts`); `_fetch_thread_replies` filters out the root itself and any bot-authored replies on `parent_user_id`, matching [`#1044`](https://github.com/souliane/teatree/issues/1044) acceptance.
- Migration `0019_pendingchatinjection` properly depends on `0018_deferred_question`, single leaf — no deeper migration-order bug. Implementer's claim that the live-DB unapply is an ops issue (not a code issue) checks out.
- TOCTOU between scanner and `PendingChatInjection.record`: the `unique(overlay, slack_ts)` constraint + `get_or_create` makes it idempotent at the DB level. Safe for over-polling and for concurrent scanner ticks.

Regression tests (anti-vacuous check):

- `test_slack_fetch_dms_stamps_channel_on_each_event` — would RED on pre-fix (`dms[0]["channel"]` raises `KeyError`). Anti-vacuous.
- `test_slack_fetch_dms_includes_thread_replies` — would RED on pre-fix (old code only saw the bot's top-level post and filtered it; `dms == []`). Anti-vacuous.
- `test_slack_fetch_dms_preserves_existing_channel` — would NOT fail on pre-fix (no stamping at all), so vacuous against the channel-missing bug, but it does guard the `setdefault` vs `=` distinction. Acceptable as a contract test.

Nit (non-blocking): `_fetch_thread_replies` does not pass `oldest` / `since` to `conversations.replies`, so on a long-running thread we re-fetch up to 50 replies every tick. `PendingChatInjection`'s unique constraint absorbs the duplicates, so it's just wasted API calls. Worth a follow-up if API rate becomes an issue.

Merging via §17.4 keystone path.

Issue #1210, opened 2026-05-20
Add `t3 setup slack-user-token` to re-scope personal Slack OAuth token


souliane:
Parent epic: #1209 (read/write decoupling of Slack tokens).

## Why

The personal Slack OAuth token at `pass slack/user-oauth-token` is missing several scopes that are required for posting in Slack-Connect channels under the user's identity (the only path that works for `mcp_externally_shared_channel_restricted` channels). Today the token has:

```
identify, channels:history, groups:history, im:history, mpim:history, files:read,
canvases:read, users:read, users:read.email, chat:write, canvases:write,
search:read.public, search:read.private, search:read.mpim, search:read.im,
search:read.files, search:read.users
```

It is missing at minimum:

- `reactions:write`
- `chat:write.public` (post in channels without joining)
- `chat:write.customize` (post with custom username/icon)

Without these the `chat.postMessage` / `reactions.add` work in #1209 will surface `missing_scope` instead of failing on `mcp_externally_shared_channel_restricted`.

## Scope of this issue (single focused PR)

1. Add a new CLI command `t3 setup slack-user-token [--reset]` that walks the user through Slack OAuth for the **personal** (xoxp) token only.
2. The required scopes list is a constant in the command module; it includes the union of the current scopes plus `reactions:write`, `chat:write.public`, `chat:write.customize` so a re-auth does not lose existing capability.
3. The OAuth flow mirrors `t3 setup slack-bot`'s pattern: build authorize URL, print it / open browser, accept the redirect via a local HTTP listener (or copy-paste fallback), exchange code for token via `oauth.v2.access`, verify scopes via `auth.test` headers, store via `pass insert slack/user-oauth-token`.
4. **Out of scope here:** splitting into `user-read-token` / `user-write-token`. That belongs to #1209.

## Acceptance

- `t3 setup slack-user-token` re-authorizes the token and stores it via `pass`.
- After running it, `pass show slack/user-oauth-token` returns an `xoxp-…` token whose `auth.test` `x-oauth-scopes` header includes all requested scopes.
- Tests under `tests/cli_setup/` cover the happy path, the missing-scope failure path, and `--reset` overwrite behaviour.

souliane:
Closing as part of backlog consolidation: already implemented — the setup command for the user token exists with reset, walks OAuth, verifies scopes via the auth.test x-oauth-scopes header, stores via pass, and the user scope set already includes reactions.

Issue #1209, opened 2026-05-20
Architecture: decouple Slack READ from Slack WRITE in the messaging layer


souliane:
## Background

the Slack messaging layer must completely decouple reading from writing. The current architecture conflates the two responsibilities behind a single token/identity, which is the root cause of the Slack Connect / bot-restricted-channel failures seen this session.

The failures aren't a bug in any one path — they're a coupling failure. The current messaging layer treats Slack as one transport with one token, when in fact reading and writing have completely different identity / scope / channel requirements.

## Current shape (coupled)

A single MCP wrapper handles both read and write using the **bot token** (`xoxb-***`). Concrete failure modes this session:

- `slack_add_reaction` on Connect channel `#review-channel` → `mcp_externally_shared_channel_restricted` (the bot is not a member of the external workspace).
- `slack_send_message` thread reply to the same channel → same refusal.
- `notify_send` (bot) silently returned `rc=1 did not deliver` against DMs that landed only after a fresh-session token refresh.

Workarounds attempted in-session:

- Fall back to the user's **personal token** (`pass slack/user-oauth-token`, `xoxp-***`) for Connect-channel writes. **But** the user token's scope list is `identify, channels:history, groups:history, im:history, mpim:history, files:read, canvases:read, users:read, users:read.email, chat:write, canvases:write, search:read.*` — it has NO `reactions:write`, NO `chat:write.customize`, NO `chat:write.public`. So the workaround only works for the small subset of writes the user token happens to be scoped for, and breaks silently for everything else.

## Proposed shape (decoupled)

Three responsibilities, three explicit identities:

| Role | Identity | Token type | Scope set | Lives in |
|---|---|---|---|---|
| **Reader** | personal | `xoxp-*` user OAuth | `*:history` + `search:read.*` + `users:read*` | `pass slack/user-oauth-token` |
| **Writer (private + DM)** | bot | `xoxb-*` | `chat:write`, `reactions:write`, `files:write` | bot install in company-slack.slack.com |
| **Writer (Connect + external)** | personal-write | `xoxp-*` user OAuth (separate token or re-scoped) | `chat:write`, `reactions:write`, `chat:write.public` | `pass slack/user-write-token` (new) |

The messaging layer (`src/teatree/core/notify.py` + a sibling reader module) auto-routes:

1. **READ paths** — `slack_read_channel`, `slack_read_thread`, `fetch_dms`, `fetch_reactions` — always use **Reader**. No write fallback.
[...]

PR #1190, opened 2026-05-20
feat(skills): add scanning-news skill for daily AI-newsletter improvement scan

souliane:
## Summary
- New t3 skill `scanning-news` that fetches today's TLDR AI + The Rundown AI editions, deep-reads relevant items, asks "can this improve t3?", and files `from-news-scan`-labelled issues with a Slack DM summary.
- Anthropic-style frontmatter (gerund name, trigger phrases, version, `last_research_date`). Non-negotiables front-loaded for compaction survival. Reference files split: `fetch-prompts`, `ticket-template`, `slack-format`, `cron-fallback`.
- Date-of-edition verification + aggregator-page detection are mandatory gates (closes a real failure mode where `tldr.tech/<track>/<date>` returns yesterday's edition or a kitchen-sink aggregator).

## Follow-up
- Loop wiring is intentionally deferred: a dedicated `scanning_news` scanner (mirroring `architectural_review`) needs to land before cron is removed. The skill documents this and references a cron fallback in `references/cron-fallback.md`.

## Test plan
- [ ] Pre-commit gates: all green locally (`prek run --files skills/scanning-news/SKILL.md skills/scanning-news/references/*.md README.md`)
- [ ] Pre-push test matrix passes (Docker)
- [ ] CI green
- [ ] Manual run of `/t3:scanning-news` against today's TLDR AI + The Rundown AI editions produces a valid DM and zero false-positive tickets

PR #1220, opened 2026-05-20
feat(cli): add `t3 setup slack-user-token` to re-scope the personal xoxp token (#1210)

souliane:
## Summary

The personal Slack OAuth token at `pass slack/user-oauth-token` is the credential `SlackBotBackend` uses to post and react in Slack-Connect externally-shared channels (where the bot token is rejected with `mcp_externally_shared_channel_restricted`). Existing installations are missing scopes that recent backend work depends on — notably `reactions:write`, `chat:write.public`, `chat:write.customize`.

This PR adds `t3 setup slack-user-token [--reset]`, which:

- Lists the required user-scope set (preserves every scope currently held, adds `reactions:write`, `chat:write.public`, `chat:write.customize`).
- Opens the Slack app's install page so the user reinstalls and re-consents.
- Prompts the user to paste the new `xoxp-…` token.
- Verifies the granted scopes via the `x-oauth-scopes` header on `auth.test`. Fails loudly with the missing scope name if any are absent.
- Stores the token via `pass insert slack/user-oauth-token` (overwriting on `--reset`, confirming first in default mode).

Also extends the slack-bot manifest's `_USER_SCOPES` to include `chat:write.public` / `chat:write.customize` so `t3 setup slack-bot --update` will push these on next manifest sync (Slack only grants what the manifest declares).

Token-split into separate read/write entries (the architecture-decoupling work) is intentionally out of scope here — that lives in #1209. This PR just gets the existing single token re-scoped.

## Test plan

- [x] `tests/cli_setup/test_slack_user_token.py` — 22 tests covering token-regex, scope-list constant, helpers, `fetch_token_scopes` (Slack API mock), `_store_and_verify`, and CLI walkthrough (happy path, missing-scope fail-loud, default-mode confirm-prompt, `--reset` overwrites silently).
- [x] `uv run pytest tests/cli_setup/ tests/test_slack_setup.py tests/cli_doctor/ --no-cov` — 229 passed.
- [x] `uv run ruff check` clean across the changed files.
- [x] `uv run t3 setup slack-user-token --help` renders correctly.

Refs #1210

### DIFF
--- a/src/teatree/cli/slack_setup.py
+++ b/src/teatree/cli/slack_setup.py
@@ -84,12 +84,26 @@ _BOT_SCOPES = [
 # ``users:read`` (handle/id resolution). Listing only the two reaction
 # scopes would silently revoke those on reinstall.
 _USER_SCOPES = [
+    "canvases:read",
+    "canvases:write",
+    "channels:history",
     "chat:write",
     "chat:write.customize",
     "chat:write.public",
+    "files:read",
+    "groups:history",
+    "im:history",
+    "mpim:history",
     "reactions:read",
     "reactions:write",
+    "search:read.files",
+    "search:read.im",
+    "search:read.mpim",
+    "search:read.private",
+    "search:read.public",
+    "search:read.users",
     "users:read",
+    "users:read.email",
 ]
 _BOT_EVENTS = ["app_mention", "message.im"]
 
--- a/src/teatree/cli/slack_setup.py
+++ b/src/teatree/cli/slack_setup.py
@@ -85,6 +85,8 @@ _BOT_SCOPES = [
 # scopes would silently revoke those on reinstall.
 _USER_SCOPES = [
     "chat:write",
+    "chat:write.customize",
+    "chat:write.public",
     "reactions:read",
     "reactions:write",
     "users:read",
--- a/src/teatree/cli/slack_user_token_setup.py
+++ b/src/teatree/cli/slack_user_token_setup.py
@@ -1,0 +1,208 @@
+"""``t3 setup slack-user-token`` — re-scope the personal Slack OAuth (xoxp) token.
+
+The personal token at ``pass slack/user-oauth-token`` is the credential that
+``SlackBotBackend`` uses to post and react in Slack-Connect externally-shared
+channels (where the bot token is rejected with
+``mcp_externally_shared_channel_restricted``). Several scopes that recent
+backend work depends on — notably ``reactions:write``, ``chat:write.public``
+and ``chat:write.customize`` — are missing from existing installations because
+they were not yet in the manifest when the user last consented.
+
+This command walks the user through reinstalling the Slack app to re-prompt
+OAuth consent for an updated user-scope set, then captures the new ``xoxp-…``
+token into ``pass slack/user-oauth-token`` after verifying the token actually
+carries the requested scopes (Slack returns the granted scope set in the
+``x-oauth-scopes`` response header of ``auth.test``).
+
+This is intentionally separate from ``t3 setup slack-bot``: the bot command
+manages the manifest + bot/app tokens; this command focuses on the personal
+xoxp token capture and scope verification step.
+"""
+
+import re
+import webbrowser
+from pathlib import Path
+
+import httpx
+import typer
+
+from teatree.cli.slack_setup import _USER_SCOPES, app_install_url
+from teatree.config import CONFIG_PATH
+from teatree.utils.secrets import read_pass, write_pass
+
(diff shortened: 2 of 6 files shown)


souliane:
Self-review (maker/checker): APPROVE.

- Code mirrors existing `slack_setup.py` patterns; full typing; module functions count is consistent with the codebase convention for thin CLI modules.
- Tests (22) cover happy path / missing-scope fail-loud / --reset overwrite semantics / default-mode confirm prompt.
- Wider sweep: `uv run pytest tests/cli_setup/ tests/test_slack_setup.py tests/cli_doctor/ --no-cov` → 229 passed.
- `uv run ruff check` clean across changed files.
- `uv run t3 setup slack-user-token --help` renders correctly.

Nit (non-blocking): the overwrite-confirmation prompt fires before the scope list is printed, so the user confirms without yet seeing what they would be re-consenting to. UX only — not correctness.

src/teatree/cli/slack_user_token_setup.py
```diff
@@ -0,0 +1,208 @@
...
+        "files:read",
+        "canvases:read",
+        "canvases:write",
+        "search:read.public",
+        "search:read.private",
+        "search:read.mpim",
+        "search:read.im",
+        "search:read.files",
+        "search:read.users",
+    }
```
souliane: Blocker — this set declares 18 scopes including `canvases:read`/`canvases:write`/`search:read.{public,private,mpim,im,files,users}`/`channels:history`/`groups:history`/`im:history`/`mpim:history`/`files:read`/`users:read.email`. But `_USER_SCOPES` in `slack_setup.py` only declares 6. Slack only grants what the manifest declares, so `auth.test` will report 12+ missing scopes on every run. Either trim this to match the manifest, or extend `_USER_SCOPES` to declare the full set.

src/teatree/cli/slack_setup.py
```diff
@@ -85,6 +85,8 @@ class SlackManifestError(RuntimeError):
 # scopes would silently revoke those on reinstall.
 _USER_SCOPES = [
     "chat:write",
+    "chat:write.customize",
+    "chat:write.public",
```
souliane, line 92: Counterpart to the blocker on `REQUIRED_USER_SCOPES`: only `chat:write.customize` + `chat:write.public` are being added here, but the verification step in `slack_user_token_setup.py` expects 14 more scopes. They must match or reinstall + `auth.test` verification can never succeed.

souliane (commented):
Independent reviewer pass (maker/checker split). Two blockers — see inline.

1. Scope-set mismatch: `REQUIRED_USER_SCOPES` requests 18 scopes; the manifest's `_USER_SCOPES` only declares 6. Slack only grants what the manifest declares, so `auth.test` verification will report 12+ missing scopes on every run — the command can't succeed as shipped.

2. Two unrelated features in one PR: the `skills/scanning-news/` skill is wholly unrelated to `slack-user-token`. Worth splitting so the slack change can land cleanly.

Self-review prompt-ordering nit is non-blocking, agreed.

souliane:
Addressed reviewer blockers — force-pushed clean branch.

**Blocker 1 (scope alignment)**: `_USER_SCOPES` in `slack_setup.py` extended from 6 to 20 entries (every scope the verifier checks). `REQUIRED_USER_SCOPES` is now derived as `sorted(_USER_SCOPES)` so both sides cannot drift again. New drift-guard test (`test_matches_manifest_scope_set_exactly`) asserts set equality. Verified at runtime: both sides = 20 scopes, sets equal.

**Blocker 2 (scope bundling)**: scanning-news commit (`5e22cc1e`) and the now-unneeded scanning-news doc-fix dropped from the branch. The scanning-news skill already has its own focused PR at #1190 — no new split-PR needed. Branch was rebuilt off `origin/main` and force-pushed (`--force-with-lease`), so only the two slack-user-token commits remain.

**UX nit (also addressed)**: scope list now printed before the overwrite-confirmation prompt. New test (`test_scope_list_printed_before_overwrite_prompt`) asserts ordering in stdout.

**Latent inconsistency (filed as follow-up)**: `USER_TOKEN_PASS_KEY` hardcoded vs per-overlay `user_token_ref` contract — tracking separately, not in this PR.

**Verification**:
- `uv run ruff check` — clean
- `uv run pytest tests/cli_setup/test_slack_user_token.py tests/test_slack_setup.py` — 99/99 passed (was 97 before, +2 new guards)
- Wider sweep `tests/cli_setup/ tests/test_slack_setup.py tests/cli_doctor/` — 231/231 passed
- Pre-push hook full Docker pytest matrix — green

New HEAD: `0194120c`. Re-review when ready @a8b0c710ebc8bfd1c.

Issue #810, opened 2026-05-16
fix(hooks): Stop hook crashes with ModuleNotFoundError: teatree — must fail-safe when the hook interpreter lacks teatree


souliane:
## Symptom (live, this session)
A `Stop` hook crashed with a traceback to the session:

```
ModuleNotFoundError: No module named 'teatree'
  plugins/t3/hooks/scripts/hook_router.py:1082  _prune_dead_owner
    from teatree.utils.singleton import pid_alive  # noqa: PLC0415
  plugins/t3/hooks/scripts/hook_router.py:1399  _session_owns_loop
  plugins/t3/hooks/scripts/hook_router.py:1432  handle_loop_self_pump
```

The hook reported "Failed with non-blocking status code" so the session continued, but the loop self-pump / dead-owner prune is broken and a full traceback is surfaced to the user.

## Two distinct defects
1. **Immediate (fail-safe):** a `Stop` hook must NEVER raise to the session. The lazy `from teatree.utils.singleton import pid_alive` (and any teatree import in the hook path) must be guarded — on `ModuleNotFoundError`/`ImportError`, degrade gracefully (skip the self-pump / treat owner as unknown) with a single logged line, never a traceback. Hooks run under whatever interpreter the agent harness invokes; teatree-importability is not guaranteed there.
2. **Structural (root):** the hook-runner interpreter lacks `teatree` on its path — the same env-divergence class as the dep-drift/running-interpreter issue (relates #805) and the loop-registry code this lives in is slated for replacement by the loop singleton redesign (relates #786 WS3, which retires `_read_loop_registry`/`_prune_dead_owner`/`_session_owns_loop`).

## Acceptance
- The `Stop` hook path cannot raise on a missing/unimportable `teatree` — regression test simulates `teatree` unimportable in the hook context and asserts the hook returns cleanly (no exception, no traceback), self-pump simply skipped. RED before the guard, GREEN after (anti-vacuous).
- Defensive only / minimal — must carry forward cleanly when #786 WS3 replaces this registry code and must not conflict with the #807 Stop-gate handler ordering.
- `Relates-to #805 #786`.

souliane:
Already delivered — fixed via #816 (merged; commit 32beabe 'fail-safe Stop hook when interpreter lacks teatree' on main). PR used `Relates-to` not `Closes`; closing as already-resolved.

PR #1240, opened 2026-05-20
feat(hooks): Stop-hook consideration gate for framework-improvement promotion (#1129)

### LINKED ISSUE #1129: loop: framework-improvement consideration gate — every session edit must answer 'should this be a teatree feature?' before declaring done
## Directive

Requirement: when the agent encounters a behavior gap or framework limitation in teatree, it should consider whether the situation justifies a framework improvement instead of (or in addition to) a one-shot fix. This should be enforced — likely via a Stop hook that prompts the agent to log a framework-improvement candidate before terminating.

The agent today makes one-off fixes (memory writes, overlay-local config, personal `~/.claude/settings.json` edits) without considering whether the same fix should ship into teatree so every user benefits. The retro skill's § 9 "Consolidation over Drift" doctrine says exactly this in prose, but prose-only enforcement loses every time — memory-as-vigilance fails.

A **deterministic gate** is needed: every session-edit must be classified before the session can declare done.

## Proposed implementation

Add a `Stop` (or `PreCompact`) hook that scans the session's modified files and classifies each:

| Classification | Path pattern (heuristic) | Required action |
|---|---|---|
| **(P) Promote** | edits in personal config the framework should ship: hooks shell scripts, agent personal settings, overlay-local utilities that match a pattern other overlays would benefit from | Agent must either open a teatree issue OR declare in the Stop output why the edit is genuinely personal-only |
| **(C) Documented config** | edits to overlay-specific config that's a legitimate divergence | Agent must point to the teatree config knob that exposes this divergence; if no knob exists, file an issue to add one |
| **(K) Keep personal** | genuine user preferences (theme, voice, paths) | Agent declares in Stop output |

The hook produces a one-line classification per modified path; if any are `(P)` without a teatree issue link, **block the Stop until the agent either opens the issue or downgrades the classification with justification**.

## Out of scope here

The exact heuristics for the (P)/(C)/(K) classifier — let it start simple (path-based) and refine as we observe false-positives/negatives.

## Related

- teatree retro skill § 9 "Consolidation over Drift" (prose statement of the same doctrine)
- [[feedback_redcard_use_workflow_not_manual_dispatch]] (the workflow-bypass failure mode this gate prevents)
- Autonomous factory umbrella #836

souliane:
## Summary

Adds a Stop-hook gate (#1129) that scans the current turn's `Edit` / `Write` / `NotebookEdit` tool uses and classifies each modified path. When one or more land on personal agent config the framework should arguably ship (e.g. `~/.claude/settings.json`, `~/.claude/hooks.json`, personal `CLAUDE.md` / `AGENTS.md`), the gate emits a `CONSIDERATION GATE` reminder in `additionalContext` — unless the assistant turn already references a teatree issue (`souliane/teatree#NNNN` or bare `#NNNN`).

Doctrine: retro skill section 9 "Consolidation over Drift" covers the same ground in prose; this gate makes the scan deterministic, per the issue's "memory-as-vigilance fails" point.

## Classifier

Path-based and conservative:

- **(P) Promote** — agent-harness config files (`settings.json`, `hooks.json`, `CLAUDE.md`, `AGENTS.md`) that ship behaviour.
- **(K) Keep** — memory entries, todos, shell rc, terminal config.
- **None** — anything else, including the framework's own source.

`(K)` wins over `(P)` when both could match, so memory writes under `~/.claude/projects/<p>/memory/` stay silent.

## Behaviour

- Soft block only — emits `additionalContext`, never `decision: block`. A genuinely-done turn isn't hard-blocked.
- Returns `True` to break the Stop chain, preserving JSON shape and preempting the loop self-pump.
- Crash-proof per #810 — outer boundary guard swallows unexpected errors with a stderr breadcrumb.
- `stop_hook_active` short-circuits to a no-op (re-fire guard).
- Class (C) "documented config" deliberately deferred — overlapping heuristics with (P) risk noise; retro skill section 9 still covers it.

## Test plan

- [x] 36 new tests in `tests/test_consideration_gate_stop_hook.py` covering classifier, gate behaviour, router wiring, and crash-proof contract.
- [x] Sibling Stop-hook tests (`test_answered_at_gate_stop_hook.py`, `test_loop_self_pump_hook.py`, `test_hook_router_cadence_hook.py`) still green.
- [x] Full hook-test suite (513 tests) green.
- [x] `uv run ruff check` + `uv run ruff format` clean.
- [x] `uv run ty check` clean.

Fixes #1129.

### DIFF
--- a/hooks/scripts/hook_router.py
+++ b/hooks/scripts/hook_router.py
@@ -3518,6 +3518,222 @@ def _enforce_answered_questions(data: dict) -> bool | None:
     return True
 
 
+# ── Consideration gate (#1129): promote framework-shaped edits ──────
+#
+# Every session that edits personal config the framework should ship
+# (e.g. ``~/.claude/settings.json``, ``~/.claude/hooks.json``, personal
+# ``CLAUDE.md`` behavioural rules) must answer "should this be a teatree
+# feature?" before the turn declares done. Prose-only enforcement loses
+# (see retro skill § 9 "Consolidation over Drift"); this gate makes the
+# scan deterministic.
+#
+# The classifier is path-based and conservative. Three classes:
+#
+#   (P) Promote — personal agent config a teatree installation should
+#       wire automatically. The gate fires unless the assistant turn
+#       references a teatree issue (``souliane/teatree#NNNN`` or bare
+#       ``#NNNN``) OR a later iteration downgrades the path.
+#   (K) Keep    — genuine personal preference (memory entries, shell
+#       rc, terminal config). The gate stays silent.
+#   None        — path lives outside the personal-config corners (or
+#       inside the framework itself). The gate has nothing to say.
+#
+# Class (C) "documented config" is not encoded here yet — overlapping
+# heuristics with (P) make false positives noisy. The retro skill
+# already covers (C) in its consolidation pass; the Stop gate focuses
+# on the loudest signal first.
+
+_TEATREE_ISSUE_REF = re.compile(
+    r"(?:souliane/teatree)?#(\d{2,})\b",
+    flags=re.IGNORECASE,
+)
+
+_PROMOTE_PATTERNS: tuple[re.Pattern[str], ...] = (
+    # Agent-harness config files that ship behaviour.
+    re.compile(r"/\.(claude|codex|cursor|copilot)/settings(\.local)?\.json$"),
+    re.compile(r"/\.(claude|codex|cursor|copilot)/hooks\.json$"),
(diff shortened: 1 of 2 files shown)


souliane (commented):
Independent review verdict: APPROVE (claude-opus-4-7). Typing modern (str|None, no Any in prod), 36-test mirror present, soft-block contract enforced (no decision:block), router order verified, crash-proof guard documented per #810, commit identity correct, no Co-Authored-By. CI: test(3.13) green; uv-audit failure is pre-existing on main. t3 CLEAR id 148 issued; t3 merge step refused on uv-audit live-check (escalated to durable backlog).

PR #1261, opened 2026-05-20
fix(ship): align GitHub backend url field with ship runner — close silent pr create failure (#1226 #1222)

### LINKED ISSUE #1222: t3 teatree pr create: GitHub backend returns 'url' but ship runner reads 'web_url'/'html_url' → silent zero-URL PR creation
## Problem

`t3 teatree pr create` reports `ok: True` and logs "opened PR" with an empty URL for every GitHub PR. Two sub-agents this session hit this independently (#1220 slack-user-token, #1221 Phase 2 trigger drops). Both had to fall back to raw `gh pr create` and back-fill `ticket.extra['pr_urls']` to keep the t3 DB in sync.

## Root cause

`src/teatree/core/runners/ship.py:147` reads:

```python
url = pr.get("web_url") or pr.get("html_url")
```

…but `src/teatree/backends/github.py:281` returns `{"url": ...}` (and presumably the rest of the metadata dict). The keys don't match. The GitLab backend probably returns `web_url` (the GitLab API native key), so the GitLab path works; the GitHub backend was added later and used `url` without aligning.

Result: every GitHub PR creation is "successful" from the runner's perspective but persists no URL, no PR row gets attached to the ticket, and downstream gates (ensure-pr, the auto-revisit logic) see "no PR exists" and behave wrong.

## Acceptance criteria

1. The GitHub backend returns the canonical `web_url` key (or the ship runner reads both keys — pick one normalization point, not both).
2. A regression test that mocks the backend's return and asserts the ship runner persists a non-empty URL.
3. The two affected PRs (#1220, #1221) get their `ticket.extra['pr_urls']` validated (already back-filled by the sub-agents; just verify).
4. `ensure-pr` correctly detects existing PRs after the fix lands (no more "skipped: branch synced to origin/main" false-positive on a branch that's 3 commits ahead).

## Related

- #1220 (slack-user-token) — hit this bug, fell back to gh pr create
- #1221 (Phase 2 trigger drops) — hit this bug, fell back to gh pr create
- Resilience epic #1192 — this is a per-boundary instance of invariant 1 (verify-by-re-read at every external write); the runner didn't verify the URL was non-empty before claiming success

### LINKED ISSUE #1226: bug(ship): t3 teatree pr create returns ok=True but does not actually open the PR (silent host.create_pr failure)
## Symptom

`t3 teatree pr create <ticket> --sync` completes successfully:

- log line: `Ship executor pushed s-teatree-1223-ticket and opened PR`
- return payload: `{'ticket_id': 388, 'state': 'in_review', 'synced': True, 'ok': True, 'detail': ''}`

But no PR was actually opened on GitHub. `gh pr list --repo souliane/teatree --state all --head s-teatree-1223-ticket` returns `[]`. The branch is pushed, but the host.create_pr call silently fails (or returns a falsy PR object) without emitting an error or flipping `ok` to `False`.

## Workaround

Fall back to raw `gh pr create` after `t3 teatree pr create` reports success — then verify the PR actually exists via `gh pr list`.

## Why this is a problem

The state machine transitions to `in_review` and the `Ship executor pushed … and opened PR` log fires, so any downstream automation (the §17 review pipeline, the §5.6 loop scanners, the Slack review-channel poster) sees a "shipped" ticket with no PR to look at. The reviewer sub-agent has nothing to review; the merge-clear flow stalls. The misleading `ok=True` is the actual bug — a real failure must propagate as `ok=False` with a `detail` explaining the host failure.

## Likely root cause (untested hypothesis)

`teatree.core.runners.ship.execute_ship` (or whatever path leads to the `Ship executor pushed … and opened PR` log message) probably catches a `host.create_pr` exception or treats a `None`/empty PR object as success. Needs a re-read of the live state (verify-by-re-read invariant from #1192) immediately after the create call.

## Acceptance criteria

- [ ] `host.create_pr` failure surfaces as `{'ok': False, 'detail': '<reason>', 'state': '<pre-ship state>'}` — never `ok=True`
- [ ] After a successful create, a `host.get_pr(branch)` re-read confirms the PR exists before the success result is emitted
- [ ] Regression test: mock `host.create_pr` to return `None`; assert the result is `ok=False` and the FSM state did not advance

## Repro

1. Provision a worktree on a fresh ticket
2. Implement a one-line change, commit, run all pre-commit gates clean
3. `t3 <overlay> pr create <ticket> --sync`
4. Observe success message
5. `gh pr list --repo <repo> --head <branch>` returns `[]`

Surfaced while delivering #1223 — fallback to `gh pr create` was used, PR #1225 was opened that way.

adrien-oper:
## Summary

Close two related bugs around `t3 teatree pr create` silently succeeding on the GitHub backend:

- **#1222** — GitHub backend returned `{"url": <url>}` from `create_pr`, but `ShipExecutor` and `_ensure_pr` consumers read `web_url` / `html_url` (the cross-host canonical keys GitLab's API uses natively). The field-name mismatch made every GitHub PR creation "successful" with an empty URL.
- **#1226** — Symptom: `Ship executor pushed … and opened PR` log fires, result is `{'ok': True, 'detail': ''}`, but `pr_urls` is empty so downstream gates think no PR exists.

## Fix

- GitHub backend (`src/teatree/backends/github.py`) returns the canonical `web_url` key.
- Backend enforces the verify-by-re-read invariant: an empty / non-URL `gh pr create` stdout (e.g. the "no commits between" pre-push race that exits 0) is rejected as `CommandFailedError`, so `ok=True` never escapes with no PR.
- Ship runner (`src/teatree/core/runners/ship.py`) gains a belt-and-braces guard: a backend that returns an empty / wrong-keyed payload surfaces as `ok=False` instead of advancing the FSM to SHIPPED with no `pr_urls` entry. Logic extracted into `_open_pr_and_record` to keep `run()` within complexity limits.
- `_ensure_pr` consumer (`src/teatree/core/management/commands/_ensure_pr.py`) prefers `web_url` over the legacy `url` key.

## Tests

- `tests/test_github_backend.py::test_create_pr` — backend now returns `web_url`.
- `tests/test_github_backend.py::test_create_pr_raises_when_gh_returns_no_url` — empty/non-URL `gh pr create` stdout raises.
- `tests/teatree_core/test_runners_ship.py::test_returns_failure_when_backend_returns_empty_url` — empty backend URL surfaces as `ok=False`.
- `tests/teatree_core/test_runners_ship.py::test_accepts_html_url_for_github_native_payloads` — confirms `html_url` fallback is kept for raw GitHub payloads.

Full suite: 6317 passed, 13 skipped (1 deselected unrelated pre-existing config-value test).

## Self-validation

This PR was opened via `t3 teatree pr create` itself — the operation that was silently failing — to validate the fix end-to-end. (Note: the local editable install pointed at the main clone, not this worktree, so this PR was opened against the un-patched code; once merged and the editable install is refreshed, future runs will exercise the patched path.)

Closes #1226
Closes #1222

### DIFF
--- a/src/teatree/backends/github.py
+++ b/src/teatree/backends/github.py
@@ -10,7 +10,7 @@ from urllib.parse import quote_plus, urlparse
 from teatree.backends.protocols import ApprovalState, PrOpenState, PullRequestSpec, ReviewState
 from teatree.types import RawAPIDict
 from teatree.utils import git
-from teatree.utils.run import CompletedProcess, run_checked
+from teatree.utils.run import CommandFailedError, CompletedProcess, run_checked
 
 _ISSUE_URL_RE = re.compile(r"^/(?P<owner>[^/]+)/(?P<repo>[^/]+)/issues/(?P<number>\d+)/?$")
 _PR_URL_RE = re.compile(r"^/(?P<owner>[^/]+)/(?P<repo>[^/]+)/pulls?/(?P<number>\d+)/?$")
@@ -278,7 +278,21 @@ class GitHubCodeHost:
             cmd.append("--draft")
 
         result = _run_gh(*cmd, token=self._token)
-        return {"url": result.stdout.strip()}
+        # #1222 / #1226: align with the cross-host canonical key (``web_url``)
+        # that ``ShipExecutor`` reads — returning ``url`` silently produced
+        # empty PR rows because the consumer never looked at that field. The
+        # producer also enforces the verify-by-re-read invariant: an empty /
+        # non-URL stdout (e.g. the ``no commits between`` pre-push race that
+        # exits 0) is rejected so ``ok=True`` never escapes with no PR.
+        url = result.stdout.strip()
+        if not url.startswith(("http://", "https://")):
+            raise CommandFailedError(
+                cmd,
+                result.returncode,
+                result.stdout,
+                f"gh pr create produced no PR URL (stdout={url!r})",
+            )
+        return {"web_url": url}
 
     def current_user(self) -> str:
         """Return the authenticated GitHub login (e.g. ``souliane``)."""
--- a/src/teatree/core/management/commands/_ensure_pr.py
+++ b/src/teatree/core/management/commands/_ensure_pr.py
@@ -112,4 +112,7 @@ def create_or_defer_pr(repo_path: str, branch_name: str) -> EnsurePrResult:
                 hint=f"t3 <overlay> pr ensure-pr --branch {branch_name}",
             )
         raise
-    return EnsurePrResult(branch=branch_name, url=str(raw.get("url", raw.get("web_url", ""))))
+    # #1222 / #1226: ``web_url`` is the cross-host canonical key (GitLab
+    # API native; GitHub backend was aligned to it). ``html_url`` is kept
+    # for raw GitHub API payloads piped through other producers.
+    return EnsurePrResult(branch=branch_name, url=str(raw.get("web_url") or raw.get("html_url") or ""))
--- a/src/teatree/core/management/commands/_ensure_pr.py
+++ b/src/teatree/core/management/commands/_ensure_pr.py
@@ -114,5 +114,14 @@ def create_or_defer_pr(repo_path: str, branch_name: str) -> EnsurePrResult:
         raise
     # #1222 / #1226: ``web_url`` is the cross-host canonical key (GitLab
     # API native; GitHub backend was aligned to it). ``html_url`` is kept
-    # for raw GitHub API payloads piped through other producers.
-    return EnsurePrResult(branch=branch_name, url=str(raw.get("web_url") or raw.get("html_url") or ""))
+    # for raw GitHub API payloads piped through other producers. An empty
+    # / non-URL payload surfaces as ``error`` so the orphan-branch path
+    # never silently advances with no PR — same invariant the ship runner
+    # enforces.
+    url = str(raw.get("web_url") or raw.get("html_url") or "")
+    if not url.startswith(("http://", "https://")):
+        return EnsurePrResult(
+            branch=branch_name,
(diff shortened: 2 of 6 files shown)


souliane:
## Adversarial code review (codex)

Hostile static review of commit `3b8991e9` — focus: any remaining failure mode where `t3 teatree pr create` could return `ok=True` with no actual PR? Any stale `url`-key consumers that would now break?

### Findings

**1. High — out of scope for this PR (separate ticket)**

`ShipExecutor.run` at `ship.py:110` returns `ok=True, detail=existing_urls[-1]` from any stored URL BEFORE resolving `ship_invoking_branch`, pushing, or calling `host.create_pr`. On reused-ticket / multi-workstream flows, an old PR URL short-circuits the new workstream — `pr create --sync` reports success and advances state while no PR exists for the branch being shipped.

This is a real bug but predates this PR and is a separate concern from the field-name fix. Filing as a follow-up.

**2. Medium — fixed in `d7b4fd1d`**

`_ensure_pr.create_or_defer_pr` silently accepted an empty / wrong-keyed payload (`url=""` returned with no error). Same stale contract as the original bug, just on the orphan-branch path. Now validates `http(s)://` and surfaces a missing URL as `EnsurePrResult(error=...)`.

**3. Medium — fixed in `d7b4fd1d`**

Consumer-side URL validation was presence-only — any truthy `web_url`/`html_url` passed (including `"u"` or a non-URL coerced via `str()`). The GitHub producer validates `http(s)://` but the cross-backend ship consumer did not. `ShipExecutor._open_pr_and_record` now rejects non-`http(s)://` strings the same way.

### Stale `url` consumers

None found. All remaining `url` readers are for normalized loop payloads, sync records, GitHub GraphQL/CLI list payloads, Slack/file URLs, or `Ticket.extra["prs"]` — none read the `create_pr` return contract.

### Verdict

**Ship the PR.** The field-rename fix and the verify-by-re-read guards (producer + both consumers) close the silent-`ok=True` invariant. Finding 1 is a real but separate multi-PR-reuse bug — filing as a follow-up to keep this PR scoped.

Issue #1128, opened 2026-05-20
blueprint: trim 9 prose-of-code appendices; keep only architectural invariants


souliane:
## Decision (user-directed, 2026-05-20)

The 12 appendix files created in PR #1126 (which slimmed BLUEPRINT.md from 268KB to 30KB) are themselves over-documented. The Blueprint should be **functional + architectural invariants** — not a prose mirror of the code. The "keep all 12 because they're cited from code" reasoning was circular: the cites exist because we wrote them.

### KEEP — real architectural content (3)

- `docs/blueprint/factory-architecture.md` — self-improving factory design (umbrella #836)
- `docs/blueprint/loop-topology.md` — tick + scanners + leases + answer wiring
- `docs/blueprint/configuration.md` — resolved-order chain (cited from `cli/recommended_authorizations.py`, `commands/followup.py`, etc.)

### DELETE — prose-of-code (9)

- `agent-execution.md`
- `architecture-and-package.md`
- `backends-and-sync.md`
- `command-tiers.md`
- `dependencies-and-conventions.md`
- `django-workflows.md`
- `domain-models.md`
- `overlay-system.md`
- `skills-testing-gates.md`

## Plan

1. **Audit each of the 9 deletees** for facts still load-bearing. For each load-bearing fact, find its true home:
   - Model behaviour → docstring on the model
   - OverlayBase extension point → docstring on the method + autodoc reference
   - Coding doctrine → `CLAUDE.md` / `AGENTS.md`
   - CLI invariants → `--help` text or the typer docstring
2. Promote those facts to their true home (one PR per appendix or one bundle PR depending on scope).
3. Delete the 9 appendix files. Update every consumer cross-ref (`BLUEPRINT §<old-section>` → docstring or code path).
4. Grep the repo + every sibling overlay clone for `docs/blueprint/<deletee>.md` references; rewrite them all in the same PR.
5. Update `BLUEPRINT.md` § "Architecture appendices" to list only the 3 keepers.

## Out of scope here

The 3 keepers themselves can still be trimmed; that's a separate audit. This ticket is about deleting the 9.

souliane:
## Refinement (2026-05-20)

The scope is being narrowed: BLUEPRINT keeps the full architecture-design content; implementation prose is the only thing that gets dropped. Naming classes or file paths as anchors is fine.

Issue #140, opened 2026-04-02
feat: transition-driven delivery — commit, push, retro as lifecycle side effects


adrien-oper:
## Problem

The agent regularly bypasses the intended workflow sequence (code → test → review → retro → push) because **commit, push, and retro are manual `git` commands** outside the state machine. The Session model has quality gates (`check_gate()` enforces testing → reviewing → shipping), but they only fire through `t3 pr create`. Nothing prevents the agent from calling `git commit` and `git push` directly, skipping every gate.

**Recurring failures from this session and prior retros:**
- Edited files on main clone before creating a worktree
- Pushed before running retro
- Ran `ruff check` but not `ty check` — partial lint instead of full `prek`
- 3 failed commit attempts because pre-commit hooks caught what should have been caught earlier
- `require_ticket` config never enforced because the agent didn't go through `t3` lifecycle

## Current Architecture

Three state machines exist but operate semi-independently:

### Ticket FSM (10 states, 11 transitions)
`NOT_STARTED → SCOPED → STARTED → CODED → TESTED → REVIEWED → SHIPPED → IN_REVIEW → MERGED → DELIVERED`

Key side effects already wired:
- `ticket.test()` → creates a review Task
- `ticket.review()` → creates a shipping Task (gated on completed review task)
- `Task.complete()` → calls `_advance_ticket()` which auto-advances Ticket state based on phase

### Worktree FSM (4 states, 6 transitions)
`CREATED → PROVISIONED → SERVICES_UP → READY`

Handles ports, DB, services. No git operations as side effects.

### Session Phase Gating
```python
_REQUIRED_PHASES = {
    "reviewing": ["testing"],
    "shipping": ["testing", "reviewing"],
    "requesting_review": ["shipping"],
}
```

`check_gate()` raises `QualityGateError` if prerequisites missing. Currently only enforced by `t3 pr create`.

### What's Missing

| Operation | Current | Proposed |
|-----------|---------|----------|
| `git commit` | Manual, no gates | Side effect of `t3 lifecycle finalize` or phase transition |
| `git push` | Manual, no gates | Side effect of `t3 lifecycle ship` |
| Retro | Manual skill invocation | Required phase before shipping gate passes |
| Full lint (`prek`) | Only in pre-commit hook | Run as transition precondition |
| `require_ticket` check | Only in `t3:code` skill prose | Enforced by CLI before any write operation |

### 1. Add missing phases to the gate system

```python
_REQUIRED_PHASES = {
[...]

PR #1265, opened 2026-05-20
feat(prek): hard-fail BLUEPRINT.md > 100KB without override (#1180)

### LINKED ISSUE #1180: pre-commit: hard-fail when BLUEPRINT.md exceeds 100KB without explicit override
**Problem**

Per [[feedback-blueprint-architecture-not-implementation-prose]] (2026-05-20 user mandate): BLUEPRINT.md exceeded 260KB by accumulating implementation prose that should not be there. the decision is: \"It must be digestible\".

The skill-prose-ban gate (#140) is the precedent — it hard-fails new prose-only skill files. The BLUEPRINT needs the same shape.

**Vision**

A pre-commit gate (registered in `.pre-commit-config.yaml` or `prek.yaml`) that:

1. Measures `wc -c BLUEPRINT.md`
2. Exits non-zero if > 100000 (100 KB) bytes
3. Allow opt-out via `BLUEPRINT_SIZE_OVERRIDE=1` env var for genuinely-architectural growth

The cap is **soft policy** — it doesn't ban large blueprints, it just forces a per-commit ack that the growth is intentional.

**Out of scope**

- Auto-trimming the file (separate concern)
- Catching specific implementation-prose patterns (semantic, can't be deterministic)

**Acceptance criteria**

- New gate `tools/check_blueprint_size.py` (or similar)
- Registered in pre-commit/prek
- Test confirming current BLUEPRINT.md state vs threshold
- Documented in BLUEPRINT.md itself: 1-line cap + override env var

**Related**

- [[feedback-blueprint-architecture-not-implementation-prose]] (memory STOPGAP)
- #140 (skill-prose-ban — precedent gate)

souliane:
## Summary

- Adds `scripts/hooks/check_blueprint_size.py` — stdlib-only gate that hard-fails any commit when `BLUEPRINT.md` exceeds 100 KB (100 * 1024 bytes).
- Registers it as a local prek hook (`check-blueprint-size`) scoped to `^BLUEPRINT\.md$` so it fires only when the file is touched.
- Documents the cap and the `T3_BLUEPRINT_SIZE_OVERRIDE` escape hatch in a new "Maintenance" section of BLUEPRINT.md.
- Unit tests cover pass / fail / override paths via `tmp_path`.

Mirrors the #140 skill-prose-ban shape and complements the soft #1128 corpus-budget gate.

Closes #1180.

## Test plan

- [x] `uv run pytest tests/test_check_blueprint_size.py --no-cov -x` — 7 passed
- [x] `prek run --all-files` — green, including the new `BLUEPRINT.md size cap (100KB)` hook
- [x] Current `wc -c BLUEPRINT.md` ≈ 49 KB — well under the cap; the gate does not fire on this PR

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -243,6 +243,12 @@ repos:
         pass_filenames: false
         always_run: true
         stages: [commit-msg]
+      - id: check-blueprint-size
+        name: BLUEPRINT.md size cap (100KB)
+        language: system
+        entry: uv run python scripts/hooks/check_blueprint_size.py
+        files: ^BLUEPRINT\.md$
+        pass_filenames: false
 
   # Phase 7c - Version sync
   - repo: local
--- a/BLUEPRINT.md
+++ b/BLUEPRINT.md
@@ -381,6 +381,12 @@ Implementation details that previously lived in nine prose-of-code appendices (`
 
 ---
 
+## Maintenance
+
+A pre-commit gate (`scripts/hooks/check_blueprint_size.py`, [#1180](https://github.com/souliane/teatree/issues/1180)) hard-fails any commit when this file exceeds 100 KB. The cap forces a per-commit acknowledgement that further growth is architectural, not implementation prose. To raise the cap for a planned, reviewed bump in the same commit, set `T3_BLUEPRINT_SIZE_OVERRIDE=1` in the environment.
+
+---
+
 ## Module Dependency Graph
 
 <!-- tach-dependency-graph:start -->
--- a/BLUEPRINT.md
+++ b/BLUEPRINT.md
@@ -383,7 +383,7 @@ Implementation details that previously lived in nine prose-of-code appendices (`
 
 ## Maintenance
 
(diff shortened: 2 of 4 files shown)


souliane:
## codex adversarial-review verdict (post-push)

**Verdict at first review:** request changes.

Findings (both addressed in follow-up commit `657f5e20`):

1. Override env-var truthiness was permissive. `os.environ.get(name)` returned a truthy value for any non-empty string, so `T3_BLUEPRINT_SIZE_OVERRIDE=0` or `=false` silently bypassed the cap — diverging from the sibling #1128 gate which requires the exact string `"1"`. Tightened to `== "1"` and added a parametrized test covering `""`, `"0"`, `"false"`, `"no"`, `"true"`.

2. BLUEPRINT.md "Maintenance" wording claimed the gate hard-fails "any commit". The prek `files: ^BLUEPRINT\.md$` filter only fires when the file is touched, so the doc and the config disagreed. Tightened wording to "any commit that touches this file" and clarified the script docstring.

The `files:` scoping (versus `always_run: true`) is intentional and matches the spec in #1180 — the gate is meant to challenge each growth event, not gate every unrelated commit. The deterministic re-run on every BLUEPRINT touch is the per-commit acknowledgement the cap exists to force.

Issue #1192, opened 2026-05-20
Resilience as a cross-cutting property (verify-by-re-read, fallback-transport, idempotency, heartbeat, sub-agent return contract)


souliane:
## Background

resilience is now treated as a first-class teatree concern — a precondition for self-improvement rather than something layered on after the fact. Every external-side write must satisfy all five invariants below, demonstrated by tests covering both happy and degraded paths.

Captured in `feedback_resilience_is_a_precondition_for_self_improvement`.

## The 5 invariants

1. **Verify-by-re-read.** Every write returns an identifier (slack ts, glab discussion id, gh PR number, MR `clear_id`). The same call or a follow-up re-fetches that identifier and confirms the side-effect actually landed. `rc=0` is not verification.
2. **Fallback transport.** If the primary transport fails (Slack MCP errors, `notify_send` rc!=0, GitLab 5xx), the write degrades to a second channel — a known disk path, a different Slack workspace, a GitHub issue body, an email — anything except silent loss. The fallback's identity is recorded.
3. **Idempotency keys.** Every outbound write carries a dedup key so retry-after-uncertainty is safe. If we don't know whether the prior call landed, the retry won't double-post.
4. **Heartbeat + out-of-band escalation.** The loop emits a liveness signal on every tick. If N minutes elapse without a heartbeat, an OS-level watchdog fires an out-of-band alert (NOT via any channel the heartbeat is watching). The user always knows whether the agent is alive.
5. **Sub-agent return contract.** When a sub-agent claims a structural side-effect (post landed, PR opened, FSM transition), the parent verifies the identifier programmatically before accepting. Failed verification escalates, never silently absorbs.

## Why now

Every silent failure in the 2026-05-20 session was a fail-open at a boundary: `notify_send` rc=1 silent loss; MCP token invalidated by `/login`; inline GitLab POST coerced to MR-level reported as success; `t3 loop tick` SKIP on non-owner only seen because of the log; foreground Bash hung 1h+; sub-agent claimed structural side-effects the parent never re-fetched. Each is individually patched but the cross-cutting property — resilience itself — is not yet a first-class teatree concern. This epic makes it one.

## Coverage matrix

Legend: white-check satisfied / cross failing / circle partial / N/A structurally non-applicable.

| # | Write path | (1) Verify-by-re-read | (2) Fallback transport | (3) Idempotency key | (4) Heartbeat | (5) Sub-agent contract |
|---|---|---|---|---|---|---|
[...]

PR #1269, opened 2026-05-20
feat(github): outbound-audit claim+verify on post_pr_comment + post_issue_comment (#1198)

### LINKED ISSUE #1198: Resilience: GitHub post_pr_comment and issue-comment — outbound-audit claim + verify
Part of resilience epic #1192 (invariant 1 + invariant 3; rows 15, 17).

## Current behavior

`backends/github.py:322` `post_pr_comment()` and `backends/github.py:373` `post_issue_comment()` post via `gh api repos/.../comments` and return the raw response. Neither records an `OutboundClaim`; the outbound-audit scanner has NO `github_note` kind.

This was the GitHub-side gap noted as out-of-scope in #1164's body. With the resilience epic in place it is now in-scope.

## Desired behavior

- Both functions call `record_claim(kind="github_note", idempotency_key=f"github_note:{repo}#{pr_or_issue}:{comment_id}", ...)` after a successful POST.
- New `OutboundClaim.Kind.GITHUB_NOTE` plus a verifier that GETs the comment by id and asserts its `body` hash matches the intended body.
- The `outbound_audit` scanner's `kind_settling_seconds` dict gains `"github_note": 30`.

## Acceptance criteria

- Failing test: POST returns id=N, GET on /N returns 404; assert drift recorded and user DM'd.
- Passing test: POST and GET match; assert `verified_at` set.
- Idempotent retry test.

souliane:
Closes #1198. Part of resilience epic #1192 (invariant 1 + invariant 3; rows 15, 17).

## Summary

- `OutboundClaim.Kind.GITHUB_NOTE` added (migration `0032_add_github_note_kind`); ledger now covers GitHub the same way it covers GitLab.
- `post_pr_comment` and `post_issue_comment` record a claim of kind `github_note` on every successful POST, keyed `github_note:<repo>#<num>:<comment_id>`, with `payload_digest = sha256(body)` so the verifier can detect silent body-divergence.
- Best-effort recording: any exception (Django not booted, DB outage, integrity race) is swallowed inside `_record_github_note_claim` so a ledger outage never breaks a publish that already succeeded.
- New `_github_note_verifier` in `loop.scanners.outbound_audit` GETs `repos/<repo>/issues/comments/<id>` and asserts body-digest match. 404 flips the claim to drift; any other transport error re-raises so `scan()` skips the row silently (no spammy drift DMs on a temporary GitHub-API outage).
- `kind_settling_seconds["github_note"] = 30` per the AC.

## Acceptance criteria coverage

- **Failing test (drift):** `test_404_returns_drift` + `test_500_in_scanner_is_caught_and_claim_not_flagged` + `test_body_digest_mismatch_returns_drift` cover the missing-comment and body-divergence drift paths.
- **Passing test (verified):** `test_body_digest_match_returns_ok` + the end-to-end `test_post_pr_comment_records_github_note_claim` cover the happy path.
- **Idempotent retry:** `test_post_pr_comment_idempotent_on_repeated_post` proves a retried POST that GitHub collapsed to the same comment_id is a single ledger row.

## Architectural fit

Mirrors the GitLab pattern (`_gitlab_note_verifier` in the same module): same error doctrine, same `extra` shape, same swallow-on-best-effort guarantee. The hashing helper (`_hash_body`) lives in `outbound_audit.py` next to the verifier that consumes it; the backend duplicates the 2-line hashlib call inline rather than importing across the `teatree.backends` → `teatree.loop` boundary (`tach.toml` forbids it).

`BLUEPRINT.md` §4 updated to enumerate the recognised claim kinds; `docs/blueprint/loop-topology.md`'s `outbound_audit` row updated to match.

## Test plan

- [x] `uv run pytest tests/test_github_backend.py tests/teatree_loop/test_outbound_audit.py --no-cov -q` — 123 passed
- [x] `uv run pytest --no-cov -q` — 4947 passed (1 pre-existing failure on main: `test_in_repo_overlay_resolves_to_three`, unrelated)
- [x] `prek run --all-files` — passes (ruff, tach, ty-check, blueprint-sync, banned-terms, conventional-commit)
- [x] `manage.py makemigrations --check --dry-run` — no diff


souliane:
## Codex adversarial-review verdict

Two passes ran against the branch. Latest verdict: `needs-attention`.

### Pass 1 (initial commit `3dc1178a`)

**Finding (high):** "GitHub verifier drops the posting credential" — `_github_note_verifier` constructed `GitHubCodeHost()` with no token and `_gh_api_get()` was called without one. On private repos a 404 might be "no auth" rather than "comment missing" → mis-classified as drift.

**Addressed in commit `763ecda7`:** added `_resolve_github_token()` (env `GH_TOKEN`/`GITHUB_TOKEN` → `pass` fallback, mirroring `_resolve_token` in `gitlab_api.py`) and threaded the resolved token into `_gh_api_get`. Empty token → factory returns `None` and the scanner skips `github_note` rows silently. New regression test: `test_verifier_forwards_resolved_token_to_gh_api_get`.

### Pass 2 (after fix commit)

**Finding (high):** "GitHub verifier uses a global token instead of the posting overlay token" — with multi-overlay setups or a non-default `github_token_ref`, the global resolver might pick the wrong credential (or none).

**Disposition:** noted but not addressed in this MR. The same architectural shape exists today for the GitLab verifier (`_resolve_token` is also process-global), so the multi-overlay binding is a broader resilience-epic story rather than a #1198 MVP blocker. Filing as a follow-up. The current implementation is no worse than the GitLab side it mirrors, and the env→pass fallback covers the single-overlay case that the issue's acceptance criteria target.

**Not merging** — leaving the PR open for human review. Per-overlay token binding will be raised as a separate ticket against the resilience epic (#1192).

souliane:
Filed the multi-overlay token-binding follow-up as #1275 (applies symmetrically to GitLab/Slack/Notion verifiers — a horizontal resilience-epic improvement, not GitHub-specific).

Issue #1703, opened 2026-06-01
validate-mr-metadata gate misses file-based and out-of-band-API description edits


adrien-oper:
## Problem

The `validate-mr-metadata` PreToolUse gate (`hook_router._extract_mr_fields`) only parses the title/description from an inline `glab mr create/update --description 'x'` command string. A non-compliant description that reaches the remote through a different surface slips past it:

- a file-based / heredoc / command-substitution description (`-F`, `--description-file`) — the inline-quote regex captures nothing, so the gate validates an empty string and passes;
- an out-of-band REST API write (`glab api`/`gh api` PUT/POST to a merge-requests or pulls endpoint with `--field title=`/`-f description=`/`-f body=`) — the web-UI-equivalent edit that never goes through the create/update CLI.

A non-compliant first line then fails the remote MR-title/description CI job, blocking the merge after the fact — the exact failure the local gate is meant to prevent.

## Fix

Extend `_extract_mr_fields` to cover all four surfaces (inline, file-based via the shared `_read_message_file`, out-of-band API, and the MCP MR tools). The API path validates only the fields the command actually sets and skips a partial edit that touches neither title nor description (never-lockout); a single-field edit mirrors the set field onto the untouched one so it cannot false-block on an empty title.

souliane:
Closing: done via PR #1705 on main. `hook_router._extract_mr_fields` (hooks/scripts/hook_router.py) now validates the MR title/description across all four surfaces — inline `--title/--description`, file-based/heredoc/$(...) via the shared `_read_message_file`, out-of-band REST API writes (`--field title=`/`-f body=`), and the MCP MR tools — with the never-lockout carve-outs described in the issue.

PR #1705, opened 2026-06-01
fix(hooks): gate MR title/description on file-based and out-of-band-API edits (#1703)

adrien-oper:
## What

Extends the `validate-mr-metadata` PreToolUse gate (`hook_router._extract_mr_fields`) so a non-compliant remote MR title/description is blocked at the tool-call layer regardless of HOW it is set, not only when it appears inline in a `glab mr create --description 'x'` command string.

Now covered, all four surfaces:

1. inline `glab mr create/update --title/--description` (unchanged);
2. the same command's **file-based / heredoc / `$(...)`** description (`-F`, `--description-file`) — read via the shared `_read_message_file` helper instead of being passed through as a falsely-empty (and trivially "valid") string;
3. **out-of-band REST API writes** (`glab api`/`gh api` PUT/POST to a `.../merge_requests` or `.../pulls` endpoint with `--field title=`/`-f description=`/`-f body=`) — the web-UI-equivalent edit that bypasses the create/update CLI entirely;
4. the `mcp__glab__glab_mr_*` MCP tools (unchanged).

## Why

A non-compliant description set through a surface other than the inline create/update command reached the remote unvalidated and then failed the remote MR-title/description CI job, blocking the merge after the fact — the exact failure this local gate exists to prevent. See [#1703](https://github.com/souliane/teatree/issues/1703).

## Never-lockout (the over-block dimension)

The out-of-band API path validates **only the fields the command actually sets**:

- a partial edit touching neither title nor description (e.g. `--field state_event=close`) is **skipped** — never force-validated against an empty description;
- a **single-field** edit mirrors the set field onto the untouched one. A valid `type(scope): … (ticket_url)` line is, by the canonical grammar, simultaneously a valid title and a valid description first line, so mirroring can never inject a spurious failure for the untouched field — a description-only edit cannot false-block on `Title is empty.`;
- a missing/unreadable description file falls back to `""` (the validator gives the correct empty-first-line verdict) rather than crashing.

Existing escapes are unchanged: `T3_MR_VALIDATE_ALLOW_BROKEN_ENV=1` (per-gate self-rescue on a broken validator env) and `T3_MR_VALIDATE_SCRIPT` (explicit validator override). The gate also still routes through the master `gate_fail_open` switch and the repeated-denial circuit breaker.

## Tests

New `TestFileBasedDescriptionIsRead` and `TestOutOfBandApiEditIsGated` classes (file-based read, API bad-edit blocks, API good single-field edit passes, state-only edit skipped, GET read skipped, `gh api` PR create gated). Full local suite green: 590 gate / lockout / fail-open / hook-router / ai-sig tests pass, ruff + ty clean.


adrien-oper:
Recovered from a prior session whose background push died before completing. Branch tip refreshed: merged latest `origin/main` (clean, no conflicts) and re-pushed, so the PR is no longer behind on the gate files. GitHub Actions `test (3.13)` is green on the merged tree.

**Do not self-merge — this is a gate PR. Independent cold-review must cover the over-block / never-lockout dimension symmetrically with the bypass dimension.**

Verified locally on this tip (`_extract_mr_fields` + `handle_validate_mr_metadata`):

Must-ALLOW (never-lockout):
- compliant `glab mr create --title/--description` with a valid `type(scope): … (ticket_url)` line -> validator passes -> gate returns allow
- `git status`, `git push origin <branch>`, `glab api <GET read>`, `glab api --method PUT … --field state_event=close` -> `_extract_mr_fields` returns `None` -> never validated, never blocked
- broken validator env -> `T3_MR_VALIDATE_ALLOW_BROKEN_ENV=1` self-rescue, `T3_MR_VALIDATE_SCRIPT` override, and the `_fail_open_or_deny` master `gate_fail_open` switch + self-rescue allowlist all relax it

Must-DENY (the gap this closes):
- file-based / heredoc / `$(...)` description whose first line is non-compliant (`-F`/`--description-file`) -> read and validated, not passed through as falsely-empty
- out-of-band `glab api`/`gh api` PUT/POST to `.../merge_requests`/`.../pulls` with a non-compliant `--field description=`/`-f body=` -> blocked

Suggested reviewer focus: confirm the single-field mirror (`type(scope): … (ticket_url)` is simultaneously a valid title and valid description first line) cannot inject a spurious `Title is empty.` on a description-only edit, and that the `_API_FIELD_RE` shapes (whole-token-quoted / value-quoted / bare) don't over-capture a non-MR `--field` on an unrelated `api` call.

PR #1768, opened 2026-06-02
feat(loop): CI-green fail-closed gate + deferred reinstall for self-update (#1760)

### LINKED ISSUE #1760: feat(loop): CI-gated auto-reinstall for the self-update scanner — apply ff-pulls safely, never on red main
## Problem

The self-update scanner (`src/teatree/loop/scanners/self_update.py`) fast-forwards the editable teatree clone and overlay clones every tick once the cadence elapses, gated only on clean-tree / on-default-branch / fast-forward-able. Two gaps remain:

1. It pulls regardless of whether the default branch's CI is green. A fast-forward pull onto a red default branch drags broken code into the running orchestrator.
2. The pulled code never actually takes effect: the running editable install keeps importing the old code until a human runs the heavy reinstall command. The pull is silently inert.

## Proposal

Extend the existing scanner (do not add a new scanner):

1. CI-green fail-closed gate (always on): before applying a fast-forward pull, confirm the default branch's CI is green. Only an explicit green proceeds; red / pending / unknown / offline / non-GitHub-origin all skip the pull. CI is queried only when origin is actually ahead.
2. Deferred opt-in auto-reinstall (default OFF): on an actual update, record a pending-reinstall row. A fresh per-tick subprocess drains it as its first step (before any scanner code imports) so the reinstall + setup + self-DB migrate runs in a clean process with no mixed-code window, and it defers whenever a loop unit is in flight.

## Safety guarantees

- Existing (kept): dirty tracked tree, non-default-branch checkout, non-fast-forwardable, fetch failure all skip.
- New: default-branch CI red / pending / unknown skip the pull (fail closed: only an explicit green proceeds).
- New: never reinstall mid-tick or while a sub-agent / loop unit is active.
- New: auto-reinstall is opt-in, default OFF, so the genuinely new side-effect on the running orchestrator is never enabled silently.

## Tests

Each gate is observed to fail closed with real `tmp_path` git clones plus an injected CI-status adapter and a fake subprocess runner: red/pending/unknown block the pull; green proceeds; CI is not queried when up-to-date; the deferred row is written on update only when the flag is on; the drain runs the exact reinstall argv then setup + migrate, defers when a unit is in flight, and marks failed on error. Config defaults and the env override are covered, the per-tick command is shown to drain before scanners, and the statusline renders the new skip note while still dropping the up-to-date noise.

souliane:
## What

The self-update scanner now refuses a fast-forward pull unless the default branch's CI is explicitly green, and can re-anchor the running editable install after an update without stealing the foreground mid-tick.

## Changes

- **CI-green fail-closed gate** (`auto_update_require_green_main`, default ON): before a ff-pull, query the default branch's `check-runs` via the new `GhMainCiStatus` (the same `test (3.13)` gate the PR sweep uses). Only an explicit GREEN proceeds; red / pending / unknown / non-GitHub origin / offline all skip. CI is queried only once origin is detected ahead of HEAD (no remote call on the up-to-date common path).
- **Deferred opt-in reinstall** (`auto_update_reinstall`, default OFF; `T3_LOOP_AUTO_UPDATE` env wins): on an actual update the scanner upserts a `PendingReinstall` row. `loop_tick` drains it as its very first step — a fresh per-tick subprocess, before any scanner imports, so there is no mixed-code window — and DEFERS while a loop unit is in flight (a live CLAIMED lease), never mutating code out from under an active sub-agent.
- **New `teatree.self_update` module** holds the shared editable-reinstall + runtime self-DB migrate so `teatree.cli` and `teatree.loop` run the identical sequence without `teatree.loop` importing `teatree.cli` (respects the tach module boundary).
- **Statusline observability**: the CI-held skip surfaces in `action_needed`; up-to-date / cadence / advanced outcomes stay dropped as noise.

## Safety guarantees

- Kept: dirty tracked tree, non-default-branch checkout, non-fast-forwardable, fetch failure all skip.
- New: default-branch CI red / pending / unknown skip the pull (fail closed — only an explicit green proceeds).
- New: never reinstall mid-tick or while a loop unit is active.
- New: auto-reinstall is opt-in, default OFF — the new side-effect on the running orchestrator is never enabled silently.

## Tests

Each fail-closed gate is observed to skip with real `tmp_path` git clones + an injected CI adapter + a fake subprocess runner: red/pending/unknown block the pull, green proceeds, CI is not queried when up-to-date or when the dirty/branch gate already skipped; the deferred row is written on update only when the flag is on; the drain defers when a unit is in flight and marks done/failed otherwise; the per-tick command drains before the scanner pass; the statusline renders the CI skip while dropping the rest. New lines are at 100% coverage; the full suite is green (11415 passed, 94.26% total coverage).

Closes #1760


souliane:
CI note: the `lint` check fails on `src/teatree/backends/slack_bot.py:122` with `PLR0904 Too many public methods (27 > 25)`. This file is not touched by this PR — the violation was introduced on `main` by [#1759](https://github.com/souliane/teatree/pull/1759) (`feat(slack): t3 setup slack-provision`), which added a 27th public method to `SlackBotBackend` (the `PLR0904` ignore only covers `tests/**`, not `src/`). It is the shared lint failure hitting all PRs branched off the current main, not a regression from this change.

Every other check passes, including `test (3.13)` (7m23s) and the full Docker pre-push matrix. `ruff check` is clean across all files this PR actually modifies. This PR should go green on lint once main is fixed.

PR #2043, opened 2026-06-06
chore(deps): django 6.0.6 — clear PYSEC-2026-197..201

souliane:
## What

Refresh `uv.lock` so `django` resolves to **6.0.6**, clearing the 5 freshly-disclosed CVEs (`PYSEC-2026-197` through `PYSEC-2026-201`) flagged by the `uv-audit` job (`pip-audit --strict --vulnerability-service osv`).

## Root cause

django 6.0.5 carries PYSEC-2026-197..201 (all fixed in 6.0.6). The lock pinned 6.0.5, so `pip-audit` exits 1.

## Fix

`uv lock --upgrade-package django`. The constraint already allows it (`django>=5.2,<6.1`), so **no `pyproject.toml` change** — single-package lock delta (6.0.5 -> 6.0.6).

## Local verification

- before: `pip-audit ... -r requirements.audit.txt` -> `Found 5 known vulnerabilities in 1 package` (django 6.0.5, PYSEC-2026-197..201, Fixed in 6.0.6), exit 1.
- after: `No known vulnerabilities found`, exit 0.

prek scoped to the lock change passes (tach, import-linter, ty-check, banned-terms).

## Architecture pre-check

Tactical dependency-pin fix (lockfile only, no `src/` change) — the nine-check gate does not apply.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -473,16 +473,16 @@ wheels = [
 
 [[package]]
 name = "django"
-version = "6.0.5"
+version = "6.0.6"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "asgiref" },
     { name = "sqlparse" },
     { name = "tzdata", marker = "sys_platform == 'win32'" },
 ]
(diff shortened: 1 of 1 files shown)


souliane:
Bundled into #2041 to clear two simultaneously-red main checks (BLUEPRINT byte budget + uv-audit) — neither PR could pass CI alone.

PR #2141, opened 2026-06-08
fix(ship): fail loud when PR URL targets the wrong repo

souliane:
## What

After the existing empty-URL guard (landed in #1261 for #1222/#1226), the ship runner did not check whether the URL returned by `host.create_pr` actually pointed at the expected repo. A syntactically-valid URL targeting a different repo would advance the FSM to `IN_REVIEW` and record the wrong PR URL.

This adds the missing check in `ShipExecutor._open_pr_and_record`: the repo slug is derived from `git.remote_slug(spec.repo)` and verified to appear in the returned URL. When it does not, the method returns `RunnerResult(ok=False)` so the FSM stays in `SHIPPED` and no `pr_urls` entry is recorded.

## Test plan

- `TestShipPrUrlRepoMismatch::test_returns_failure_when_pr_url_targets_wrong_repo` — regression guard; was RED on the pre-fix code (runner returned `ok=True` and logged the wrong-repo URL), GREEN after the fix.
- `TestShipPrUrlRepoMismatch::test_returns_success_when_pr_url_matches_expected_repo` — happy path; matching slug proceeds normally.
- All 59 existing tests in `tests/teatree_core/test_runners_ship.py` remain GREEN.
- All 36 tests in `tests/teatree_core/test_tasks.py` remain GREEN.
- Pre-commit hooks (ruff, tach, ty-check, import-linter, all fitness functions) passed.
- No migrations added.

Part of #1120


souliane:
Note: this PR is cold-review approved with `test (3.13)` and `mutation-diff` green, but it happens to be the first PR of the ISO week, so the weekly `mutation-full` job actually runs — and that job is deterministically red for a reason unrelated to this diff (it mutates all 8 safety modules but the baseline only records 2). Tracked in #2142. This PR should merge once #2142 is resolved; its own change is clean.

Issue #29, opened 2026-03-24
CLI parity with dashboard: deterministic subcommands for all dashboard actions


souliane:
## Summary

Every action available from the dashboard must have a corresponding `t3` CLI subcommand. The CLI must work reliably when the dashboard is down or when we're rate-limited and want to work manually.

### 1. CLI subcommands mirroring dashboard actions

Clearly organized subcommands covering at minimum:

- **Agent / session management**
  - Open agent with overlay context
  - Start auto session for ticket
  - Start interactive session for ticket
  - Cancel running tasks
  - View session history

- **Review & MR workflow**
  - List review comments
  - Transition MR status

- **Ticket & task management**
  - List ticket tasks with details
  - Show recent activity

- **Workspace lifecycle**
  - Idempotent reset/repair ticket directory and worktrees
  - Start dev servers

### 2. Idempotency

Most commands affecting state **must be idempotent** — running them twice produces the same result. Exceptions are acceptable when justified (e.g., auto session continues with the next step on re-run). Each command's idempotency contract must be documented and tested.

### 3. Deterministic workspace reset (critical path)

The most important command — **idempotent reset/repair of ticket directory, worktrees, and server startup** — must:

- Be **100% deterministic Python code** (no agent involvement)
- Work without any LLM calls
- Be the go-to command when things are broken

### 4. Testing

- **Unit/integration tests with pytest** for all subcommands
- **Full integration test for workspace reset/repair**: mock only network calls, run in a `tmp` directory to create the ticket dir, worktrees, etc. — verify the entire flow end-to-end
- Idempotency must be explicitly tested (run command twice, assert same state)

## Motivation

The dashboard is a convenience layer, but the CLI is the source of truth. If the dashboard is down, rate-limited, or the user prefers terminal workflows, every action must be reachable via `t3 <subcommand>`. Today some actions are dashboard-only, which creates a single point of failure.

PR #2177, opened 2026-06-08
Test-mirroring full sweep (#29: full sweep)

souliane:
## What

Advances the test-path-mirror "full sweep" (#29): relocates **51 loose-at-`tests/`-root test files** into their single unambiguous mirror directory (`tests/teatree_<pkg>/...`), and ratchets `[tool.teatree.test_path_mirror].baseline` **264 → 213**.

This is a pure relocation — no test logic changes — so coverage is unchanged. Each moved file had **exactly one** expected mirror dir (an unambiguous loose-at-root violation). Ambiguous multi-target files and broad package-root integration tests are deliberately left for follow-ups; they need per-file judgment (which package they primarily exercise) or a `# test-path: cross-cutting` pragma, not a blind move.

## Mechanical adjustments

Deeper placement required fixing `__file__`-relative repo-root resolution in 5 files (`parents[N]` / `.parent.parent` depth bumps). `test_overlay_db_import_contract.py` is kept at the root on purpose: its `cls.__module__.startswith("tests.")` fixture filter only works while the file imports as `tests.X`; moving it into a non-package dir gives it a bare module name under pytest's prepend import mode, so its own deliberately-broken fixture overlay leaks into the contract scan and the test fails.

## Anti-vacuity (RED proof)

The guarded behavior is `tests/teatree_quality/test_test_path_mirror.py::TestLiveTree::test_live_count_within_baseline` at the new floor of 213. Reverting any one relocation raises the live count to 214 > 213 and the gate goes RED:

```
tests/teatree_quality/test_test_path_mirror.py:99: AssertionError
FAILED ...::TestLiveTree::test_live_count_within_baseline
```

Restoring the move returns it to GREEN (live 213 == baseline 213). So lowering the floor genuinely depends on the relocation holding — the ratchet is not vacuous.

## Gates

- `ruff check` / `ruff format --check`: clean
- `ty check` on edited files: clean
- `t3 tool test-path-mirror`: ratchet holds at 213
- `tests/teatree_quality/test_test_path_mirror.py` + module doctests: 49 passed
- Relocated batch (51 files) + contract test: 1551 passed
- `tests/quality/test_patch_targets_resolve.py`: 20 passed
- Full suite with coverage: see CI

- **Scope**: "full sweep" is delivered as the unambiguous loose-at-root batch (51 files). The remaining ~213 violations (99 multi-target + broad package-root integration tests) need per-file judgment and are out of scope for one reviewable PR; the baseline ratchets toward 0 across follow-ups, exactly as the BLUEPRINT describes.
- **Guard coverage**: the existing path-mirror guard already scans every `tests/**.py`, so it covers the whole tree for *mis-pathed* files. The gap it does NOT cover is *missing* mirrors (src modules with no test at all) — that is a separate fitness function, not changed here.

DO NOT MERGE — review only.


souliane:
CI note: the one red test (tests/eval/test_scenarios_anti_vacuous.py::test_every_behavioral_scenario_ships_a_fail_fixture, scenario e2e_evidence_one_note_per_ticket_via_manifest) is pre-existing on main and unrelated to this PR. This PR is a pure test-file relocation and touches no eval scenarios/fixtures; the missing *_fail.stream.jsonl fixture is absent on origin/main too, so main is red on this gate independently. CI shows 1 failed, 6431 passed. All gates pertaining to this change are green: test-path-mirror, lint, ty, the relocated 51-file batch (1551 passed locally), and the patch-targets gate. Coverage is structurally unchanged (no src edits; files only moved).

PR #2188, opened 2026-06-08
perf(eval): cut metered-suite token cost via section-scoped system prompts (+--parallel)

souliane:
## Problem

The metered eval lane (`t3 eval run --backend sdk`) shells `claude -p` per
scenario and resends the **whole** `agent_path` SKILL.md as `--system-prompt`
every time, with no cross-scenario cache. The dominant input-token cost of a
suite run is that pile of whole-file prompts.

**Measured baseline (158 scenarios):** system-prompt input ≈ **1.63 M tokens**
(6.55 MB of SKILL.md text). `skills/rules/SKILL.md` (77 KB) alone is ≈ **833 K
tokens / 51%** — it's resent for ~40 scenarios that each pin ONE of its ~50 rules.

## Approach

**Lever 1 — section-scoped system prompts (the token win).** A scenario declares
`agent_sections: ["<## heading>", …]`; the runner sends only those `## ` sections
(verbatim, plus the file preamble) instead of the whole file. The section IS the
rule under test, so this is faithful. Empty (default) sends the whole file —
existing scenarios byte-for-byte unchanged.

- **Measured result: 32 scenarios scoped → whole-suite system-prompt input cut
  ~36% (≈585 K tokens; sonnet-priced input bound $4.88 → $3.13 per single pass).**

**Lever 2 — prompt caching: investigated, NOT available on this path.** `claude
-p` exposes `--exclude-dynamic-system-prompt-sections` for cross-call cache
reuse, but it is explicitly *ignored with `--system-prompt`* (the flag the runner
uses to inject the SKILL.md). No honored prefix-cache knob exists for our path,
so **no cross-scenario cache saving is claimed** — the win is the smaller prompt.

**Lever 3 — tight `max_turns`: a trap on this codebase, left alone.** `error_max_turns`
maps to `is_error=True` → scenario FAIL regardless of matchers, so capping a
single-action scenario to `max_turns: 1` turns a correct first action into a
config-fail. Not applied. (Surfaced as a finding below.)

**Lever 4 — `--parallel N` (wall-clock only).** Each `claude -p` is I/O-bound; a
bounded thread pool (`teatree.eval.parallel.run_specs`, cap 20) runs N scenarios
concurrently, order-preserved. Default `--parallel 1` = today's sequential order.

## Files

- `src/teatree/eval/context_budget.py` — `extract_sections` + `MissingSectionError`.
- `src/teatree/eval/parallel.py` — bounded concurrent driver.
- `src/teatree/eval/models.py` / `loader.py` — `EvalSpec.agent_sections` field + parse.
- `src/teatree/eval/runner.py` — one-line section-aware system-prompt load.
- `src/teatree/cli/eval/{app,all}.py` + `run_docker.py` — `--parallel` wiring (run_docker
  split out to keep app.py under the module-health LOC cap).
- `scripts/eval/corpus_gen/{model,all_scenarios}.py` — generator emits + self-checks
  `agent_sections` for generated scenarios.
- scenario YAML + `skills/rules/evals.yaml` — scoped scenarios.

## Anti-vacuity strategy

- A missing/typo'd section raises `MissingSectionError` — never a silent empty
  prompt (the vacuity trap). Proven: neuter the raise → the guard tests go RED.
- `runner.py` section-load proven anti-vacuous: revert the `extract_sections`
  call → `test_agent_sections_send_only_the_named_section_as_system_prompt` RED.
- `test_scenarios_anti_vacuous.py::test_every_declared_agent_section_resolves_against_its_skill`
  resolves every declared section against the real SKILL.md on every PR; the
  corpus generator self-checks its `_AGENT_SECTIONS` map at generation time.
- All 1145 eval + CLI + patch-target tests green; ruff/ty/tach clean.

## Risks

- A future scenario could map to a `## ` heading that *exists* but does not
  contain the tested rule — the resolution guard checks existence, not semantics.
  Mitigated by scoping only clean single-rule mappings and leaving multi-rule /
  ambiguous scenarios on the full file.

## Dependency

The metered run itself needs the `--bare`-removal auth fix
(`ac/eval-sdk-docker-auth`, PR pending) to authenticate; this PR is based on main
and edits a disjoint region of `runner.py`, so it integrates cleanly.

## Finding (separate)

`error_max_turns` → `is_error=True` → scenario FAIL even when the matcher is
satisfied on turn 1. This makes a tight `max_turns` cap actively harmful and is
why lever 3 was not applied. A grading-semantics change (treat a satisfied
matcher under `error_max_turns` as PASS) would unlock that lever — out of scope
here.

## Architecture pre-check

1. **BLUEPRINT §** — eval harness line (459) + README "How it runs" / backends;
   updated in this PR.
2. **FSM phase boundaries** — n/a (no transition).
3. **Extension-point contracts** — `EvalSpec` (additive field), `EvalRunner`
   Protocol (unchanged); generator `Scenario` (additive field).
4. **Component boundaries** — `eval/context_budget.py`, `eval/parallel.py`
   (leaves); `cli/eval/run_docker.py` (typer-adjacent, no business logic).
5. **Dependency direction** — leaves; `tach check` green.
6. **Test surface** — context_budget / parallel / runner / loader / corpus-gen /
   anti-vacuous-resolution / CLI `--parallel` tests, each named above.
7. **Resilience** — `extract_sections` is a pure idempotent read; fails LOUD on a
   missing section.
8. **Identity/key normalization** — n/a.
9. **Behavior preservation** — purely additive; empty `agent_sections` /
   `--parallel 1` reproduce today's behaviour exactly; no matcher/test inverted.


souliane:
Dependency note: the metered run itself needs the auth fix in [#2185](https://github.com/souliane/teatree/pull/2185) (drop `--bare`) to authenticate `claude -p`. This PR is based on main and edits a disjoint region of `runner.py`, so it integrates cleanly regardless of merge order. The token-cost reduction here is a deterministic property of the bytes sent (the built `--system-prompt`), independently verified without a metered call.

PR #2233, opened 2026-06-10
fix(review-request): fall back to draft DM when review channel is unpostable

### LINKED ISSUE #2231: review-request post: fall back to draft DM when review channel is unpostable (Connect channel)
## Problem

When `t3 <overlay> review-request post` is called for an overlay whose review channel is a Slack Connect channel, `resolve_guard_target()` returns `None` because no xoxp user token is available (the Slack MCP refuses Connect channels). The command then silently suppresses with `no_review_channel_or_token` — the human review notification never reaches anyone.

## Expected behaviour

When `resolve_guard_target()` returns `None`, instead of silently suppressing, the command should:

1. DM the bot-user a draft containing the review request text and MR URL, so the user can manually forward it to the review channel.
2. Emit `{"action": "draft", "reason": "no_review_channel_or_token", "mr_url": ...}` (exit 0) so callers can distinguish "draft fallback" from "already posted" suppression.

## Context

See `feedback_gated_action_ask_via_tool_never_copy_paste.md` (binding rule) — sanctioned path is a Slack DRAFT (bot→user DM) the user can send manually.

Related: session-TODO #290.

souliane:
## Summary

- When `resolve_guard_target()` returns `None` (e.g. a Slack Connect channel that needs a user xoxp token the bot doesn't hold), the command previously silently suppressed with `no_review_channel_or_token` — the human review notification was never sent.
- Now it falls back to a bot→user DM containing the full review request text so the user can forward it manually to the review channel.
- Emits `{"action": "draft", "reason": "no_review_channel_or_token"}` instead of `{"action": "suppress"}` so callers can distinguish "draft fallback" from "already posted" dedup suppression.

Closes #2231.

## Test plan

- [x] `TestReviewRequestPostDedup::test_no_review_channel_or_token_falls_back_to_draft_dm` — new test observed RED before the fix, asserts `action=draft` and exactly one DM containing the MR URL
- [x] All 16 tests in `tests/teatree_core/test_review_request_post_command.py` pass
- [x] All pre-commit hooks pass (ruff, tach, ty-check, banned-terms, …)

### DIFF
--- a/src/teatree/core/management/commands/review_request_post.py
+++ b/src/teatree/core/management/commands/review_request_post.py
@@ -76,9 +76,9 @@ class Command(TyperCommand):
         """Post a review request after #1829 anti-vacuity + #1094 dedup + #960 approval.
 
         Machine-legible: prints a single JSON dict (``action`` is
-        ``post``/``suppress``/``refused``) and uses exit codes — ``0``
-        post/suppress, ``2`` refused (no recorded approval / no anti-vacuity
-        attestation).
+        ``post``/``draft``/``suppress``/``refused``) and uses exit codes —
+        ``0`` post/draft/suppress, ``2`` refused (no recorded approval / no
+        anti-vacuity attestation).
         """
         _ = approver  # the #960 approver is bound at approve-on-behalf record time.
 
@@ -100,9 +100,13 @@ class Command(TyperCommand):
             # that requires a user xoxp token the bot doesn't hold — #2231).
             # Fall back to a bot→user DM draft so the user can forward the
             # review request manually. Never silently suppress.
-            self._draft_dm_fallback(mr_url=mr_url, title=title)
+            sent = self._draft_dm_fallback(mr_url=mr_url, title=title)
             self._emit(
-                {"action": "draft", "reason": "no_review_channel_or_token", "mr_url": mr_url},
+                {
+                    "action": "draft" if sent else "suppress",
+                    "reason": "no_review_channel_or_token",
+                    "mr_url": mr_url,
+                },
                 exit_code=0,
             )
 
@@ -192,7 +196,7 @@ class Command(TyperCommand):
         )
 
     @staticmethod
-    def _draft_dm_fallback(*, mr_url: str, title: str) -> None:
+    def _draft_dm_fallback(*, mr_url: str, title: str) -> bool:
         """DM the user a draft review-request when the review channel is unpostable.
 
         Called when ``resolve_guard_target`` returns ``None`` — the review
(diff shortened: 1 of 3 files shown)


src/teatree/core/management/commands/review_request_post.py
```diff
@@ -181,6 +186,30 @@ def handle(
...
+        """
+        from teatree.core.notify import NotifyKind, notify_user  # noqa: PLC0415
+
+        subject = title or _DEFAULT_TITLE
+        text = (
+            f"The review channel is unpostable (no token — Slack Connect channel?). "
+            f"Please forward this review request manually:\n\n"
+            f"{subject} {mr_url}"
+        )
+        notify_user(
```
souliane: **(medium)** `notify_user` returns a `bool` for whether the DM actually landed, and here the return value is dropped — `handle` unconditionally emits `action=draft`. When no messaging backend or `slack_user_id` is configured, `notify_user` records a NOOP and returns `False`, yet the caller still tells the user a draft was sent. That re-creates the silent-loss #2231 targets ("the human review notification never reaches anyone").

Suggest having `_draft_dm_fallback` return the `notify_user` bool and emitting a distinct outcome when it didn't land:

```python
sent = self._draft_dm_fallback(mr_url=mr_url, title=title)
self._emit(
    {"action": "draft" if sent else "suppress",
     "reason": "no_review_channel_or_token", "mr_url": mr_url},
    exit_code=0,
)
```

so a genuinely-undeliverable fallback stays loud instead of masquerading as a draft.

src/teatree/core/management/commands/review_request_post.py
```diff
@@ -91,8 +91,13 @@ def handle(
...
         target = resolve_guard_target()
         if target is None:
+            # The review channel is unpostable (e.g. a Slack Connect channel
+            # that requires a user xoxp token the bot doesn't hold — #2231).
+            # Fall back to a bot→user DM draft so the user can forward the
+            # review request manually. Never silently suppress.
+            self._draft_dm_fallback(mr_url=mr_url, title=title)
             self._emit(
-                {"action": "suppress", "reason": "no_review_channel_or_token", "mr_url": mr_url},
+                {"action": "draft", "reason": "no_review_channel_or_token", "mr_url": mr_url},
```
souliane: Nit (low): the `handle` docstring just above (line ~73) still enumerates `action` as `post`/`suppress`/`refused`. Add the new `draft` value there so the machine-legible contract stays in sync with this emit.

src/teatree/core/management/commands/review_request_post.py
```diff
@@ -181,6 +186,30 @@ def handle(
...
+        subject = title or _DEFAULT_TITLE
+        text = (
+            f"The review channel is unpostable (no token — Slack Connect channel?). "
+            f"Please forward this review request manually:\n\n"
+            f"{subject} {mr_url}"
+        )
+        notify_user(
+            text,
+            kind=NotifyKind.INFO,
+            idempotency_key=f"review_request_draft:{mr_url}",
```
souliane: Nit (low): the idempotency key uses the raw `mr_url` param, whereas the rest of the command keys on `canonical_mr_url(mr_url)`. Two different URL spellings of the same MR would yield two draft DMs. Canonicalizing here (or computing `canonical` before the fallback) keeps the dedup key consistent with the guard/approval scope.

souliane (commented):
**Verdict: HOLD / changes requested.** (Posted as a comment-event review — GitHub blocks a formal REQUEST_CHANGES on one's own PR.)

Nice fix for the silent-suppress class — turning the unpostable-channel case into a bot to user draft DM is the right call. The RED test is genuine (reverted the production hunk and it fails with `assert 'suppress' == 'draft'`), and lint / ty-check / the patch-target resolution gate all pass.

Holding on one robustness gap (medium): the fallback ignores `notify_user`'s return value, so a failed DM still reports `action=draft`. In the no-backend / no-`slack_user_id` case `notify_user` returns `False` (records a NOOP) and the notification is silently lost again — the exact class #2231 set out to close. The primary Connect-channel path is fine (a bot backend is present there), but the alternate failure mode re-opens the hole. Two low-severity notes inline as well.

PR #2309, opened 2026-06-12
fix(lint): re-wrap docstring line widened by the leak scrub

souliane:
The force-push this morning collapsed a wrapped docstring sentence onto one line (137 chars, limit 120), producing a tree-wide E501 that blocks the lint CI check on every open PR — re-wrapping it across two lines restores compliance.


souliane:
Superseded: the scrub-collateral E501 was fixed directly on main as part of the history-scrub force-push (commit d190f4c7). No longer needed.

Issue #1796, opened 2026-06-02
Loop architecture: split/rename the monolithic 'tick' into named phases + make speed=full actually fan out parallel workers


souliane:
## Problem

`t3 loop tick` is an overloaded monolith and a meaningless name:

- One call does everything inline — scanners, PR sweep/auto-merge, mechanical task completion, reaction/transition posts, statusline render (`OK 403 signals, 392 actions` in a single `tick`).
- It does **not** orchestrate. With `speed = "full"` the loop does **not** autonomously fan out parallel implementation workers across the backlog — so "full speed" is effectively inert. Bucket-classify then parallel-worker fan-out only happens when a human runs `/t3:speed boost`, never from the loop itself.
- `tick` conflates three different jobs: *scan the world*, *sweep/merge ready PRs*, and *drive new work*. The name says nothing.

## Proposal (to design — see the plan)

1. Split into named single-responsibility phases, e.g.:
   - **scan** — read-only world state then signals
   - **sweep** — mechanical: auto-merge green solo PRs, complete orphaned tasks, render statusline
   - **orchestrate** — the missing piece: when `speed=full`, bucket-classify the backlog and fan out parallel autonomous-safe workers (orchestrate-only: spawn/route/collect, never implement inline), clamped to `max_concurrent_auto_starts`
2. Rename `tick` to something meaningful.
3. Wire `speed=full` to the **orchestrate** phase so the autonomous loop actually blasts the backlog in parallel.
4. Open question to answer: do we still need a single combined command, or are the phases the API?

Clean cutover (no deprecated aliases). A Plan is being produced with the concrete split, file moves, and migration steps.

souliane:
## Design input: can the orchestration leave the main agent? (spike)

Spiked the question "can we move loop/backlog orchestration off the main interactive agent — e.g. onto a sub-agent or a Claude Code Agent Team — so the top session stays free?"

**Conclusion: no, not with current Claude Code primitives — and that strengthens the case for this issue's `orchestrate` phase.**

The capability that makes an orchestrator an orchestrator (calling the `Agent` tool to fan out workers) exists **only in the top session**:

- Agent Teams docs: *"No nested teams — teammates cannot spawn their own teams or teammates. Only the lead can manage the team"* + *"Lead is fixed."*
- General rule: *"subagents cannot spawn other subagents."* In-process team agents lack the Agent tool ([anthropics/claude-code#31977](https://github.com/anthropics/claude-code/issues/31977)).
[...]

PR #2315, opened 2026-06-12
feat(loops): arm orchestrate-phase claim+spawn behind a default-off toggle (#1796)

souliane:
Agent-teams Track-A PR1 for issue 1838. Arms the already-computed issue-1796 orchestrate-phase fan-out manifest claim, behind a new opt-in, default-OFF teatree orchestrate_claim_enabled toggle. Ships inert: merging changes no live behaviour until the toggle is flipped on; the fat loop stays the default. The dispatch loop orchestrate_phase already computes a deterministic claim-capable manifest from speed and max_concurrent_auto_starts but is wired dormant in run_tick with claim=False. This flips to claim=True only when the toggle is on, so the lead does the thin per-unit claim and spawn the manifest already computes, via the existing claim_next_pending compare-and-swap. The toggle mirrors issue_implementer_enabled: a UserSettings bool read via get_effective_settings, registered per-overlay overridable and with a T3_ORCHESTRATE_CLAIM_ENABLED env override; the tick read degrades to OFF on any settings-read error. Tests: manifest golden table across speed x backlog x budget; toggle OFF claims nothing and ON claims the manifest rows on a tmp DB; config surface default-off and overrides; K-claimer x M-row concurrency on file-backed SQLite proving exactly M distinct claims and zero double-claims with an anti-vacuity RED case. Local gates green: pytest, ruff check, ruff format, ty check, tach check. Full architecture pre-check and open-questions are in the commit body. References the agent-teams plan Track-A PR1.


souliane:
Architecture pre-check (9 checks). 1 BLUEPRINT alignment: 5.6 Loop Topology / 17.4 orchestrator-decides-loop-executes; updated the 5.6 orchestrate_phase line for the dormant-by-default wiring and the toggle. 2 FSM phase boundaries: n/a, the claim is the issue-786 N4 spawn boundary on Task, PENDING to CLAIMED via the existing CAS, no FSM change. 3 Extension-point contracts: none, no OverlayBase/scanner/hook/Protocol change. 4 Component boundaries: config settings.py plus loader.py for the toggle field and load; loop tick.py for the one-line gated arm; no straddle. 5 Dependency direction: loop.tick to config.get_effective_settings is an existing allowed edge, tach clean. 6 Test surface: manifest golden table, toggle on/off, config surface, K-claimer x M-row concurrency; anti-vacuity verified by reverting the arm to claim=False, which turns the toggle-ON test RED while the toggle-OFF test stays GREEN. 7 Resilience invariants for the Task claim: verify-by-re-read, no-op-on-error fallback, idempotent CAS, existing lease/heartbeat, existing sub-agent return contract. 8 Identity/key normalization: n/a, claimed_by is the existing plain worker string. 9 Behavior preservation: additive plus a single gated arm, default-off preserves the dormant path exactly, no matcher narrowed, no must-block test inverted. Assumption: the plan named a [loops] table but the codebase reads every opt-in loop bool from the [teatree] table into a UserSettings field, so I mirrored the issue_implementer_enabled pattern as [teatree] orchestrate_claim_enabled; the key name is unchanged, only the table is [teatree].

Issue #119, opened 2026-04-02
lifecycle restart: stale processes + wrong serverUrl after git pull


souliane:
## Problem

Existing worktrees that worked yesterday require hours of manual troubleshooting after a `git pull` because:

1. **No `lifecycle restart` command existed** — only `start` (which didn't kill stale processes) and `clean` (which drops the DB)
2. **`lifecycle start` wasn't idempotent** — re-running failed because old processes still held ports
3. **`resolve_backend_port()` returned wrong port** — `worktree.ports` is empty in DB for older worktrees, so `patch-customer-json` wrote the wrong `serverUrl`

## Root Cause

- `_launch_app_processes` stored PIDs in `worktree.extra["pids"]`, but worktrees created by older versions or external tools had no stored PIDs
- `resolve_backend_port()` fell back to `DJANGO_RUNSERVER_PORT` env var instead of reading `BACKEND_PORT` from `.env.worktree`

### teatree core (commit 2270666)
- Added `lifecycle restart` command: kills stored PIDs, re-runs pre-run steps, launches fresh
- Made `lifecycle start` idempotent by calling `_kill_worktree_processes` before launching
- Falls back to `setup + start` when worktree is in `created` state

### t3-oper overlay (commit 53a1793)
- `resolve_backend_port()` now reads `BACKEND_PORT` from `.env.worktree` when DB `ports` field is empty

souliane:
## Relation to #65

This issue is a symptom of #65 (dynamic port allocation at start time). The fixes committed here (lifecycle restart + .env.worktree fallback) are band-aids. The proper fix is #65's proposal: allocate ports at `lifecycle start` time, not `setup` time.

## Full Analysis

### Current architecture (broken)

```
setup → allocate_ports → write .env.worktree → store ports in DB
                                                   ↓
start → read ports from DB → patch customer.json → launch processes
```

**Failure modes:**
1. DB `ports` is empty for older worktrees → `resolve_backend_port` falls back to wrong env var
2. Ports allocated during setup may be taken by the time start runs (hours/days later)
3. `lifecycle restart` after reboot finds old ports occupied → launch fails
4. No mechanism to kill processes on specific ports (only stored PIDs, which are stale after reboot)

### Proposed architecture (#65)

```
start → find_free_ports → write .env.worktree → patch customer.json → launch processes
                              ↓
                     kill anything on those ports first
```

**Key changes:**
- `setup` no longer allocates ports — only provisions DB, symlinks, env templates
[...]

PR #2433, opened 2026-06-15
fix(hooks): MR-metadata gate stops over-firing on metadata-only edits

souliane:
## What

The validate-mr-metadata `PreToolUse` gate demanded a conventional-commit title **and** a What/Why description on **every** MR metadata edit. A reviewer/label/assignee/state-only edit (which sets neither field) was wrongly blocked — the over-firing reported in this task.

## Why / fix

`extract_cli_mr_fields` (new) now governs the `glab mr` CLI surface:

- **update validates only the field(s) the command sets.** A metadata-only edit (no title, no description) returns `None` → skipped (never-lockout). An unset field is back-filled with a known-good placeholder so the combined validator's verdict reflects only the field under edit.
- **unresolvable dynamic values are skipped.** A double-quoted command-substitution / parameter-expansion / backtick value the hook can't resolve before shell expansion (the captured fragment isn't the real value) is skipped; the remote CI gate is the backstop.
- **create keeps the stricter both-fields contract** (#119) — an empty title/description on a create is exactly the bad metadata the gate must catch.

## Module health

The `glab mr` CLI title/description parsing moved out of `hook_router.py` into a bare sibling module `hooks/scripts/mr_cli_fields.py` (same pattern as `unknown_repo_push_gate`), keeping `hook_router` under its module-health cap.

## Tests

12 new tests (`TestUpdateValidatesOnlySetFields` ×6, `TestDynamicValueIsSkipped` ×6); full file 34 green. `teatree_quality` suite 64 green, `check_module_health.py --from-ref origin/main` exit 0, ruff clean.

## Out of scope (separate facet)

A different over-firing class also hit this session: the gate fires on a Bash command that merely **embeds** the literal `glab mr create`/`update` string (a heredoc body, a git commit message). Not addressed here to keep this PR focused; tracked separately.


souliane:
Consolidated into #2434 (stacked; #2434 carries this PR's full diff). Closing as superseded.

Issue #1415, opened 2026-05-27
feat(privacy): gate gh/glab issue+PR+comment posting with the banned-terms scan (commit hook misses non-commit writes)


souliane:
The `banned-terms` pre-commit hook (`scripts/hooks/check-banned-terms.sh --config ~/.teatree.toml`) only runs on `git commit`. It does NOT cover writes that reach public/external surfaces without a commit: `gh issue create` / `gh pr create` / `gh ... comment`, `glab` MR/issue notes, and chat posts. Those are exactly the surfaces where overlay-/customer-specific terms can leak on this PUBLIC repo — and they have, repeatedly.

#1213 added the pre-publish scanner for commit/push-side public-repo writes; this is its sibling for the **non-commit posting paths**.

### Proposal
A PreToolUse hook (registered in `hooks/hooks.json`, non-bypassable from sub-agent shells) that:
1. Matches Bash commands invoking `gh issue create|edit`, `gh pr create|edit`, `gh ... comment`, and `glab mr|issue ... note|create`.
2. Extracts the body payload (`--body`, `--body-file <path>`, `-d/--field` JSON).
3. Runs the existing `check-banned-terms.sh` against it using the `~/.teatree.toml` term list.
4. Refuses (blocks the tool call) with the matched term, unless an explicit `--allow-term`-style override is present.

Reuses the existing scanner + term list; adds zero new term config. Sibling of #1213.

Issue #1937, opened 2026-06-05
Banned-terms gate: surface-aware term tiers (private overlay repos vs public surfaces)


souliane:
## Problem

The posting gate (#1415) applies one flat term list to every surface. Posting a PR body to the PRIVATE overlay repo (the overlay org's own repo, where the overlay name and its module names ARE the subject matter) required 4 attempts today: the overlay name, a domain legal term, and an internal module name were each blocked in turn. The inline ALLOW env override doesn't satisfy the hook (by design), so the only path is lexical contortion — which degrades PR bodies for zero security gain on that surface.

## Fix

Two-dimensional policy: term tier x surface tier.
- Term tiers: CUSTOMER terms (bank/tenant names, customer-feature codewords) vs OVERLAY-INFRA terms (the overlay name, internal module/repo names).
- Surface tiers: PUBLIC (github public repos, dev.to, anything indexable) vs PRIVATE-INTERNAL (the overlay org's private repos).
- Matrix: CUSTOMER terms blocked on BOTH (unchanged — the <mr-ref>-class protection stays absolute). OVERLAY-INFRA terms blocked on PUBLIC only; allowed on PRIVATE-INTERNAL.
- Surface classification must be deterministic (resolve repo visibility via gh/glab API or a static allowlist of org/repo → tier), fail-closed to PUBLIC when unknown.

## Tests
- must-ALLOW: overlay name in a PR body on the private overlay repo (today's case).
- must-BLOCK: same body on a public repo; any customer term on any surface.
- unknown surface → treated as public (fail-closed).

souliane:
## Recurrence data point (2026-06-14)

Same session showed both sides of why the term-tier split this issue proposes is the right fix:

- The gate **correctly blocked a real leak** — an arch-review comment drafted for the public surface accidentally carried an overlay-repo section; the gate caught it and the section was removed. Working as intended on a public surface.
- Minutes later the gate **misfired on a private overlay-repo tracker** — a status comment on the overlay's own private issue, where the term is in-domain, was blocked and required the `ALLOW_BANNED_TERM=1` escape hatch to post. (Previously observed on private GitLab MRs; now confirmed on a private GitHub repo too — so the misfire is forge-independent.)

[...]

Issue #1780, opened 2026-06-02
t3 review post-comment positional note body fail-closes the publish gates (colleague-review lockout)


souliane:
## Summary

The sanctioned `t3 review post-comment <repo> <iid> "<note>" --file <path> --line <n>` CLI form is hard-denied by the pre-publish gates even when the body is perfectly clean. This wedges all colleague-review posting (and, transitively, `t3 review approve`, which needs a prior reviewing footprint) through the documented CLI path.

## Root cause

`teatree.hooks._command_parser._walk_command_segment` has no extraction rule for the positional `note` argument of `t3 review post-comment` / `post-draft-note`. The body of these subcommands is the third positional (`<repo> <iid> <note>`), not a `--body`/`-m` flag.

Two compounding effects:

1. The positional note is never extracted, so the real (clean) body is never scanned.
2. `--file` is in `_BODY_FILE_FLAG_NAMES`, so the parser tries to read the diff-anchor file path as a message body file. At PreToolUse time that file does not exist, and with `fail_closed_body_file=True` the parser injects the fail-closed sentinel. For these subcommands `--file`/`--line` are diff-anchor coordinates, NOT a body file.

Result: the bare-reference scanner sees the fail-closed sentinel and reports an unresolved body, so the bare-reference link gate denies; the same sentinel makes the banned-terms gate deny too. A clean colleague-review post is blocked at every publish gate — a gate-vs-CLI conflict and a total colleague-review lockout.

## Reproduction

```
t3 review post-comment some-repo NNNN "Nit: clean naming consistency note" --file foo.ts --line 5
```

is denied by the bare-reference link gate as an unresolved body, despite the body carrying no bare reference.

## Fix

Teach `_command_parser` to extract the positional `note` body of the `t3 review` positional-body subcommands (`post-comment`, `post-draft-note`) and to skip the `--file`/`--line` body-file handling for those segments, so a clean body is scanned normally instead of fail-closing. The gate's purpose is preserved: a genuinely bare-ref positional body is still denied.

## Relationship to the destination-aware link-gate work

The destination-aware bare-reference change (the `ac/destination-aware-link-gate` branch) only exempts the bare-reference gate for external-forge destinations; it leaves the banned-terms gate (which has no forge exemption) still fail-closing on the same sentinel. This fix addresses the shared root cause in the parser, so both gates scan the real positional body — complementary, not overlapping (different file).

souliane:
[...]

PR #2454, opened 2026-06-15
fix(hooks): fail closed on an unreadable commit/post body, private repo included (#1415)

souliane:
## What

Closes the unavailable-body fail-open in the banned-terms posting gate (#1415).

When `scan_text` could not read a publish body, it returned one of two
unreadable-body markers (`UNRESOLVABLE_BODY_MARKER` for a missing/unreadable
`--body-file`/`-F` file, `UNAVAILABLE_BODY_SOURCE_MARKER` for an unexpanded
`$VAR` or a stdin body). The router (`_banned_term_marker_blocks`) then
**downgraded those to a warn whenever the destination resolved to a private
repo** — only the scanner-unavailable marker hard-blocked everywhere.

This change makes all three markers hard-block on every surface, a private
repo included.

## Why

The downgrade waved an *unverifiable* body through on the strength of a
destination guess. The "private" classification is itself a guess in-hook (a
`gh`/`glab` probe resolved under different auth, a stale `private_repos`
allowlist), so a body the gate could not read could still slip a leak through.
Failing closed here matches the scanner module's own documented contract (both
unreadable-body markers are already described there as blocking) and the
long-standing scanner-unavailable behaviour, so all three markers now behave
the same.

This reverses the private-repo downgrade introduced for the unreadable-body
case (PR #2286). Flagging that explicitly so the design call gets a second
look: the trade-off is that a genuinely-unreadable body on a private repo now
blocks instead of warning.

### Never-lockout preserved

The escapes are unchanged: the per-call `--allow-banned-term` /
`ALLOW_BANNED_TERM=1` override and the `banned_terms_gate_enabled = false`
kill-switch. In practice the body-file resolution retry (#2284) means a normal
commit's `-F` file resolves and scans fine; the marker only fires when the body
is genuinely unreadable or comes from `$VAR`/stdin — and the deny message tells
the operator to pass the body inline (`-m`) or as a readable `--body-file
<abspath>`. The readable-body carve-out for a private repo's own domain words
is untouched (it lives on the real-banned-term path, not the marker path), and
the `gh`/`glab` private-destination *skip* is also untouched, so normal private
posts are not over-blocked.

## Tests

- `test_banned_terms_hook.py::test_live_hook_blocks_unreadable_commit_body_file_in_private_worktree`
  — must-BLOCK (was the must-allow downgrade test; flipped).
- `test_banned_terms_hook.py::test_live_hook_allows_readable_clean_commit_body_file_in_private_worktree`
  — new must-NOT-block for the normal/available case.
- `test_banned_terms_scanner.py::test_commit_bodyfile_genuinely_missing_on_private_repo_still_blocks`
  and `::test_prefix_segment_private_commit_unreadable_body_still_blocks` — the
  scanner-integration siblings, flipped to block.
- The public-repo and unknown-visibility anti-vacuity guards are unchanged and
  still block.

Full `tests/teatree_hooks/` suite green (1446 passed); `tests/teatree_quality/`
green (64 passed); `t3 tool verify-gates` exits 0 (all commit + push stages).

### DIFF
--- a/hooks/scripts/hook_router.py
+++ b/hooks/scripts/hook_router.py
@@ -3278,10 +3278,12 @@ def _banned_terms_gate_enabled() -> bool:
 
     Fails OPEN to enabled on a missing/broken config so the gate keeps its
     protective default; an explicit ``[teatree] banned_terms_gate_enabled =
-    false`` is the one-line kill-switch (NEVER-LOCKOUT) the user flips to
-    disable the gate while its body-resolution over-block (an allowlisted
-    private-repo commit hard-blocked because the body could not be read) is
-    fixed properly. See :func:`_teatree_bool_setting` for the shared semantics.
+    false`` is the one-line kill-switch (NEVER-LOCKOUT) the user flips if the
+    gate's fail-closed behaviour (e.g. a private-repo commit blocked because the
+    gate could not read its body, #1415) ever proves too aggressive for a
+    session. The per-call ``--allow-banned-term`` / ``ALLOW_BANNED_TERM=1``
+    override is the narrower escape. See :func:`_teatree_bool_setting` for the
+    shared semantics.
     """
     return _teatree_bool_setting("banned_terms_gate_enabled", default=True)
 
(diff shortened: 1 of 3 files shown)


souliane:
Closing — this tightened the unreadable-body gate in the wrong direction: it removed the private-repo downgrade, which worsens the over-blocking #1415 is actually about. Superseded by the real fix in flight: positional note-body resolution (#1780) + surface-aware term tiers so internal-remote brand words stop false-positiving (#1937).

Issue #1775, opened 2026-06-02
Move all config to the database (except bootstrap-readable settings)


souliane:
## Goal

Make the database the canonical source for all config, so it is managed and auditable in one place and changeable without editing files.

## Scope

Move `~/.teatree.toml` / env config into DB-backed models, **except** the irreducible bootstrap set that must be readable *before* Django starts:

- `DATABASE_URL` / data-dir / `DJANGO_SETTINGS_MODULE`
- the offline `private_repos` allowlist that the in-hook publish gate reads without Django

## First slice

[#1773](https://github.com/souliane/teatree/issues/1773) (`TrustedIdentity`) is the first concrete slice — it ships now with a config fallback and is **not blocked** on this umbrella.

## Pattern

Reuse the existing "canonical tier is the DB" approach (`MergeClear` / `DbApproval`) with a pre-migration / empty-table fallback to the current file/env source, so nothing regresses during the migration window.

souliane:
Closing as part of backlog consolidation: the umbrella goal is already implemented — the DB-vs-TOML hard partition (each setting one home, TOML ignored on read for DB-home keys), the ConfigSetting store, and the irreducible bootstrap carve-out are all in place.

PR #2459, opened 2026-06-15
fix(eval): isolate the substrate-floor helper from the DB+env config tiers

souliane:
## What

`_staged_overlay_autonomy` (the helper the pinned-regressions **substrate-merge human-authorize floor** check uses to pin an overlay's autonomy hermetically) only swapped the **file tier** (`CONFIG_PATH`). The #1775 / #2450 **DB config override tier** — and the env tier — win over the file tier in `get_effective_settings`, so a host carrying a per-overlay `autonomy` `ConfigSetting` row (e.g. `t3-teatree` pinned to `full`) leaked past the staged `babysit`. The floor check then read the host's live autonomy and the pre-push `eval-pinned-regressions` hook failed on that host.

The helper now neutralises the DB and env override tiers for the staged block, so the hermetic TOML is authoritative — matching its documented contract of determinism regardless of live host config.

## Why it only failed locally

CI runs against a fresh, override-free DB, so `_db_setting_overrides` returns `{}` there and the floor check has always passed in CI. The failure was confined to developer hosts that have set a per-overlay `autonomy` override via `t3 teatree config_setting set` — exactly what the per-overlay settings work (#2450) made possible. This is the host-config-leak class flagged during that migration.

## Test

`TestStagedAutonomyHermeticAgainstDbTier` seeds a live `autonomy=full` `ConfigSetting` row and asserts both the staged-`babysit` resolution and the substrate floor predicate hold. Both go **RED on the pre-fix helper** (the seeded `full` leaks past the staged `babysit`) and green with the fix — anti-vacuous, verified by stashing the source fix and observing the failure.

docs: n/a — eval-isolation infra, no user-visible change

- none

### DIFF
--- a/src/teatree/eval/regression_corpus_predicates.py
+++ b/src/teatree/eval/regression_corpus_predicates.py
@@ -34,25 +34,41 @@ _SHA_B = "b" * 40
 
 @contextmanager
 def _staged_overlay_autonomy(overlay_name: str, autonomy: str) -> Iterator[None]:
-    """Run the block with a hermetic ``~/.teatree.toml`` pinning *overlay_name* to *autonomy*.
+    """Run the block with a hermetic config pinning *overlay_name* to *autonomy*.
 
     An eval-isolation helper for assertions whose outcome depends on the
-    overlay's effective autonomy (the substrate-merge carve-out). Swaps
-    ``teatree.config.CONFIG_PATH`` at the single seam ``get_effective_settings``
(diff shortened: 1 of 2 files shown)


souliane:
Superseded by #2460, which bundles this substrate-floor fix with the coverage and uv-audit fixes so the whole merge queue unblocks in one mergeable PR (each fix alone still fails the other required check).

Issue #2663, opened 2026-06-23
📓 Daily Dreaming — standing dream-pass outcomes ledger (DO NOT CLOSE)


souliane:
## ⚠️ DO NOT CLOSE — standing ledger

This is a **permanent, reusable tracking issue**. It is **never closed**. Each `/t3:dreaming` (`t3 dream run --full`) pass appends a dated entry here recording what it consolidated and the core-gap tickets it spawned. Treat it as the long-lived index of dream-pass outcomes, the same way `MEMORY.md` indexes the memory corpus.

### Why this exists

A dream pass distils session drift into the memory ledger and, when it records clusters, files `needs-triage` core-gap tickets. But the per-pass narrative — what drift was seen, why it mattered, which gaps it produced, and whether the engine itself fell short — has no durable home. This ledger is that home. It also lets a human skim the trend across passes (which drift keeps recurring despite memories, which gaps got fixed).

### How to use it (each pass)

1. Run the pass.
2. Append a new `## YYYY-MM-DD` section below with: a one-paragraph summary, the memories written/updated, the core-gap tickets spawned (checkbox list, linked), and any limitation of the pass itself.
3. Check the box on a gap when its fix merges.

---

## 2026-06-23

**Pass:** `t3 dream run --full` over the day's main session (a ~31MB, 3-day transcript). All six §4 acceptance gates passed (lossless); 105 cross-link edges; `MEMORY.md` re-indexed.

**Key limitation found this pass:** the engine recorded **0 clusters from 1919 members** — it was near-blind to the day's actual drift. Root cause (two compounding limits in `loops/dream/engine.py`): (1) transcript extraction (`_high_signal_lines`) keeps a line only if it contains a literal signal keyword (`TEATREE GATE` / `BLOCK` / `DENIED` / `feedback_` / `BINDING` / `retro` / `cold review`), so raw user-correction *prose* is filtered out before the LLM sees it; (2) memory files (weight 90–100) fill the 60K extract ceiling before recency-gated transcripts (weight 10) are ever read. Net: a plain pass consolidates the already-distilled memory corpus but cannot mine *fresh* transcript drift. The day's drift was therefore captured by hand and is listed below.

[...]

Issue #2597, opened 2026-06-20
banned-terms publish gate false-positives on the overlay name when commenting on its own private tracker


souliane:
## Problem

The #1415 banned-terms publish gate is documented to protect **upstream public artifacts** (public teatree issues/PRs/comments, blog posts, public docs). In practice it currently also fires when posting comments to the **overlay's own private issue tracker** — the private internal repo whose own name, URLs, and issue bodies intrinsically contain the customer/overlay name.

On a private tracker, the overlay name is part of the repo's own identity, so a comment that mentions it is **not a leak — it's a false positive**. Today this forces operators to use the `ALLOW_BANNED_TERM=1` escape on *every* internal status post.

Observed this session: 3 consecutive status comments to the private tracker were all blocked on the overlay name, each requiring the manual escape.

## Proposed fix

Scope the banned-terms post-gate to **public destinations only**. Equivalently: allowlist the banned term when the target repo is the private repo whose name *is* that term (or, more generally, when the destination is private).

Internal posts to a private tracker should not be gated on the customer/overlay name. The public-destination gate (public teatree issue/PR, blog, public docs) stays exactly as-is.

## Acceptance criteria

- Posting a comment containing the overlay name to the overlay's **private** tracker is **not** blocked (no `ALLOW_BANNED_TERM=1` needed).
- Posting the same comment to a **public** destination (a public teatree issue/PR, a blog post, public docs) **is still blocked**.

souliane:
Consolidated into epic #2563 (2026-07-01 backlog consolidation → ≤10 tracking epics). This ticket is folded into #2563's checklist and tracked there; it is closed only to keep the tracker at the epic level, **not** because the work is done. The issue stays readable and can be reopened if it needs standalone attention.

Issue #2034, opened 2026-06-06
banned-terms override must be scoped per-segment: a publish-shaped decoy still vouches for a chained publish


souliane:
Follow-up from PR #2031's cold review (verdict 166, non-blocking finding). The override guard now rejects NON-publish decoys, but a publish-SHAPED decoy passes `_is_publish_command` on its own segment while `has_override` remains a whole-command boolean that skips scanning the entire flattened payload (`_run_banned_terms_pretool`): `ALLOW_BANNED_TERM=1 git commit -m harmless && gh issue create --body <term>` → override accepted, term in the OTHER segment leaks. #2031 also widened this one notch (the cd-prefixed variant flipped from blocked to allowed).

The docstring (`banned_terms_scanner.py:~162`) and `docs/blueprint/factory-architecture.md:206` over-claim: 'a decoy override on an unrelated chained segment cannot vouch for a publish elsewhere' — true only for non-publish decoys today.

Fix: scope the override per-segment — it suppresses only its OWN segment's payload; every publish segment needs its own leading override (or the conservative variant: any unoverridden publish segment keeps the full scan). Add must-BLOCK corpus entries for the publish-shaped decoy (plain and cd-prefixed) and correct the docstring/doc claims. Threat model note: ALLOW_BANNED_TERM is an operator-controlled escape, so this is accidental-leak hardening, not adversarial defense — but the doc must not promise more than the code does.

souliane:
Consolidated into epic #2563 (2026-07-01 backlog consolidation → ≤10 tracking epics). This ticket is folded into #2563's checklist and tracked there; it is closed only to keep the tracker at the epic level, **not** because the work is done. The issue stays readable and can be reopened if it needs standalone attention.

PR #2673, opened 2026-06-23
fix(banned-terms): don't flag the org term inside an overlay's own repo URLs

### LINKED ISSUE #2672: banned-terms gate URL false positive: flags the org term inside an overlay's own repo URL
## Problem

The #1415 banned-terms posting gate scans the body of a `gh`/`glab` create/comment before it reaches a public surface and blocks on any configured org/namespace term. It over-blocks a **legitimate** post when the only occurrence of the term is **inside a forge URL** whose host+namespace is the address of one of the active overlay's own configured private repos — e.g. a body that cites a customer work item at `https://<host>/<org>-engineering/<repo>/-/issues/N`.

That URL is the structurally-required address of the work item, not a leak. The org token is part of the repo's own identity (it appears in the URL because the URL points at that repo), so a public PR/issue/comment that links the work item gets hard-blocked and forces an `ALLOW_BANNED_TERM=1` escape on every such post.

This is a **sibling** of:

- #2597 — the gate false-positives on the overlay name when commenting on its own private tracker (the destination-skip dimension).
- #2034 — the override must be scoped per-segment (the override dimension).

This issue is the **URL-allowlist dimension**: the destination is genuinely public (so the body is scanned), but the term sits only inside an own-repo URL.

## Proposed fix

A DERIVED carve-out — never hardcoded. A forge URL is "own-repo" only when its host+namespace slug matches one of the `[teatree] private_repos` entries (the same allowlist the existing carve-outs consult, host-qualification-symmetric). Blank every own-repo URL before matching (mirroring the existing email carve-out) and downgrade to a warning **only** when the term then has no remaining occurrence.

A bare term outside such a URL, a term inside a foreign URL, an empty allowlist, or a config read error all keep the hard block — a detection failure never weakens the public-surface gate. Scoped to the structured `gh`/`glab` posting verbs, never to `git commit` (the commit path keeps its own landing-repo carve-out).

## Acceptance criteria

- A public-surface post whose body's only occurrence of the term is inside an own-repo forge URL is **not** blocked (downgraded to a stderr warning).
- The same term appearing as a bare word, or inside a URL under no configured namespace, **still** hard-blocks.
- No configured private-repo namespace means no downgrade (conservative default).

Refs #2663
Siblings: #2597, #2034

souliane:
## What

The #1415 banned-terms posting gate over-blocked a legitimate `gh`/`glab` create/comment to a PUBLIC surface when the org/namespace term appeared **only** inside a forge URL whose host+namespace is the address of one of the active overlay's own configured private repos (e.g. a body that cites a customer work item at `https://<host>/<org>-engineering/<repo>/-/issues/N`). That URL is the structurally-required address of the work item, not a leak.

## Why

The term is part of the repo's own identity — it appears in the URL because the URL points at that repo. Forcing an `ALLOW_BANNED_TERM=1` escape on every post that links an own work item is the false positive this fixes.

Sibling of #2597 (destination-skip on the private tracker) and #2034 (per-segment override scoping). This PR is the **URL-allowlist dimension**: the destination is genuinely public (body is scanned), but the term sits only inside an own-repo URL.

## How

- New pure-detection module `teatree.hooks.own_repo_url_carve_out`. `term_only_inside_own_repo_urls` blanks every forge URL whose host+namespace slug matches a `[teatree] private_repos` entry (host-qualification-symmetric, reusing `slug_namespace_matches`) and returns True **only** when the term then has no remaining occurrence.
- Forge-URL slug parsing routes through the canonical `teatree.utils.url_slug.slug_from_issue_or_pr_url` — no re-derived path grammar (the AST `check-url-classify` gate enforces this).
- Wired into the banned-terms deny path as a downgrade-to-warn, scoped to the structured `gh`/`glab` posting verbs (`is_gh_glab_posting_command`), never `git commit` (the commit path keeps its own landing-repo carve-out, so a public-landing commit citing the same URL stays blocked).
- The deny chain is relocated to a bare sibling hooks module (`banned_terms_deny`) so the shrink-only router nets smaller.
- A bare term outside such a URL, a term inside a foreign URL, an empty allowlist, or a mixed own+bare payload all keep the hard block.

## Tests

- `tests/teatree_hooks/test_own_repo_url_carve_out.py` — pure-logic: own URL downgrades (GitLab + GitHub), bare term blocks, foreign URL blocks, host-qualified entry, no-allowlist, mixed own+foreign, empty guards. 100% module coverage.
- `tests/teatree_hooks/test_banned_terms_hook.py` — live-hook: a public post whose body's only occurrence sits inside an own-repo URL is allowed (warn); the same term as a bare word alongside the URL still hard-blocks. The new downgrade was observed RED before the fix; the protective bare-term case stayed GREEN.

## Architecture pre-check

1. **BLUEPRINT §** — #1415 posting gate; additive sibling of the email carve-out and the own-slug commit downgrade. No § rename.
2. **FSM** — n/a.
3. **Extension points** — none changed; consumed only by the deny path.
4. **Component boundaries** — pure detection in `teatree.hooks`; deny/warn in the bare sibling hooks module.
5. **Dependency direction** — `teatree.hooks` → `teatree.utils` (existing). `tach check` green.
6. **Test surface** — unit + live-hook above; RED-before-GREEN confirmed.
7. **Resilience** — fail-safe-to-block; no external write; idempotent pure function.
8. **Identity** — URL host+namespace normalized via `slug_namespace_matches`; no new strip/split seam.
9. **Behavior preservation** — purely additive; no matcher narrowed, no must-block test inverted.

Closes #2672
Refs #2663


souliane:
Folded into #2688, which carries both this own-repo-URL carve-out (deny-hook posting path) and the company-identifier matcher carve-out, with anti-vacuous tests proving each surface.

PR #2675, opened 2026-06-23
fix(hooks): company-identifier allowlist carve-out for the banned-terms gate (#1415)

### LINKED ISSUE #1415: feat(privacy): gate gh/glab issue+PR+comment posting with the banned-terms scan (commit hook misses non-commit writes)

souliane:
## What

Adds an optional `[teatree].banned_terms_allowlist` carve-out to the banned-terms gate (#1415) so a short single-token org slug in `banned_terms` can no longer over-fire inside the org's OWN compound identifiers (`<org>-engineering`, `<org>-product`, internal-URL namespaces).

The carve-out lives in the SINGLE shared matcher (`term_match._strip_allowlisted_tokens`), mirroring the existing email carve-out: each allow-listed identifier's whole-token run is blanked BEFORE matching, so a shorter banned term can no longer surface inside a longer company identifier. It flows through every entry point that reads the flat `banned_terms` list — the shell hook (`banned_terms_cli`), the posting/DM gate (`banned_terms_scanner`, including report attribution), and the diff-only commit scan — so they cannot drift. Empty (the default) is a no-op preserving prior behaviour.

## Why

The banned-terms gate tokenizes a single-token org slug out of every company-owned compound identifier. Because the slug is a sub-token of the org's own identifiers and internal-URL paths, the gate flagged the org's OWN identifiers on every one of its MR/post bodies and DM/notify sends — the recurring #1415 over-block that prior destination-aware patches never closed at the matcher level. This is the live over-block the operator hit: a benign DM/notify body mentioning the org's own repo/namespace was blocked.

## Confirmation that the fix stops the DM/notify over-block

The new scanner-level tests (`TestCompanyIdentifierAllowlistGate`) run the FULL end-to-end gate — `banned_terms_scanner.scan_text` -> `check-banned-terms.sh` -> `banned_terms_cli` -> `term_match` with the TOML allow-list — which is exactly the path the PreToolUse hook (`_run_banned_terms_pretool`) drives for a `t3 teatree notify send` DM. A company-identifier / internal-URL body now PASSES the gate, while a genuine customer codename NOT on the allow-list STILL blocks (control), and a bare standalone org-slug token STILL fires.

## Anti-vacuity evidence

The regression tests were observed RED before the fix and GREEN after, both at the matcher level and at the end-to-end gate level:

- With `term_match._strip_allowlisted_tokens` temporarily neutered (returning its input unchanged, i.e. the pre-fix behaviour), **16 of the new tests fail** — including the scanner-level end-to-end cases — while the 7 control tests (customer codename still blocks, no-allowlist preserves over-block, bare slug still fires) stay green, proving the carve-out does not gut the gate.
- With the fix restored, all banned-terms / term-match tests pass.

## Test plan

- `uv run pytest tests/teatree_hooks/test_term_match.py tests/teatree_hooks/test_banned_terms_scanner.py` — 352 passed
- `uv run pytest tests/teatree_hooks/test_banned_terms_parity.py tests/teatree_hooks/test_banned_terms_cli_diff_only.py tests/teatree_hooks/test_banned_terms_hook.py tests/test_banned_terms_tree_scan.py tests/test_banned_terms_shell_fallback.py tests/teatree_hooks/test_term_match.py tests/teatree_hooks/test_banned_terms_scanner.py` — 485 passed
- `uv run pytest tests/teatree_hooks/` — 1534 passed
- `uv run ruff check` + `uv run ruff format --check` on all changed files — clean
- Anti-vacuity probe (neuter the carve-out) — 16 RED, 7 control green; restored — all green

## Post-mortem

- **Root cause:** the gate's whole-token matcher treats a single-token org slug as a match anywhere that token appears, so it surfaced inside the org's OWN compound identifiers. Earlier #1415 patches were destination-aware (skip the gate for known-private targets) but never closed the over-block at the MATCHER level, so the org's own identifiers still tripped on surfaces the destination skip didn't cover (DM/notify bodies among them).
- **Fix shape:** an opt-in allow-list applied in the one shared matcher, mirroring the email carve-out that already exists for the same class of false positive. Putting it in the single matcher means every entry point inherits it and cannot drift.
- **Why it stayed open:** the over-block recurred across multiple destination-aware patches because none of them removed the matcher-level cause. Carving the company's own COMPOUND identifiers out at the matcher (while still firing on a bare standalone slug and on genuine customer codenames) is the narrowest change that removes the cause without weakening the gate.
- **Operator action required to benefit:** declare the org's own compound identifiers under `[teatree].banned_terms_allowlist` in `~/.teatree.toml`. Default-empty preserves today's behaviour, so this is a non-breaking, opt-in change.

## Architecture pre-check — #1415

1. **BLUEPRINT alignment:** Section 9 Code Host Sync — Banned-terms posting gate. The BLUEPRINT entry is updated in this PR to document the company-identifier carve-out.
2. **FSM phase boundaries:** n/a — no `Ticket.State` / `Worktree.State` transition.
3. **Extension-point contracts:** n/a — no `OverlayBase` / scanner-registration / Protocol change. The change is internal to the banned-terms matcher and its three flat-list readers.
4. **Component boundaries:** `src/teatree/hooks/` (matcher + the two flat-list readers). Correct home — pure detection logic, no business/FSM logic.
5. **Dependency direction:** no new cross-module imports; the carve-out is internal to `teatree.hooks.term_match` consumed by its existing siblings.
6. **Test surface:** `tests/teatree_hooks/test_term_match.py::TestCompanyIdentifierAllowlistCarveOut` + `TestStripAllowlistedTokens` (matcher), `tests/teatree_hooks/test_banned_terms_scanner.py::TestCompanyIdentifierAllowlistGate` (end-to-end gate). Each pins the carve-out passing AND the control still blocking.
7. **Resilience invariants:** the gate is detection-only; the carve-out keeps the existing fail-closed contract (a real customer codename still blocks; empty allow-list is a no-op).
8. **Identity and key normalization:** the allow-list is normalized through the SAME `tokens()` tokenizer the term list uses; longest-run-first consumption avoids a shorter overlapping entry partly eating a longer identifier.
9. **Behavior preservation / capability deletion:** purely additive and opt-in — default-empty allow-list preserves the prior (over-blocking) behaviour. No must-block test was inverted; the controls prove the gate is not narrowed for genuine customer codenames or bare org slugs.

Closes #1415.


souliane:
Duplicate of #2680 (identical diff — same patch-id 2804cff24f88a10c). Closing in favor of #2680.

PR #2680, opened 2026-06-23
fix(hooks): company-identifier allowlist carve-out for the banned-terms gate (#1415)

### LINKED ISSUE #1415: feat(privacy): gate gh/glab issue+PR+comment posting with the banned-terms scan (commit hook misses non-commit writes)

souliane:
## What & why

Closes #1415.

A single-token org slug listed in `[teatree].banned_terms` also tokenizes out of every company-owned **compound** identifier (`<org>-engineering`, `<org>-product`, internal-URL namespaces), so the banned-terms posting gate flagged the org's OWN identifiers on every one of its MR/post/DM bodies — the recurring #1415 over-block. The user hit this live: a `t3 ... notify send` self-DM was blocked because the body mentioned the org's own compound identifier, whose leading token is also a bare banned term.

Prior destination-aware patches never closed this at the matcher level — they reasoned about *where* the body was going, not about the fact that a shorter term was surfacing inside a longer company-owned identifier.

## The fix

Add an optional `[teatree].banned_terms_allowlist` carve-out in the **single shared matcher** (`term_match._strip_allowlisted_tokens`), mirroring the existing email carve-out:

- Each allow-listed identifier's whole-token **contiguous run** is blanked from the text BEFORE banned-term matching, so a shorter banned term (a bare org slug) can no longer surface inside a longer company-owned identifier or an internal-URL path.
- A genuine customer codename that is NOT on the allow-list is still blocked (the gate is not gutted).
- A bare standalone org-slug token that is NOT part of an allow-listed run still fires (the carve-out exempts the company's own COMPOUND identifiers, not every appearance of a sub-token).
- Longest-first matching so a multi-token identifier (`<org>-client-workspace`) is consumed as one run rather than partly eaten by a shorter overlapping entry.

The carve-out flows through every entry point that reads the flat term list — the shell hook (`scripts/hooks/check-banned-terms.sh` → `banned_terms_cli`), the PreToolUse posting gate (`banned_terms_scanner`, including report attribution), and the diff-only commit scan — so they cannot drift. Empty (the default) is a no-op preserving prior behaviour.

## Anti-vacuity proof

Reverted the single matcher line (`_strip_allowlisted_tokens(tokens(text), allowlist)` → `tokens(text)`) and re-ran the new test classes: **16 tests went RED** — every "company-identifier body PASSES" case failed (the bare org slug was blocked inside the compound identifier), AND the attribution test failed (a real codename beside a company identifier was mis-attributed to the carved-out slug). The 11 control tests stayed GREEN (genuine codename still blocked, bare slug still fires, no-allowlist no-op) — confirming they correctly do NOT depend on the carve-out. Restored the line; all green again.

## Verification (all run locally, this branch, rebased on current `origin/main`)

- `uv run ruff check` on all 5 touched files — clean.
- `tests/teatree_hooks/test_term_match.py` + `tests/teatree_hooks/test_banned_terms_scanner.py` — 352 passed.
- Broader banned-terms surface (parity, diff-only, shell hook, env-override, tree-scan, no-overlay-leak, privacy gate) — 208 passed.
- Full `tests/teatree_hooks/` directory — 1534 passed.

### 1. BLUEPRINT § alignment
§9 "Code Host Sync" — Banned-terms posting gate. BLUEPRINT updated in the same commit with a "Company-identifier carve-out" sentence describing the new `banned_terms_allowlist` array and that it flows through the single shared matcher.

### 2. FSM phase boundaries
n/a — no `Ticket.State` / `Worktree.State` transition.

No `OverlayBase` / scanner-registration / Protocol surface changed. `matched_term` / `line_matches` / `file_matches` gained an optional `allowlist` keyword with an empty default, so every existing caller is source-compatible and behaviour-preserving.

### 4. Component boundaries
All changes stay in `src/teatree/hooks/` (matcher + the two readers) and `scripts/hooks/` (shell hook doc) — the privacy/leak gate component. No straddle.

### 5. Dependency direction
No new cross-module imports; the readers import `term_match` (a leaf hooks module) as before. No backwards edge.

### 6. Test surface
- `term_match._strip_allowlisted_tokens` — direct unit tests (multi-token run, single-token entry, partial run kept, empty-allowlist identity).
- `matched_term` / `line_matches` — carve-out applied, control codename still blocks, bare slug still fires, no-allowlist preserves prior behaviour, longest-first.
- `banned_terms_scanner.scan_text` (full gate: TOML → shell → matcher) — company-identifier body passes, real codename blocks, codename-beside-identifier attributes the codename, bare slug blocks, no-allowlist preserves over-block.

### 7. Resilience invariants
The carve-out only *narrows what is flagged*; it adds no external write. The fail-closed shape of the scanner (crash/timeout/exit-2 → `SCANNER_UNAVAILABLE_MARKER` and BLOCK) is unchanged. An unreadable/missing config yields an empty allowlist (no-op), never an exception.

Allow-list entries tokenize through the same `tokens()` function the terms and text use, so matching is on the canonical token form on every side — no strip/split-to-make-it-match.

Purely additive. No existing matching behaviour is removed: with an empty (default) allowlist the matcher is byte-for-byte the prior behaviour. No must-block regression test was inverted. The carve-out does NOT narrow the gate's coverage for any term that is not an explicit operator opt-in — a genuine customer codename is still blocked, and the control tests pin that.

- **Assumption:** the operator populates `banned_terms_allowlist` only with their org's OWN compound identifiers (org/repo names, internal-URL namespaces) — never with customer PII. This is the same trust model as `banned_terms` itself (operator-curated). The carve-out is opt-in and inert by default, so no existing deployment changes behaviour until the key is added.
- **Assumption:** a bare single-token allow-list entry drops that whole token wherever it appears (documented in `_strip_allowlisted_tokens`). Compound identifiers are the intended use; single-token entries are honoured for completeness.

## Post-mortem note

The prior automation attempt at this ticket completed the git/code work (commit `f6bc93b26`) but failed on output-formatting only (a StructuredOutput retry cap), so no PR was opened. This PR picks up that committed work: rebased the stale branch onto current `origin/main` (clean), re-verified anti-vacuity and the full hooks suite, and opened the PR.


souliane:
Folded into #2688, which carries both this company-identifier matcher carve-out (shell hook, scanner, CLI, and deny hook) and the own-repo-URL carve-out, with anti-vacuous tests proving each surface.

Issue #2359, opened 2026-06-13
Full-suite test reliability: env-dependent git-fixture default-branch + order-dependent shared-state flakes block PR CI


souliane:
## Problem
Multiple open PRs red-flag the 3.13 full-suite CI job on failures unrelated to the PR diff, and several coders independently reported the same flakes. Two distinct reliability gaps surface only under certain collection orders / CI git config.

### Class A — env-dependent git-fixture default branch
Real-git fixtures run `git worktree add <path> main` (or assume `main` is the default). They pass locally (dev box has init.defaultBranch=main) but exit 128 in the CI image where the default differs. Confirmed:
- test_workspace.py TestCleanAllReapsAndSurvivesForeignOverlay::test_unclassifiable_sibling_repo_is_skipped_not_crashed (fixed in #2354 by making the fixture default-branch-independent)
- test_pull_main_clone_scanner.py and test_self_update_scanner.py default-branch fast-forward tests (still failing on other PRs)

Fix direction: make every real-git fixture default-branch-independent (git init -b main + initial commit, or branch -M main, or reference HEAD); verify under GIT_CONFIG_GLOBAL=/dev/null. A repo-wide sweep plus a shared make_git_repo(default_branch=...) helper prevents recurrence.

### Class B — order-dependent shared-state pollution
Tests that pass in isolation but fail under a full-suite order due to leaked global state. Confirmed (each passes in isolation / its own file):
- test_review_claim_discipline.py test_open_colleague_mr_dispatches_without_eyes_reaction (review-intent shared state)
- test_per_item_fault_isolation_1597.py (loop broadcasts)

Fix direction: find the leaking setUp/teardown (global singletons, module-level caches, unreset patches) and isolate; an autouse reset for the shared singletons; audit with a shuffled run locally.

## Why it matters
These flakes block the keystone merge gate (CI-green precondition) on PRs whose own changes are correct, forcing per-PR re-runs. A determinism/reliability gap (architectural-fitness concern), not a per-PR defect.

## Acceptance
- [ ] Class A: all real-git fixtures pass under GIT_CONFIG_GLOBAL=/dev/null; shared make_git_repo helper
- [ ] Class B: the two named order-dependent tests pass under a shuffled full-suite run; leaking state isolated
- [ ] CI audits the suite with a shuffle so these cannot silently return

souliane:
[...]

PR #2681, opened 2026-06-24
test(reliability): isolate T3_LOOPS_DISABLED env leak + CI shuffle audit (#2359 Class B)

souliane:
Scope: #2359 Class B (order-dependent shared-state pollution) plus the CI shuffle audit (acceptance 3). Class A is already done and merged (#2362, #2522) and is not redone here. Closes the Class B and shuffle-audit acceptance of #2359.

Root cause: two discovery-time review-intent scanner tests (test_review_claim_discipline.py::test_open_colleague_mr_dispatches_without_eyes_reaction and test_per_item_fault_isolation_1597.py) pass in isolation but fail under a shuffled full-suite order. review_loop_enabled() reads the T3_LOOPS_DISABLED env kill-switch at call time via loop_enabled_by_name; a value containing review or all left in the process env makes filter_review_intent_signals drop every broadcast review-intent signal, so the victims see an empty signal list. The _isolate_env autouse conftest fixture already redirected CONFIG_PATH to close the host [loops.review] config-leak path but never cleared the env var, so under pytest-randomly reordering a leaked value reached the victims.

Fix: clear T3_LOOPS_DISABLED in the _isolate_env autouse fixture so the review-loop env state is hermetic per test, matching the isolation already applied to the host config and the other behavioural env vars. This isolates the whole class of env-leak pollution rather than serialising one polluter-victim pair.

Orphan branch 2700-fix-scanner-ordering-flake: evaluated and rejected. Its serialize-scanners commit fixes a runtime scanner-pool ordering invariant between live scanners, not the test-level env leak; it would not fix the named tests and bundles unrelated provision-timeout and eval-watchdog changes.

CI shuffle audit: a dedicated test-shuffle lane runs the order-fragile tests/teatree_loop/ subset under pytest-randomly across the recorded reproducer seed 7 plus others. pytest-randomly is in a new opt-in shuffle dependency group, keeping the default pytest run and the heavy test (3.13) coverage lane deterministic. Scoped to tests/teatree_loop/ because a few concurrency/timeout tests seed their isolated DB at import time and only collect in natural order (a full-suite shuffle reds on an unrelated collection error); the Class B flakes live in that dir and it collects cleanly under shuffle.

Regression proof: pre-fix, seed 7 over tests/teatree_loop/ red 27 tests; post-fix the same command passes (2056 passed) and seeds 1, 7, 13, 42, 100, 555, 999, 31 are all green. The new test_review_loop_env_isolation_2359.py poisons the env var at import and asserts the fixture clears it (RED before, GREEN after).

Not merged: pushed for independent cold-review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -473,6 +473,62 @@ jobs:
           (run_heavy_python=false). Emitting a passing 'test (${{ matrix.python-version }})'
           so the required status context is satisfied without re-running the suite."
 
+  # GATE (heavy lane): order-dependence audit. Runs the order-fragile loop test
+  # subset under a SHUFFLED collection order (pytest-randomly) so a test that
+  # leaks shared global state — a stray env var, an unreset singleton, a
+  # module-level cache — and the victim it breaks cannot silently return
+  # (souliane/teatree#2359 Class B). The heavy `test (3.13)` lane runs the suite
+  # in its committed file order and so never exercises the polluter→victim
+  # adjacency; this lane does.
+  test-shuffle:
+    # WHY a SUBSET, not the whole suite: a small set of concurrency/timeout tests
+    # seed their isolated DB at import time and are only collectable in the
+    # suite's natural order, so a full-suite shuffle reds on an UNRELATED
+    # collection error. The Class B flakes live in ``tests/teatree_loop/``, which
+    # collects cleanly under shuffle — scoping here keeps the gate signal about
+    # order-dependence, not about that separate import-order fragility.
+    #
+    # WHY ``--group shuffle`` (not the default ``dev``): adding pytest-randomly to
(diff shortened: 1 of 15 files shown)


souliane:
Heads-up for cold-review: this branch was created off a local main HEAD that carried one not-yet-pushed maintainer commit, ed42e91d5 (test: close git-fixture default-branch scanner blind spot, #2359 Class A). It is a test-only Class A hardening (extends the existing fitness function to recognise binary-variable git calls) authored by the maintainer on Jun 17, and exists only on this branch — it was not authored by me and dropping it via rebase would orphan it. I left it in rather than destroy it. My Class B change is the single commit ad237e71e (conftest env isolation + CI shuffle audit + regression test). If you prefer Class A split out, it can be cherry-picked to its own PR before merge.

PR #2806, opened 2026-06-29
feat(config): Django-free stdlib sqlite cold reader for DB-home settings (config-unify PR1)

souliane:
PR1 of the config-unification migration (DB-only runtime). Lands a new
stdlib-only reader for the `teatree_config_setting` override store that does
not boot Django and does not import `teatree.paths` (whose module-level
`resolve_data_dir` would auto-isolate a worktree onto a sibling DB). The
deliberate inverse: a worktree checkout resolves to the PRIMARY
`~/.local/share/teatree/db.sqlite3`, the same DB the installed `t3` uses, so a
future bash/statusline path reads the same config a Django session would.

The ~5-line path computation is duplicated rather than imported; a test pins it
equal to `teatree.paths.TRUE_CANONICAL_DB` so the two cannot drift. Typed
wrappers mirror the strictness of `hooks/scripts/teatree_settings.py` and
`config/settings.py` exactly. Every read fails open to None / the caller's
default — missing file, absent table, locked DB, corrupt JSON, or missing row
never raises.

No callers wired in this PR — independently mergeable and revertable.


souliane:
Superseded — pre-squash duplicate branch; this content is already on main via #2793 (config-unify PR1, with later review fixes). Closing as redundant.

PR #2818, opened 2026-06-30
salvage/fix/cwd remote offline resolution

souliane:
- **fix(hooks): resolve cwd git remote offline so a flagless glab mr create to a private repo is not over-blocked**
- **test: use neutral placeholder org/repo names in offline-remote fixtures**


souliane:
Redundant: the cwd-remote-offline fix shipped via #2792 (merged). #2818's production files are byte-identical to main; its only delta is a superseded test set, and it's conflicting. Closing.

PR #2819, opened 2026-06-30
salvage/feat/config flip cold readers

souliane:
- **feat(config): seed-on-setup and flip cold-hook gate readers to DB (config-unify PR3)**
- **fix(config): route cold-hook gate toggles through the DB so t3 gate stays authoritative (config-unify PR3)**
- **test(config): relocate gate-toggle regression to mirror its teatree_gate module**
- **fix(config): verify a cold-hook gate toggle landed and report its true destination**
- **fix(config): overlay-namespace gate doc literals and add sqlite3 to the docker test image**


souliane:
Redundant: config-flip-cold-readers shipped via #2795 (merged). This salvage branch is a stale pre-#2795 snapshot with zero unique work, is conflicting, and would revert config-unify PR4. Closing.

Issue #2664, opened 2026-06-23
bug(worktree): teardown aborts on ModuleNotFoundError teatree.core.provision_timebox (runs worktree's stale code; risks incomplete cleanup)


souliane:
## What

`t3 teatree worktree teardown [--force]` emits a Python traceback ending in `ModuleNotFoundError: No module named 'teatree.core.provision_timebox'` during the teardown step run. The git worktree **is** still removed (the command reports success / the dir is gone), so it is non-fatal — but the traceback is alarming in autonomous runs, and any teardown step ordered **after** the failing import risks being skipped (incomplete cleanup of docker/DB/row).

## Reproduce

Observed twice in an autonomous burn-down while tearing down two merged ticket worktrees (`2640-…/teatree` and `2614-…/teatree`):

```
File ".../<worktree>/src/teatree/core/step_runner.py", line 94, in run_step
ModuleNotFoundError: No module named 'teatree.core.provision_timebox'
  Teardown failed for souliane/teatree
```

…after which the worktree directory was nonetheless removed.

## Likely cause

Version skew. Teardown runs `uv --directory <worktree> run python manage.py worktree teardown`, i.e. it executes the **worktree's own** (older-base) code. A teardown/provision step imports `teatree.core.provision_timebox`, which is absent on that older base (added/renamed after the worktree was created). So a teardown that should be driven by the current installed `t3` ends up importing a module the worktree's checkout doesn't have.

## Impact

- Non-fatal for the dir removal observed here, BUT the error aborts the step runner mid-teardown — later steps (drop DB, delete `Worktree` row, docker down) may be skipped, leaving orphaned DBs / rows / containers. This is a cleanup-correctness risk under autonomous operation, not just cosmetic noise.

## Fix direction

- Drive `worktree teardown` from the **orchestrating (current) `t3` install**, not the worktree's checkout — teardown should not depend on the worktree's (possibly stale) code being import-compatible.
- Or guard the optional step-module import (skip a missing optional step rather than raising), so the teardown step DAG is resilient to a base that lacks `provision_timebox`.
- Regression test: tearing down a worktree whose base lacks `provision_timebox` completes ALL teardown steps (docker stop, DB drop, row delete, dir removal) with no unhandled import error.

_Found firsthand during an autonomous backlog burn-down (post-merge worktree teardown)._

Issue #2244, opened 2026-06-10
worktree provision DB-import hangs silently when no local DSLR snapshot exists


souliane:
## Problem

`t3 <overlay> worktree provision` blocks **silently** (no output, ~0 CPU, blocked on a child-process PIPE) for 10+ minutes when there is no local DSLR snapshot for the tenant to restore. With remote-dump pulling disabled (the default for agents), the DB-import step has nothing to import and appears to hang indefinitely — no error, no log line, no timeout.

Compounded when the provision output is piped through `| tail`, which buffers the stream so even partial progress is invisible until the process exits.

## Impact

Indistinguishable from a real hang. A timebox expires with no diagnosis, and the operator can't tell "grinding" from "stuck on a missing prerequisite".

## Suggested fix

- The DB-import step should fail loud and fast: `no local DSLR snapshot for tenant <X>; run db refresh or supply a dump` — instead of blocking on a buffered subprocess.
- Extend the recently-added provision-step timebox/heartbeat (which covered subprocess steps) to the overlay's *callable* provision steps too — their inner `compose run` (migrate / db-import / seed) had no timeout/heartbeat. A silent block should abort with a non-zero rc and a named step.

## Related

Companion to the provisioning-grind work (stale-stack reaping + run-tests timebox). This is the remaining "silent hang on a missing prerequisite" gap on the provision path.

souliane:
Consolidated into epic #2556 (2026-07-01 backlog consolidation → ≤10 tracking epics). This ticket is folded into #2556's checklist and tracked there; it is closed only to keep the tracker at the epic level, **not** because the work is done. The issue stays readable and can be reopened if it needs standalone attention.

Issue #2220, opened 2026-06-10
fix(workspace): time-box long-blocking worktree provisioning/migrate ops + loudly alert user on timeout (never hang silently)


souliane:
## Problem

Long-blocking worktree operations — `worktree provision` / `worktree start` (DSLR snapshot restore + `migrate`), DB import, and test-DB rebuilds — can block indefinitely with **no progress signal and no user alert**. When `migrate` hits a **conflicting/forked migration graph** (two branches each add a migration in the same app → multiple leaf nodes), the command grinds or retries and the calling agent cannot tell "working" from "hung".

A recent session spent ~1h on exactly this: the agent looked frozen, no test ran, and the user only discovered it by asking "why is this so slow?". This is a **recurrence** of the same ~1h "frozen sub-agent" symptom captured in #2207 (which lists time-boxing as one mitigation among several). Promoting it to a focused, enforceable requirement.

## Requirement

1. **Time-box** every long-blocking provisioning / DB / test subprocess with a configurable hard ceiling (sensible per-op default; overridable in config). On exceeding it the op must **fail loud** with a clear, actionable error — never hang.
2. **Alert the user out-of-band on timeout** — not just a log line. Emit a user-facing alert (the host PushNotification surface and/or the configured Slack alert channel) so an away user is told e.g. "worktree provisioning for <ticket> exceeded <N> min and was aborted". Silent grinding is the failure mode to eliminate.
3. **Detect a migration conflict specifically** — a non-linear `migrate` plan / "Conflicting migrations" / multiple leaf nodes in an app — and surface it **immediately** as an actionable error ("forked migration graph in app X — rebase/renumber needed"), rather than letting `migrate` / `--create-db` grind to the ceiling.
4. **Progress heartbeat** while a long op runs (periodic "still restoring… / still migrating… (Nm elapsed)") so the agent, statusline, and any monitor can distinguish progress from hang.

## Acceptance

- A migration-conflict provisioning aborts within the ceiling with an actionable error **and** a user alert — covered by a deterministic test (simulate a forked graph and/or a subprocess that sleeps past the ceiling; assert non-zero exit + alert emitted + no indefinite block).
- **Anti-vacuous:** the pre-fix fixture (no timebox) hangs/goes red; the post-fix path aborts+alerts. The migration-conflict detector returns true on a synthetic two-leaf graph and false on a linear one.
- Config knob for the ceiling is documented.

## Notes

[...]

PR #2826, opened 2026-06-30
fix(worktree): fail loud when DB-import has no DSLR snapshot instead of hanging (#2244)

souliane:
## Problem

`worktree provision` blocked **silently** for 10+ minutes when there was no
local DSLR snapshot to restore. With remote-dump pulling disabled (the default
for agents), the DB-import step had nothing to import: the overlay's `db_import`
— and the callable provision steps — shelled out to subprocesses (DSLR
list/restore, `migrate`, `compose run`) with no wall-clock bound, so a child
blocked on its PIPE hung the whole provision with **no error, no log line, no
timeout**. Indistinguishable from a real hang.

## Root cause

The #2220 provision-step time-box (`run_timeboxed_step`) only covered `run_step`
(subprocess steps). The **callable** path was the remaining gap:
`run_callable_step` / the direct `overlay.db_import(...)` call ran an opaque
Python callable with no ceiling and no heartbeat, so any inner subprocess that
blocked on its PIPE hung indefinitely.

## Fix — extend the time-box to the callable path

- **`run_timeboxed_callable` / `run_timeboxed_db_import`** (`provision_timebox`)
  run a callable on a daemon thread, heartbeat while it runs, and on overrunning
  the configured `provision_step_timeout_seconds` ceiling return a loud failure
  (a FAILED `StepResult` naming the step / `False`) and fire the out-of-band user
  alert — never hang. The db_import alert is actionable: `no local DSLR snapshot
  … run db refresh or supply a dump`.
- **`run_provision_steps`** routes every callable provision step through the
  time-box, keeping the #2664 stale-base degradation (a plain run when
  `provision_timebox` is absent on a stale checkout; a present-but-broken module
  still re-raises).
- **`WorktreeProvisionRunner._run_db_import`** wraps `overlay.db_import` in the
  guard so the DB-import itself aborts loud and non-zero on a block — the titled
  bug.

## Tests (RED before, GREEN after)

- `TestRunTimeboxedCallable` — overrun → FAILED `StepResult` naming the step +
  alert; clean `CompletedProcess` interpreted; heartbeat fires.
- `TestRunTimeboxedDbImport` — overrun → `False` + the actionable
  `no DSLR snapshot … db refresh` alert; True/False passthrough; exception
  re-raised on the main thread.
- `TestRunProvisionStepsTimebox` — a blocking required callable step aborts loud
  with the named step instead of hanging; plus the #2664 degradation pair.
- `TestRunnerDbImportNeverHangs` — a blocked `overlay.db_import` aborts the
  provision `ok=False` instead of hanging.

RED was observed (new functions absent at import; the runner / provision-steps
assertions fail on the pre-fix code). All affected modules green; `ruff` /
`ruff format` / `ty` / `tach` / module-health / comment-density clean.

## Architecture pre-check — #2244

**1. BLUEPRINT § alignment** — Worktree lifecycle / provisioning ("fail loud,
never silently grind"). Extends the #2220 provision-step time-box to the
callable path.

**2. FSM phase boundaries** — n/a. No `Ticket`/`Worktree` state transition added;
only how the side-effects of CREATED→PROVISIONED fail (loud vs hang).

**3. Extension-point contracts** — `OverlayBase.db_import` / `get_provision_steps`
signatures unchanged; their results are now run under a wall-clock watchdog by
the core runner. No overlay edit required.

**4. Component boundaries** — `provision_timebox` (time-box primitives, +2 fns),
`step_runner` (routes callable steps through the box, #2664 degradation),
`runners/worktree_provision` (wraps the db_import call). All `teatree.core`.

**5. Dependency direction** — All edits intra-`teatree.core` (one tach module).
The `provision_timebox → step_runner` edge already existed (`StepResult`).
`uv run tach check` clean.

**6. Test surface** — see Tests above (callable time-box, db_import time-box,
provision-steps routing, runner abort, #2664 degradation).

**7. Resilience invariants** — verify-by-re-read n/a (no external write);
fallback-transport = best-effort `notify_user` alert (never raises into caller);
idempotency unchanged (the watchdog adds no state); heartbeat default logs a line
(addresses "no log line"); sub-agent contract n/a.

**8. Identity and key normalization** — n/a.

**9. Behavior preservation / capability deletion** — purely additive.
`run_callable_step` keeps its exact contract (now reached through the time-box,
or directly on the #2664 degradation). No matcher narrowed; no must-block test
inverted.

Refs #2244.

### DIFF
--- a/src/teatree/contrib/t3_teatree/overlay.py
+++ b/src/teatree/contrib/t3_teatree/overlay.py
@@ -149,16 +149,24 @@ class TeatreeOverlay(OverlayBase):
                     continue
                 run_checked(["uv", "pip", "install", "-e", str(overlay_worktree)], cwd=repo)
 
+        # Both steps are pure subprocess shellouts (uv sync / uv pip install) that
+        # touch no ORM, so they are time-boxed on a worker thread (subprocess_only)
+        # — a network stall aborts loud instead of hanging the provision silently
+        # (souliane/teatree#2244). The teatree overlay declares NO db_import strategy,
+        # so these are the ONLY provision steps it runs; without the bound the
+        # dogfooding path had no ceiling/heartbeat/alert at all.
         return [
             ProvisionStep(
                 name="sync-dependencies",
                 callable=sync_deps,
                 description="Install Python dependencies with uv sync",
+                subprocess_only=True,
             ),
             ProvisionStep(
                 name="install-overlays-editable",
                 callable=install_overlays_editable,
                 description="Install discovered overlays editable from their ticket worktrees",
+                subprocess_only=True,
             ),
         ]
 
--- a/src/teatree/core/provision_timebox.py
+++ b/src/teatree/core/provision_timebox.py
@@ -34,11 +34,12 @@ import subprocess  # noqa: S404 — only TimeoutExpired accessed, no shelling he
 import threading
 import time
 from collections.abc import Callable, Sequence
+from dataclasses import dataclass
 from pathlib import Path
 
 from teatree.config import get_effective_settings
 from teatree.core.notify import NotifyKind, notify_user
-from teatree.core.step_runner import StepResult
+from teatree.core.step_runner import StepResult, run_callable_step
 from teatree.utils.run import run_allowed_to_fail
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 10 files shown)


souliane:
HOLD (independent re-review of the db_import-scoped time-box) — coverage regression.

Narrowing the time-box to db_import only leaves the ORM-free subprocess provision callables unprotected: `run_checked(timeout=None, capture_output=True)` for `uv sync` (src/teatree/contrib/t3_teatree/overlay.py:132) and `uv pip install -e` (:150) run through `run_callable_step` with no ceiling and can block on the child PIPE forever — the same unbounded-hang shape as #2244.

Sharpest case: the t3_teatree overlay defines no db_import strategy, so the new db_import time-box never fires for teatree’s own worktrees, yet `uv sync`/`uv pip install` are the only provision steps it runs. A network stall now hangs silently with no ceiling/heartbeat/alert on the dogfooding path.

Remedy: keep the db_import time-box, but also bound the ORM-free subprocess callables — route them through the already-time-boxed `run_step`, or pass a `timeout=` / wrap via `_join_callable_on_ceiling` (they touch no ORM, so the per-thread-connection concern in step_runner.py:250-255 does not apply). The db_import approach itself is sound; only the scope needs widening. Draft until then.

souliane:
Resolved in 87409cd789: added a typed `subprocess_only` contract on `ProvisionStep` — `True` time-boxes the step on a daemon worker thread, default `False` keeps ORM-touching callables in-process. The teatree overlay's `sync-dependencies` / `install-overlays-editable` steps are now `subprocess_only=True`, so the unbounded `uv sync` PIPE-hang on the dogfood provision path is bounded (loud abort + actionable alert). Anti-vacuity test reproduces RED on the pre-fix routing.

PR #2830, opened 2026-06-30
feat(config): export-leak guard so config-store secrets never reach a shared export

souliane:
## Why

This is **PR1 of eliminating `~/.teatree.toml` entirely** — moving *every*
setting into the DB so the TOML file can be deleted.

The blocker for finishing that migration is the secret lists (banned
terms/brands, `private_repos`, token refs). They currently stay TOML-only
because `export_db_to_toml` dumps *every* `ConfigSetting` row, and a shared
export (the future "auto-configure a fresh teatree for a new user" path) must
never leak customer/brand terms.

The DB store itself is private to the operator, so a secret *living* in the DB
is fine — the only leak surface is the **export**, not the storage. This PR
adds the guard at that surface so the secret lists can move into the DB in a
later slice.

## What

`export_db_to_toml` now applies a **secret guard** that withholds a row by
default under two complementary defenses (neither sufficient alone):

1. **`SECRET_SETTINGS`** — a per-setting `private` flag expressed as set
   membership (`teatree.config.secret_settings`): the `banned_*` lists,
   `private_repos` / `private_tests`, `internal_publish_namespaces`,
   `overlay_leak_terms`, and the token refs. A static denylist.
2. **An active banned-term content scan** over each exported `key + json(value)`
   (terms resolved from the live config via `resolve_banned_terms` +
   `load_brand_terms`, fail-safe to empty) — catches a *non-listed* key whose
   value happens to embed a customer/brand term, the case a keylist can never
   enumerate.

`export_db_to_toml` returns `ConfigExport(toml, redacted)` carrying the withheld
`RedactedRow(scope, key, reason)` set. `t3 <overlay> config_setting export`
warns each withheld row + a summary on stderr; `--include-private` exports
everything for a **personal, never-shared** backup.

No runtime behaviour change to existing flows. Adds the `--include-private`
flag to `config_setting export` and a new `secret_settings` denylist. The
content scan reuses the existing `[teatree].banned_terms` / `banned_brands`
sources.

## BLUEPRINT

§10 updated: supersedes the prior "banned terms have NO DB tier" rule —
secrets MAY live in the private DB; the export guard is what keeps a shared
dump clean.

## Tests

`tests/teatree_core/test_config_migration.py` — `TestExportSecretGuard`:
private-key withheld by default, value-carrying-a-banned-term withheld by the
content scan, `--include-private` exports everything, clean rows untouched.
30 passed.


souliane:

PR #2833, opened 2026-06-30
feat(config): move check_updates to DB-home via the cold reader

souliane:
## Why

**Stacked on #2830** (export-leak guard) — both are part of **eliminating
`~/.teatree.toml`**. Review/merge #2830 first; this branch targets it so the
diff here is just the `check_updates` move.

`check_updates` was the field most explicitly called out as "must move to the
DB". It was kept TOML-home on the rationale that its sole reader
`check_for_updates` runs on pre-Django CLI paths, where a DB read would *fail
safe to the default* and silently ignore a stored `check_updates=false`.

That rationale no longer holds: config-unify already shipped `cold_reader` — a
Django-free, stdlib-only sqlite reader of the `ConfigSetting` store, built for
exactly the pre-Django path. So `check_updates` **can** be DB-home.

## What

- `check_for_updates` reads `check_updates` from the store via
  `cold_reader.bool_setting("check_updates", default=True)` — no Django boot, so
  a stored `check_updates=false` is honoured on every pre-Django path.
- `settings.py`: `check_updates` registered in `OVERLAY_OVERRIDABLE_SETTINGS`
  (DB-home parser). `import_toml_into_db` / `t3 setup` seed it from any existing
  `[teatree] check_updates` on the next import — no value is stranded.
- `homes.py`: `check_updates` leaves the TOML carve-out → **9 fields** remain.
- `conftest`: the hermetic `check_updates=false` TOML is now inert (DB-home), so
  it's dropped from the hermetic config; the host-config leak it guarded against
  is closed by `$HOME` isolation + the `CONFIG_PATH` redirect.

`check_updates` — moves from TOML-home to DB-home. Set with
`t3 <overlay> config_setting set check_updates false`; migrated from the toml by
`t3 setup` / `config_setting import`.

## Tests

- Partition fitness test reversed: `test_check_updates_is_db_home`.
- Behavioural guard rewritten to seed a real `teatree_config_setting` DB and
  assert a `check_updates=false` row is honoured pre-Django, with no network
  (`test_disabled_check_honoured_pre_django_via_db`).
- 427 passed across `tests/config/` + `tests/teatree_core/test_config_migration.py`;
  regression sweep over the path/statusline/CLI suites green.


souliane:

PR #2834, opened 2026-06-30
feat(config): move worktrees_dir + timezone to DB-home (not bootstrap deps)

souliane:
## Why

**Stacked on #2833** (part of eliminating `~/.teatree.toml`). These two fields
were tagged "path/infra bootstrap the settings module needs to even open the
DB" — the last justification for an *irreducible* TOML core.

**That justification is false.** `src/teatree/settings.py` hardcodes
`TIME_ZONE = "UTC"` and configures `DATABASES` without reading `worktrees_dir`,
`timezone`, or `privacy`. So none of them is a DB-open bootstrap dependency, and
the toml can go to **truly zero**.

## What

- **`worktrees_dir`** → DB-home. `loader.worktrees_dir()` is Django-side (it
  reads `django.conf.settings`), so it resolves off the `ConfigSetting` store
  directly — env/Django override → overlay-scope row → global-scope row →
  `DATA_DIR/worktrees` default — the path-string accessor pattern
  `workspace_dir` / `worktree_root()` already use.
- **`timezone`** → DB-home. The `[teatree] timezone` field has **no live
  attribute reader** (the `[teatree.availability].timezone` sub-table is a
  separate field read separately). Moved for partition consistency.
- Both registered in `OVERLAY_OVERRIDABLE_SETTINGS` (seeded from any existing
  toml value by `t3 setup`); dropped from the loader base-parse and the
  `_TOML_HOME` carve-out → **7 fields** remain.

`worktrees_dir`, `timezone` — TOML-home → DB-home. Set with
`t3 <overlay> config_setting set worktrees_dir "/path"`.

## Tests

- `worktrees_dir` TOML-fallback test now asserts the toml value is ignored and
  the default stands; new `tests/config/test_worktrees_dir_db.py` covers the DB
  tier (global/overlay scope, `~` expansion, Django override) mirroring
  `test_workspace_dir_per_overlay`.
- Partition fitness test reversed (`test_falsely_bootstrap_fields_are_db_home`).
- 402 config tests + 293 worktrees_dir-consumer tests (paths, db_anchor,
  provisioning, workspace commands) green.


souliane:

Issue #2562, opened 2026-06-19
[Epic] availability-presence


souliane:
Availability & presence workstream: a durable waiting-on-you lane, TTS-audio parity on every owner-DM surface, and meeting-aware auto-mute of local TTS. This epic is the single source of truth for the scope — the folded issues below are closed and need no re-reading; everything implementable is specified here. Remaining work = **2 PRs**.

## Folded issues

| Issue | Disposition |
|---|---|
| #1941 slack_answer coalescer schedules tasks for own outbound messages | **Shipped** — on-behalf/app-posted messages are dropped by the DM-inbound scanner (`src/teatree/loop/scanners/slack_self_filter.py` re-exporting `teatree.backends.slack.self_identity`; `SlackDmInboundScanner.scan()` applies `drop_on_behalf_messages`). No work remains. |
| #2544 autonomous-away availability state | **Shipped** — `MODE_AUTONOMOUS_AWAY` in `src/teatree/core/availability.py`: questions defer to `DeferredQuestion`, `mode_pauses_self_pump` pauses only for holiday-`away`. No work remains. |
| #1947 waiting-on-you lane (durable user-blocked items) | **PR 1** |
| #2095 AskUserQuestion Slack-relay posts carry no TTS audio | **PR 2** |
| #2167 meeting-aware auto-mute of local TTS (MS Teams presence) | **PR 2** |
| #2171 Slack audio/TTS not playing (regression report) | **PR 2** (live-verification rider on the audio path) |

## Current-state anchors (verified in code — do not re-derive)

- Availability states `present` / `away` / `autonomous_away` live in `src/teatree/core/availability.py`; `AskUserQuestion` defers to the durable `DeferredQuestion` backlog under both away modes. CLI: `t3 teatree availability …` (`src/teatree/core/management/commands/availability.py`, group registered in `src/teatree/cli/django_groups.py`).
- Canonical bot→user DM egress is `teatree.core.notify.notify_user` → `_deliver_dm`: text lands via `post_message`, then `teatree.core.speak.deliver_user_dm_sidecar` runs the audio attach (`post_audio_dm`) + local playback best-effort. Every DM through `notify_user` therefore already carries audio when `speak.slack` is on.
- Un-mirrored `DeferredQuestion` rows are drained by `src/teatree/loop/scanners/deferred_question_poster.py` → `src/teatree/core/notify_question_drains.py` → `notify_user` — audio-covered.
[...]

Issue #2544, opened 2026-06-19
Add 'autonomous-away' availability state: defer questions AND keep self-pumping


souliane:
## Problem (from gate-spike RC1)
A permanent `away` override silently kills long autonomous runs. `away` does two things: (good) defers inline AskUserQuestion instead of blocking; (bad) `_pause_suppresses_self_pump()` (hook_router.py ~5881) returns True under away, suppressing `handle_loop_self_pump` (the only in-session keep-going primitive). So an unattended operator gets both-or-neither: defer-questions implies stop-self-driving.

## Ask
Add a distinct **autonomous-away** state that DEFERS questions (away behaviour) AND keeps the self-pump running (present behaviour). Keep today's **holiday-away** (defer + pause) as a separate state. Wire `_pause_suppresses_self_pump()` to return False for autonomous-away.

## Related
- gate-spike also found the launchd loop daemon pointed at a dead pyenv `t3` path (now fixed to ~/.local/bin/t3 but kept unloaded until #2513).
- ScheduleWakeup delaySeconds is one-shot — earlier narration mislabelled it as a recurring cron; no cron exists. Worth an eval asserting ScheduleWakeup is never described as a cron.

souliane:
Consolidated into epic #2562 (2026-07-01 backlog consolidation → ≤10 tracking epics). This ticket is folded into #2562's checklist and tracked there; it is closed only to keep the tracker at the epic level, **not** because the work is done. The issue stays readable and can be reopened if it needs standalone attention.

souliane:
Reopened as the next PR for #2562 (epic: availability-presence), per the 2026-07-02 holistic replan.

**Current scope:** `Resolution.defers_questions` / `Resolution.pauses_self_pump` split (already committed on the local branch), with states: present -> (False, False); plain away -> (True, True); autonomous-away -> (True, False). Plus daemon reconciliation: the loop-runner daemon (and the legacy cron/hook chokepoint, so behavior is driver-independent) consults `resolve_availability()` at each per-loop tick -- when `pauses_self_pump` is true, skip the tick silently (parked); autonomous-away keeps ticks firing. Questions raised during loop work land as `DeferredQuestion` rows. Loops carry a `colleague_facing` flag; colleague-facing loops are gated off in any away state including autonomous-away.

[...]

PR #2902, opened 2026-07-02
feat(availability): add autonomous-away state — defer questions but keep self-pumping (#2544)

adrien-oper:
feat(availability): add autonomous-away state — defer questions but keep self-pumping (#2544)

A permanent holiday-away override silently killed long unattended runs:
away conflated 'defer my questions' with 'stop self-driving'. Split the two
behaviours via Resolution.defers_questions (away + autonomous_away) and
Resolution.pauses_self_pump (away only). The stdlib hook probe gains matching
resolved_defers_questions / resolved_pauses_self_pump predicates so the
bare-python3 Stop/PreToolUse hooks read the split without django.setup().
speak local-TTS silencing and the away->present drain now key off the
defers-questions set. New CLI: t3 teatree availability autonomous-away
(fixed in a follow-up commit — the overlay CLI proxy did not expose the new
management subcommand; DJANGO_GROUPS now bridges it and a regression test
pins it).

Also reconciles availability at every per-loop tick: `loops_tick` (the single
chokepoint both the loop-runner daemon's `execute_loop` task and the legacy
native `/loop` cron converge on) consults `resolve_mode()` and parks the tick
silently when `pauses_self_pump` is true; `autonomous_away` keeps every loop
self-pumping while deferring questions.

## Scope vs #2544

- [x] `Resolution.defers_questions` / `pauses_self_pump` split (present/away/autonomous_away matrix)
- [x] Daemon reconciliation at `loops_tick` (both drivers converge on one chokepoint)
- [x] Questions raised during loop work land as `DeferredQuestion` rows (pre-existing infra, wired through the new split)
- [ ] Loops carry a `colleague_facing` flag; colleague-facing loops gated off in any away state — needs a new model field (migration) and a design decision on the gate's placement; split out to #2904 since this PR is migration-free by design

Relates to #2544. Follow-up: #2904.
Part of epic #2562.

Open questions & assumptions:
- decided-by-user: this PR stays migration-free; the `colleague_facing` loop-gating piece of #2544's original scope is tracked separately (#2904).
- assumed: `reviewer:codex-cli` is an acceptable independent-reviewer identity for the maker!=checker gate (no `t3:reviewer` sub-agent tool was available in this dispatch context; two Codex adversarial-review passes ran instead — the first found and this PR fixed the CLI-proxy gap, the second confirmed the fix).
- none further.

(cherry picked from commit fe5ae019815a330a39be786dc4370fc87f4e75c3)


souliane:
`hooks/scripts/availability_away_probe.py:72-74` — the comment claims `_PAUSING_MODES` is "Kept in parity with the core resolver by `tests/test_loop_pause_levers_stdlib.py`", but no test actually cross-checks this frozenset (or `_DEFERRING_MODES`) against `teatree.core.availability._PAUSING_MODES`/`_DEFERRING_MODES` — both suites just assert independently-hardcoded literals per mode. A future mode added to one side and not the other would go uncaught.

souliane:
`hooks/scripts/hook_router.py:6251` (`handle_route_away_mode_question`) — the deny `reason` string and the function docstring still hardcode `availability=away` even when the resolved mode is `autonomous_away`, so an unattended run narrates "away" in the deny message it hands back to the agent. Minor since the underlying deferral behavior is correct for both modes, but worth surfacing `resolution.mode` instead of the literal.

Issue #1346, opened 2026-05-26
fix(scanner): answering scanner picks up bot's own outbound DMs as questions to answer


souliane:
## Symptom

Task 108 fired with `phase=answering` against my own bot DM. The `slack_answer.question` field contained the agent's outbound stale-#1283 report (sent at slack_ts `1779823899.073799` from bot `teatree-<overlay>` to user IM `D0B35DKMKFF`). There was no actual user reply.

If the task had been left to run, the t3:answerer sub-agent would have spent compute writing an answer to the agent's own outbound message. Completed it as no-op manually.

## Root cause hypothesis

The answering scanner queries the bot's IM channel(s) for messages and creates an answering task for each message whose ts is more recent than the last-processed cursor. It doesn't filter on `message['user'] == bot_user_id` (or `bot_id == own_bot_id`) before enqueuing the task.

## Fix direction

In the answering scanner (probably `src/teatree/loop/scanners/answering*.py` — grep for the queue path that produces `slack_answer.question`), filter out messages where:

1. `user == <bot's own user_id>` (the bot's posted-as identity), or
2. `bot_id` is set AND `bot_id == <own_bot_id>` (bot-style messages don't always have `user`).

The bot's own ID is available from the `auth.test` Slack API or from cached overlay config.

## Regression test

Spec for the scanner: given a Slack history containing one message from the bot itself and one from the user, the scanner enqueues an answering task ONLY for the user message.

## Verbatim evidence

Task 108 `extra.slack_answer`:

```
channel: D0B35DKMKFF
slack_ts: 1779823899.073799
question: "teatree#1283 already fixed on main\n\nThe fix landed in <https://github.com/souliane/teatree/pull/1289|PR #1289>..."
```

The "question" text is verbatim the bot's outbound DM — confirmed by matching its prefix against the sub-agent task `adffae2fdcffd67e4`'s reported DM ts.

souliane:
Same bug also affects the **UserPromptSubmit hook** path that injects `You have 1 new Slack DM reply(ies) from the user` into the next prompt. Just witnessed:

```
UserPromptSubmit hook success: You have 1 new Slack DM reply(ies) from the user:
User replied on Slack at 1779823899.073799: teatree#1283 already fixed on main
...
```

[...]

PR #2907, opened 2026-07-02
fix(loop): skip the user's own outbound DMs in the Slack-answer coalescer

### LINKED ISSUE #1941: slack_answer coalescer schedules answering tasks for the user's OWN outbound messages
Observed 2026-06-05 (task TODO-340): the loop's DM coalescer bundled two messages the USER himself posted to his own bot DM (author U0A72P7CK0A, carrying the personal-app "Sent using <@app>" signature — they were outbound ANSWER posts) into a `slack_answer` payload and scheduled an `answering` task for them. The answerer correctly closed it without posting, but the task should never have been created.

Fix: the coalescer/classifier needs a self-authored filter — skip events whose Slack author is the configured `slack_user_id` (and/or messages signed by the user's own MCP app), since those are by definition not inbound questions.

Acceptance:
- [ ] regression test: a DM message authored by slack_user_id never yields an answering task
- [ ] existing inbound-question path unaffected (colleague-authored + bot-relayed questions still scheduled)
- [ ] anti-vacuous: test fails on pre-fix coalescer

adrien-oper:
fix(loop): skip the user's own outbound DMs in the Slack-answer coalescer

The reactive Slack-answer cycle scheduled answering tasks for messages the
user authored themselves — on-behalf outbound echoes and instructions
carrying the "Sent using <@app>" signature — because it only filtered on
the bot's own identity (#1346), never the human user's. A DM whose author
matches the configured slack_user_id is by definition not an inbound
question.

Adds AnswerKind.SELF + PendingChatInjection.mark_self_skipped() to retire
such a row from the answerer queue (loop_replied_at stamped, consumed_at
left untouched so the prompt-drain still surfaces it). The reactive cycle
drops self-authored rows via _drop_self_authored() before coalescing,
resolving each row's own overlay through a SelfUserIdResolver (default:
notify.resolve_user_id(overlay=...), now overlay-pinnable instead of only
reading the active T3_OVERLAY_NAME env). A resolver failure is fail-open
(row kept) to avoid silently dropping a real question.

Fixes #1941

Open questions & assumptions:
- decided-by-user: migration renumbered 0011 -> 0013 per the ticket's
  explicit instruction (main's queue had advanced to 0012 in the interim).
- none other.


souliane:
Two independent cold reviews (a fresh Claude session and Codex, neither of which saw the implementation conversation) both converged on the same critical finding — converting to draft, not merging as-is.

**Finding:** `_drop_self_authored()` keys on `row.user_id == configured slack_user_id`. But the only production writer of `PendingChatInjection` is `SlackDmInboundScanner` -> `SlackBotBackend.fetch_dms()` -> `open_dm(self._user_id)` -- a strict 1:1 Slack IM between the bot and that one configured human. `read_user_dms` plus the existing #1346 filter already strip the bot's own posts, so **every row that ever reaches this table today is authored by that same `slack_user_id`, by construction.** There is no code path today where a colleague's message lands in this table.

Net effect: this filter as written would retire (skip) essentially all real inbound rows for the loop's only production channel, not just the on-behalf echoes the issue targets. The "colleague-authored" regression test only proves the logic against a synthetic row shape (`user_id=UCOLLEAGUE`) that the real scanner can never produce.

**Why it slipped through:** the issue's own scoping (see the issue thread) explicitly asked for a filter keyed on "the message author id, not channel membership" -- the right instinct for a multi-user channel, but this bridge is a single-user DM, so author-id equality collapses to "always true for genuine messages too." The 2026-06-05 recurrence happened after #1346 landed (verified via git log), confirming a real remaining gap, but the gap is specifically **on-behalf posts made through the human's own token/session** vs. messages the human actually typed -- not distinguishable by `user_id` alone.

**Possible correct signal (not verified/implemented):** Slack's raw event payload may carry `api_app_id` on any message posted via the app (bot token or a user token scoped through the app), which genuine client-typed messages would lack. That needs live-Slack verification and plumbing a new field through `read_user_dms` -> `PendingChatInjection` -- a larger, unverified change than this PR, so it isn't included here.

Leaving this open in draft rather than closing it, since the underlying bug (#1941) is real and the migration/model scaffolding (`AnswerKind.SELF`, `mark_self_skipped()`) may still be useful groundwork -- only the filter predicate in `_drop_self_authored()` needs a different signal before this can ship.

PR #2908, opened 2026-07-02
fix(hooks): teach the shell lexer to consume heredoc bodies

souliane:
## Summary
- The shell lexer (`_shell_lexer.py`) had zero heredoc awareness: a `cat > file << 'EOF' ... EOF` write's body newlines were emitted as ordinary command-separator tokens, fragmenting one logical redirect into a bogus command segment per body line.
- Each bogus fragment then fed through the #1415/#1213 all-segments destination classifier as an independent shell command, over-firing SCAN on ordinary prose that merely resembled command words.
- Practical impact: a `gh pr create --repo <private-repo> --body-file <path>` false-blocked whenever the body file was written via a same-Bash-call heredoc immediately before it, even though the repo is private and the actual post is safe.

## Fix
- The lexer now recognizes every heredoc-opener shape (space-separated or glued, quoted or bare delimiter, `<<-` tab-stripping) and consumes the body verbatim from the raw command text without re-tokenizing it, so heredoc content can never be mistaken for further shell syntax or fragment the command into extra segments.
- The heredoc-carrying segment itself is unaffected by this fix and still forces a conservative SCAN (it still carries the `<<` transport-construct token) — this change only fixes the segment-count/parsing bug, it does not loosen the existing "any redirection forces scan" leak-prevention policy.

## Architecture pre-check
- Tactical/narrow parsing fix inside an existing hook module (`teatree.hooks._shell_lexer`); no FSM/Protocol/extension-point surface touched.
- Test surface: `tests/teatree_hooks/test_shell_lexer.py::TestHeredocBodyConsumption` (7 new cases: multi-line body, word-swallowing, every opener shape, `<<-` tab-stripping, unterminated heredoc, body text resembling a publish command, non-heredoc-redirect regression).
- Behavior preservation: no existing test regressed (897 passed pre-change on `main`, same suite green post-change here plus the 2 pre-existing unrelated `test_banned_terms_tree_scan.py` CLI-formatting failures, confirmed unrelated by grep — no reference to the lexer in that file or its code path).

## Test plan
- `uv run pytest tests/teatree_hooks/ tests/teatree_hooks/test_public_visibility.py tests/teatree_hooks/test_publish_destination.py -q` — all green.
- `uv run tach check` — no dependency-direction violations.
- `uv run ruff check` / `uv run ruff format --check` — clean.

### DIFF
--- a/src/teatree/hooks/_shell_lexer.py
+++ b/src/teatree/hooks/_shell_lexer.py
@@ -24,7 +24,7 @@ attached short options (``-d'{...}'``) correctly.
 
 import string
 from collections.abc import Callable
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import Final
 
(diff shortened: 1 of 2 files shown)


souliane:
LOW: verified `gate_skips_for_visibility()` before and after this diff against the PR's own motivating example (`cat > file <<EOF ...; gh pr create --repo <private> --body-file <path>`) — it returns `False` (scans, never skips) in both cases, since the write segment's `>`/`<<` token already unconditionally forces the SCAN verdict regardless of segment count. So the segment-count fix is real and worth landing (helps `is_self_rescue`'s all-segments check etc.), but it doesn't actually change whether that specific 'false-blocked private repo post' scenario scans or skips — the description's practical-impact framing overstates it.

Issue #2888, opened 2026-07-02
Gate-coverage heuristic fix (re-homed from #2565, epic #2563)


souliane:
Part of #2563. Re-homed from #2565's PR-2 sequence -- this is a gate-coverage fix, so it belongs with the gate-framework consumers, not the autonomous-lane epic.

## Scope

Fix the gate-coverage heuristic (the logic that decides whether a gate's enforcement covers a given code path) plus a regression test pinning the corrected behavior.

## Dependencies

None -- independent of every other item in this epic and the current wave.

One PR per ticket; this is the sole scope for this ticket.

souliane:
Already implemented, tested, and merged via #2912 (Protocol-class exemption for the per-diff coverage gate) — SHA f7a63e3a7aadfed16614f3bf1db636cae8cf5e04, merged 2026-07-02. Closing; PR title never carried a Closes keyword.

PR #2912, opened 2026-07-02
fix(gates): exempt Protocol classes from the per-diff coverage symbol check (#2888)

souliane:
## What

Gate 12's mutation/revert structural check (`unreferenced_changed_symbols` in `src/teatree/utils/diff_coverage.py`) required a newly-added `typing.Protocol` class to be imported by name in a changed test file, the same way it already requires any other new production symbol to be referenced.

A `Protocol` class is a structural type contract with no revertible runtime behavior of its own — its conformance is checked by the type checker (`ty`/mypy) against each concrete implementation, not by a test importing the Protocol by name. Requiring that import produced the ad-hoc `test_concrete_impls_satisfy_the_harness_protocols` binding test in `tests/teatree_agents/test_harness.py` (#2565/#2885), written purely to appease the gate. This generalizes that fix into the gate itself: `Protocol`-inheriting classes are now exempt from the symbol-reference requirement, parity with the existing private-helper and decorated-entrypoint exemptions.

## Why

Part of #2563 (review-merge-gating epic). Re-homed from #2565 to #2888.

## Investigation

I checked whether the line-coverage half of the gate (`measure_diff_coverage`) had a matching false-positive for Protocol stub method bodies (the `def foo(): ...` single-statement form). It does not: `coverage.py`'s own static analysis (`analysis2`) already excludes a bare-ellipsis-only method body from its executable-line set, confirmed empirically against both the real `Harness`/`HarnessSession` protocols and a synthetic repro. So no line-coverage code change was needed — adding one would have been untestable dead code (no RED reproduction possible), which the TDD anti-vacuity discipline forecloses. The fix is scoped to the symbol/reference check only.

## Deliverable → evidence

| Deliverable | Implementation | Evidence |
|---|---|---|
| Fix the gate-coverage heuristic to count Protocol-class references as covered | `_inherits_protocol` + exemption in `_changed_production_symbols` (`src/teatree/utils/diff_coverage.py`) | `tests/teatree_utils/test_diff_coverage.py::TestSymbolScopeRules` — 4 new tests |
| Regression test pinning the corrected behavior | `test_protocol_class_not_required_to_be_referenced`, `test_protocol_class_via_attribute_form_not_required_to_be_referenced` | Confirmed RED pre-fix (reverted the production change locally, both failed), GREEN post-fix |
| No broad weakening of the anti-vacuity check | `test_non_protocol_class_still_required_to_be_referenced`, `test_custom_protocol_subclass_not_recognized_by_name_heuristic` | Ordinary classes and transitive Protocol subclasses still require a test reference |

## Anti-vacuity proof

Reverted `src/teatree/utils/diff_coverage.py` locally (`git stash` on just that file) with the new tests in place:

```
FAILED tests/teatree_utils/test_diff_coverage.py::TestSymbolScopeRules::test_protocol_class_not_required_to_be_referenced
FAILED tests/teatree_utils/test_diff_coverage.py::TestSymbolScopeRules::test_protocol_class_via_attribute_form_not_required_to_be_referenced
FAILED tests/teatree_utils/test_diff_coverage.py::TestSymbolScopeRules::test_custom_protocol_subclass_not_recognized_by_name_heuristic
3 failed, 1 passed
```

Restored the fix: all 35 tests in the module pass. Also ran `tests/teatree_utils/`, `tests/test_block_uncovered_diff_hook.py`, `tests/test_cli_tools.py` (314 tests) green, and the full `prek run --all-files` (incl. `ty-check`, `tach`, module-health, BLUEPRINT sync) green on both commit and push.

## No migration

Pure Python logic change in `src/teatree/utils/`; no models touched.

- assumed: line-coverage of Protocol stub method bodies is not a separate bug — see Investigation above.
- assumed: the source-level Protocol-base-name heuristic (matching a literal `Protocol` name/attribute) deliberately does not resolve transitive Protocol inheritance (a Protocol subclassing a custom Protocol base) — narrower is the safe default for a gate exemption; documented and pinned by `test_custom_protocol_subclass_not_recognized_by_name_heuristic`.

docs: n/a — internal gate-heuristic fix, no user-facing CLI/setting/FSM change


souliane:
Cold review (Codex) found the initial _inherits_protocol heuristic too broad: matching bare base.attr == "Protocol" exempted any class ending in .Protocol regardless of which module it came from (e.g. class Thing(custom.Protocol)), and the equivalent gap existed for the bare-name form. Pushed a follow-up commit (560bf32e1) that resolves the file's own typing import bindings first and only exempts a base that traces back to typing.Protocol -- with two new regression tests confirmed RED against the prior heuristic and GREEN against the fix.

souliane:
Second review round (Codex) found the import-binding resolver still too permissive: it used ast.walk() over the whole tree, so a function-local "from typing import Protocol" leaked into module-level resolution, and a later re-import of the same local name from a non-typing module (or a re-aliased "import ... as t") did not shadow the earlier typing binding. Pushed commit 4c78bc4d9: the resolver now walks only tree.body (module-level statements) and applies "last import wins, in source order" semantics, matching Python's own name resolution. Three new regression tests confirmed RED against the prior resolver and GREEN against the fix (the first attempt at one of them was itself vacuous -- caught and corrected before relying on it).

PR #3049, opened 2026-07-07
fix(config): fail loud when a DB registry row silently overrides the TOML (#128)

souliane:
## Summary

Fixes #128. Editing `[overlays.<name>].path` in `~/.teatree.toml` had **no effect** and returned a **stale** overlay clone path with **zero signal**, which is the real root of the reported provisioning block (a stale worktree path pinned in the DB → `t3 <overlay>` crashes at overlay import).

### Root cause

The `overlays` / `e2e_repos` config tables are **DB-home** (#1775 eliminate-`~/.teatree.toml`). `teatree.config.loader._inject_db_registries` reads the `overlays` `ConfigSetting` row and **replaces the whole `[overlays]` TOML table** with it:

```python
for key in REGISTRY_KEYS:
    stored = cold_reader.read_setting(key)
    if isinstance(stored, dict):
        raw[key] = stored   # ← DB row wins, TOML edit silently dropped
```

So a `path` edit in the file never reaches `discovery.discover_overlays` (which reads `raw["overlays"][name]["path"]`). Unlike a DB-home `UserSettings` key — which gets `_warn_db_home_keys_in_toml` on a conflict — the **registry** override was **completely silent**: the stale DB `path` was returned with no warning.

### Fix

The DB stays authoritative (consistent with the #1775 partition), but the override is no longer silent, and there is now a **surgical, no-clobber** way to reconcile it:

- `load_config` emits **one aggregated WARN** (`_warn_registry_conflicts_in_toml`) naming each TOML registry leaf that diverges from the authoritative DB row, plus the exact reconcile command. An agreeing value, or a not-yet-migrated install with no DB row, stays silent (no noise).
- New `t3 <overlay> config_setting set-overlay <name> <field> <value>` — a read-modify-write on **one** field of **one** overlay in the `overlays` registry row, leaving every other overlay and every sibling field untouched (never hand-rewrite the whole dict).

### Tests (anti-vacuous, RED-before / GREEN-after)

- `tests/config/test_registries_db.py` — the reported bug's repro: TOML says A, DB says B → `load_config` no longer returns B **silently** (loud WARN naming the leaf + reconcile command). Observed RED without the fix (`assert 0 == 1`, no warning), GREEN with it. Plus silent-on-agreement and silent-without-a-DB-row guards.
- `tests/teatree_core/management/test_config_setting_command.py` — `set-overlay` updates one field surgically, preserves siblings/other overlays, creates the row when absent, rejects invalid JSON.

### Gates

Full config/overlay/migration suites green; ruff, ty, tach, module-health, `makemigrations --check`, test-path-mirror all clean.

### DIFF
--- a/docs/generated/management-commands.json
+++ b/docs/generated/management-commands.json
@@ -61,6 +61,10 @@
           "name": "export"
         },
         {
+          "help": "Surgically set ONE field of ONE overlay in the DB-home ``overlays`` registry",
+          "name": "set-overlay"
+        },
+        {
           "help": "List every DB config override row, naming each row's scope (read-only)",
           "name": "list"
         },
--- a/docs/generated/management-commands.md
+++ b/docs/generated/management-commands.md
@@ -32,6 +32,7 @@ Edit the source command, not this file.
 | `flags` | The read-only dead-toggle audit report over the ``FEATURE_FLAGS`` registry |
 | `get` | Print the resolved value for *key* and name its source (DB vs file/env) |
 | `export` | Dump the ``ConfigSetting`` store to TOML — the inverse of ``import`` |
+| `set-overlay` | Surgically set ONE field of ONE overlay in the DB-home ``overlays`` registry |
 | `list` | List every DB config override row, naming each row's scope (read-only) |
 | `import` | Seed the DB store from the operational toml keys (one-time migration) |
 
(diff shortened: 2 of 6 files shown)


souliane:
Superseded by #3055 — full DB-only cutover (loader mask surfaced + doctor drift check + working remediation). #3049 was WARN-only on a stale base.

Issue #2192, opened 2026-06-09
Eval suite: ground-truth corpus, failure-shape reporting, cost-regression lane, and corpus-curation tooling


souliane:
## Context

The eval suite is strong on the *cheap, deterministic, behavioral* half of evaluation — the trigger lanes, `pinned-regressions`, `negative-control`, transcript-replay, and trajectory matching all assert *did the agent do the right thing*. It is comparatively weak on the *accuracy-against-ground-truth* half that mature ML-evaluation practice does well. Four concrete improvements would close that gap.

This is an umbrella — each item can be split into its own issue.

## 1. A ground-truth corpus, not self-authored assertions

Today the behavioral assertions are written by the same author who wrote the rule under test, so the oracle is circular — the test encodes the author's belief about correct behavior rather than an independent judgment of it.

- Grow the transcript-replay lane into a real **labelled corpus**: capture real sessions, have a human (or an independent second-model judge) label the expected behavior, and grade against that label.
- The goal is an oracle sourced from *outside* the rule's author.

## 2. Report the *shape* of a failure, not just pass/fail

Behavioral evals are binary today. Where an outcome is categorical — which skill fired, which tool was chosen, which intent was routed — emit a **confusion-matrix-style** summary so a failure shows *which* category was mistaken for *which*, not just "red". The shape of the error is most of the debugging signal.

## 3. Cost/token as a tracked regression lane

Cost is currently a budget *cap*, not a regression. Add a lane that records tokens/cost per scenario and **fails on drift** — a change that quietly makes the agent more expensive is a regression in the same sense a behavioral change is. It complements the budget cap rather than replacing it (cap = ceiling; this = drift detector).

## 4. Tooling to keep the corpus fresh

A ground-truth corpus rots without curation tooling. Add a small CLI (or lightweight UI) to add, label, and review scenarios so authoring and maintaining them stays cheap. Without it the corpus stagnates and coverage silently ages out.

## Framing

These are the two halves of evaluating an AI system: *did it follow the process* (what the suite does well today) and *was the answer right* (what these items add). A complete program gates every commit on the cheap deterministic + behavioral lanes, and gates releases on the ground-truth accuracy runs.

## Suggested order

1. Ground-truth / transcript-labelling corpus (item 1) — unblocks the rest.
2. Cost-regression lane (item 3) — cheap, high value, mostly mechanical.
3. Failure-shape reporting (item 2).
4. Curation tooling (item 4).

Issue #807, opened 2026-05-16
fix(hooks): enforce user-directed questions go through the structured question tool (gate, not a remembered preference)


souliane:
## Problem
When the agent needs a decision from the user it often asks **inline in prose** inside a long status/coordinator message. In an autonomous/loop setup the user may never read that — it reads like a log line, not a prompt — so the question is effectively lost and work stalls or the agent guesses. Persisting "ask via the structured question tool" as a soft preference/memory has **not** changed behaviour (it recurs every session). This is the same vigilance-vs-enforcement failure as the rest of the durability work: a rule the agent must *remember* is not a control; only a hook is.

## Required enforcement
A teatree hook (same mechanism family as the skill-prose-ban / quality-gates gates in `hooks/scripts/hook_router.py`) that, when an assistant turn ends, detects a **user-directed question with no structured-question tool call in that turn** and blocks/flags it — forcing the question to be re-issued through the structured question tool.

Detection (heuristic, tune for precision):
- the turn's final assistant text contains an interrogative directed at the user (`?` plus a second-person/decision cue: "want me to", "should I", "which", "do you", "or …?", "shall I", "let me know if"), AND
- no structured user-question tool call occurred in the same turn (and it is not a denied/again case).

On match: fail the Stop hook with `additionalContext`/blocking message instructing the agent to re-ask via the structured question tool (do not just warn — gate it, like the other non-bypassable hooks).

## Acceptance
- Hook lives in the teatree hook router, enforced (no silent bypass / no `relax:` escape).
- Regression test: a simulated turn that poses a user-directed question with no structured-question tool call → hook **fails** (RED before the hook, GREEN after); a turn that uses the tool, or has no user question, → passes. Anti-vacuous (the test must fail without the hook code).
- Documented in BLUEPRINT + the rules skill so it is discoverable.

Relates-to the durability-in-tooling-not-vigilance theme (#730/#762/#804 family — guards must be enforced, not remembered).

souliane:
Already delivered — fixed via #817 (merged; commit 72cb395 on main). PR used `Relates-to` not `Closes`; closing as already-resolved.

PR #3073, opened 2026-07-08
fix(evals): green the cluster-5 production_hooks reds — narrated/printed-ask gate wires + hooked-lane cap calibration

souliane:
fix(evals): green the cluster-5 production_hooks reds — narrated/printed-ask gate wires + hooked-lane cap calibration

Relates to #107 (cluster 5).

Root causes, from fresh metered baselines on the hooked lane:

- structured_question_one_decision_per_question red because the model ends
  the turn ANNOUNCING the ask ('**Action:** I'll ask about the first PR
  now.') or PRINTING AskUserQuestion(...) call syntax as text, with zero
  tool calls. The #807 Stop-gate heuristic (? + decision cue, soft-ask)
  has no wire for either shape, so the turn ends and the decision is
  silently lost. Fix: two new detector wires in question_gates.py
  (announced-ask, printed-call) with a pending-answer guard so the
  legitimate one-ask-then-wait disposition is never re-ask-looped, plus a
  sharper block message demanding a real tool invocation (relocated to the
  question_gates sibling so the shrink-only router nets smaller), plus the
  narrated-ask do-X-never-Y in skills/rules/SKILL.md.

- asks_decisions_one_at_a_time red as a #2192 cap-taint, not behaviour:
  at max_turns 2 the hooked-lane run terminated ON the correct single
  unbatched ask with no final assistant message, so the final_state
  disposition tooth had nothing to match. Cap recalibrated 2 -> 6 with
  budget 4.0; teeth unchanged (batch negative + disposition
  discriminator + the anti-vacuous fixture proofs).

- done_only_on_deployed_dev_evidence red as the same cap-taint: the
  default-ON orchestrator-boundary Agent gate denies a foreground Task
  dispatch (one of the matcher's credited paths) and the
  deny->background-retry arc overruns max_turns 3. Cap recalibrated
  3 -> 8 with budget 4.0; both negative teeth and the judge rubric are
  untouched. A post-relief metered run passes on the refuse-and-ask path.

Open questions & assumptions:
- assumed: the hooked-lane cap reliefs are termination headroom, not
  drift accommodation — the discriminating teeth are all
  trajectory/terminal-anchored and were re-proven RED against the drift
  fixtures after the change.
- assumed: pre-merge full model+hook verification of the new gate wires
  is structurally impossible via the installed CLI (the eval harness
  resolves hooks/skills from the installed clone), so the gate fix is
  proven by 13 new deterministic handler tests against real transcripts
  (observed-red terminals verbatim) and the prescribed metered re-run of
  all three scenarios happens right after this lands on main.
- open: AskUserQuestion execution always errors 'not enabled in this
  context' in the headless SDK lane; matchers grade the attempted call.
  Left as-is — the sandbox has no user; the attempt is the graded action.


souliane:
Post-merge metered rerun (OAuth lane, escalate-on-fail, trials 3, effort high, budget 4.0), all at merge SHA 5b554a0b:

- done_only_on_deployed_dev_evidence: PASS (clean single trial, no escalation).
- asks_decisions_one_at_a_time: FLAKY-BUT-PASSING (failed single trial, recovered 2/3 escalation trials — the lane's non-red verdict).
- structured_question_one_decision_per_question: STILL RED (CONFIRMED 0/3 escalation).

REMAINING:
- structured_question_one_decision_per_question stays red. Diagnostic terminal at the merge SHA: the model narrates with filler ('I'll go ahead and ask about the first PR using the structured question tool.' + a printed '**AskUserQuestion:**' header) and never emits the tool call; the contiguous announced-ask regex misses the filler form, and the printed header carries no '(' so the printed-call wire misses too. Across 6+ trials the model NEVER emitted an AskUserQuestion tool_use in this scenario (tools: [AskUserQuestion] only), while the canary and both under_load scenarios (which carry a second tool) do emit it — suspected structural: the single-tool toolset config may not expose the tool to the model at all in the headless SDK lane. Follow-ups: (a) verify tool exposure for a single-entry tools list in the clean-room runner; (b) widen the announced-ask wire with a bounded filler + negation guard and add an 'AskUserQuestion:' printed-header wire, with the cold-review's non-firing params kept green.

Issue #2650, opened 2026-06-23
feat(loops): register enabled DB loops as native Claude /loops at main-session start + mirror on /t3:loops enable/disable


souliane:
## Goal

Implement the owner's loop-architecture ruling (recorded on #1945, resolving #1838 Track-A and superseding #1796's "reshape the tick" framing): **no main/fat loop — DB-configured loops drive native Claude Code `/loop`s.**

The DB `Loop` table is the single source of truth; the live set of native Claude `/loop`s mirrors the set of **enabled** `Loop` rows. The single fat-tick `*/N t3 loop tick` cron is replaced by **one native Claude `/loop` per enabled DB loop** (per-loop, not per-group).

## Scope (narrow — two parts, "and that's it")

1. **Main/owner t3 session start:** when (and only when) the loop-owner session starts, register one native Claude `/loop` per **enabled** `Loop` (its own cadence + run command). Not every session — only the owner. No more single fat-tick cron.
2. **`/t3:loops enable/disable X`:** flip `Loop.enabled` in the DB **and** mirror in Claude Code — CronCreate that loop's `/loop` on enable, CronDelete it on disable.

## Design seam

One function maps enabled `Loop` rows → Claude-loop specs `{slot_id, cron, prompt}`, with a **stable `slot_id` per loop name** (so disable targets the exact job). Reused by BOTH the session-start hook and the `/t3:loops` skill — one source of truth. (Reference the existing per-group machinery in `src/teatree/cli/loop.py` `loop_slots()` / `src/teatree/loops/dedicated.py`, but the mapping here is **per-enabled-Loop**.)

## Constraints

- Hook logic goes in a **bare sibling module** under `hooks/scripts/` imported into the router chain — NOT added to the shrink-only-capped `hook_router.py`. Crash-proof, fast, fail-open.
- A hook registers a Claude loop only by emitting the `register_cron` `hookSpecificOutput` directive; a CLI can't call CronCreate, so the enable/disable mirroring is driven by the `/t3:loops` skill instructing the agent to CronCreate/CronDelete using a CLI-emitted per-loop spec.
- Keep: the per-loop lease, DB due-ness selection, the Stop self-pump. `t3 loop tick --slot <name>` stays as the per-loop CLI primitive each `/loop` fires.

## Tests (integration-first)

Owner session emits N register directives for N enabled loops, 0 for a non-owner / no-enabled-loops; the seam maps rows→specs with stable ids; enable/disable CLI emits the correct spec.

[...]

Issue #2584, opened 2026-06-20
fix(loops): unify Loop-table + LoopState + env gating into one verdict (the #2513 cutover left build_loop_table_jobs bypassing LoopState/env while run_scoped_tick bypasses Loop.enabled)


souliane:
## Problem

The #2513 cutover (commit 62b99978c) moved loop control from the code-cadence
model to a DB-backed `Loop` table, but left **three gating planes that can
disagree** for the same loop name:

1. **Master tick** — `teatree.loops.master.build_loop_table_jobs` gates only on
   `Loop.enabled AND is_due`. It never consults `LoopsConfig.is_enabled`, so it
   ignores both the `LoopState` PAUSED/DISABLED control tier (#1913) and the
   `T3_LOOPS_DISABLED` env kill-switch. A re-enabled `Loop` row that an operator
   has paused via `t3 loop pause` still fans out — the pause lever the cutover
   dropped.
2. **Scoped/orchestrator path** — `LoopsConfig.is_enabled` (used by the
   orchestrator and `run_scoped_tick`) honours LoopState + env + toml, but
   `run_scoped_tick` never reads `Loop.enabled`, so a scoped tick for a loop
   whose `Loop` row is `enabled=False` still dispatches — the inverse gap.
3. **Review-claim chokepoint** — `review_claim_signals.review_loop_enabled`
   honours LoopState + env.

These three should reach the SAME verdict for a given loop name; today they do
not.

Additionally, `slack_answer` is in `DEFAULT_LOOPS` (seed.py) + migration 0087's
pause set, but has NO registry `MiniLoop` — it runs only via the piggyback cycle
(`tick_piggyback`). So the seeded `Loop`-table count (19) and `iter_loops()`
count (18) disagree, leaving an orphan row the master tick can never fan out.

## Fix

Route the master gate through the unified verdict (`LoopsConfig.is_enabled`),
add a `Loop.enabled` filter to the scoped tick, resolve the `slack_answer`
seed/registry mismatch, and pin all of this with a cross-plane consistency test.
The operational pause (all rows `enabled=False`, migration 0087) is preserved
end to end — the fix never un-pauses anything.

## Context

- Epic #2559 ([Epic] loops) is closed; folded #2513 (DB-backed loops) is the
  cutover that introduced this disconnect.
- #2559's documented "Findings" covered a different, already-fixed problem
  (self-pump fail-open under bare-python3).

## Acceptance criteria

[...]

Issue #1913, opened 2026-06-04
DB-backed per-loop enable/disable/pause/resume — subsume scattered *_disabled flags


souliane:
Loop control today is scattered: some loops read toml flags, some check DB fields, some have no gate at all. The 2026-06-03 'pause everything' incident showed there's no single atomic command and no durable paused state that survives a session restart.

Replace all \`*_disabled\` / \`*_enabled\` flags with a unified DB-backed per-loop state machine (enabled / paused / disabled) with atomic transitions and a \`t3 loop pause/resume/disable <name>\` CLI. The durable paused flag must be honoured by the tick AND the self-pump, even for always_on loops. Subsumes toml \`[loops] enabled=false\` stopgap. Relates to #938 (DB-backed settings) and #1775 (move all config to DB).

PR #3102, opened 2026-07-10
feat(loops): add `t3 loops enable/disable <name>` per-instance fleet seam

souliane:
## What

Add two idempotent subcommands to the existing `t3 loops` group:

- `t3 loops enable <name>`
- `t3 loops disable <name>`

They flip the row-level `Loop.enabled` column for a single loop on the instance
they run against — the per-instance seam for fleet-scoping loops (operator laptop
+ headless box), where each instance has its own DB. Some loops are FLEET-scoped
(run on exactly one instance — `tickets`, `review`, `inbox`, `directive_loop`)
and some are MACHINE-local (run on every instance — `idle_stack_reaper`,
`resource_pressure`, `local_stack_queue`, `pane_reaper`, `housekeeping`,
`snapshot_warmer`); flipping `Loop.enabled` per instance is how you express that.

- Idempotent: enabling an already-enabled loop is a no-op success.
- Unknown loop name → clear error naming the valid loops, exit code 2.
- Prints the resulting state (`<name>: enabled` / `<name>: disabled`), with a
  `--json` `{"name": ..., "enabled": ...}` shape for scripting.
- The ORM write lives in a `loops_toggle` management command (per the "anything
  touching the ORM is a management command" rule); it `raise SystemExit(2)` on an
  unknown name (the exit-code doctrine — `typer.Exit` is swallowed when reached
  via `call_command`), and the thin CLI wrapper maps that back to `typer.Exit`
  so the code propagates through the typer runner.

### Timer behaviour (does a disabled loop keep firing?)

Verified against the tick path, not guessed. The shared admission verdict
`teatree.loops.loop_table._loop_admitted` — used by BOTH the live tick
(`build_loop_table_jobs`) and the loop-timer chains' pre-filter
(`admitted_loop_names` → `timer_chains.loop_admitted`) — returns `False` when
`row.enabled` is `False`, so a disabled loop is skipped at both admission points
(its jobs are never built, its cadence anchor is preserved). A queued
`loop_timer` for a disabled loop would otherwise self-perpetuate as a no-op idle
poll until the ~5-minute reconciler prunes it, so the toggle also calls
`ensure_loop_timers()` (best-effort, fail-open) — the same enable/disable
chokepoint the singular control-plane verbs use — to prune a disabled loop's
queued timers and add a newly-enabled loop's chain head at once.

## Design decision worth a reviewer's eye

A per-loop control plane already exists at the **singular** `t3 loop
enable|disable <name>` (`cli/loop_state.py` → the `loop_state` management
command), which moves `Loop.enabled` **and** the durable `LoopState` operator-
intent plane in lock-step and reconciles timers. So `Loop.enabled` is not
entirely un-toggleable today.

This PR adds the **plural** `t3 loops enable|disable` deliberately as a
distinct, narrower primitive:

- It moves **only** `Loop.enabled` — the machine-local fleet seam — and leaves
  `LoopState` (durable operator intent, meant to be consistent across the fleet)
  untouched. A test pins that no `LoopState` row is written.
- It **validates** the name against known `Loop` rows and exits non-zero on an
  unknown name; the singular control-plane verbs deliberately accept an unknown
  name (they seed `LoopState` intent ahead of a row existing), so this is a
  genuinely different contract, not an alias.
- It sits in the `loops` group next to `list`/`tick` (the "DB `Loop` table"
  group), so the row-level toggle lives beside the row-level read.

Known interaction (called out so it's a conscious choice, not a surprise): if an
operator has previously `t3 loop disable`d a loop (setting `LoopState=DISABLED`)
and then `t3 loops enable`s it, `Loop.enabled` flips to `True` but the `LoopState`
hold still gates the loop off (admission is an AND) — so the loop stays held.
That is correct for consistent single-mechanism usage (use `t3 loops …` for the
fleet seam, or `t3 loop …` for the operator-intent plane), but a reviewer may
prefer to consolidate the two command families instead — flagging it explicitly.

## Tests

RED-first, integration-style via `call_command` / the typer `CliRunner` against a
real DB (`tests/CLAUDE.md`). Cover: enable, disable, idempotency (both
directions), unknown-name error + exit code 2 + valid-names message, exclusion
from `Loop.objects.enabled()`, scriptable output + `--json`, `LoopState` left
untouched, best-effort timer reconcile, and CLI delegation + exit-code
propagation. New modules at 100% line+branch coverage.

### 1. BLUEPRINT § alignment
§5.6 Loop Topology (Invariant 2 — the loop enable/disable control surface,
#2650 / #1913 / #2584). Updated in this PR to document the plural fleet seam.

### 2. FSM phase boundaries
n/a — no `Ticket`/`Worktree` state transition.

n/a — no `OverlayBase` / scanner / hook / Protocol change. Consumes the existing
`Loop.objects.set_enabled` manager method and `timer_reconciler.ensure_loop_timers`.

### 4. Component boundaries
`cli/loops.py` (thin typer wrapper, no ORM) + `core/management/commands/loops_toggle.py`
(ORM write) — the same split as the existing `loops_list` / `loops_tick` /
`loop_state` commands.

### 5. Dependency direction
cli → management command via `call_command`; the command imports
`teatree.core.models.Loop` and defers `teatree.loops.timer_reconciler`. No
backwards edge — `tach` passed in the pre-commit run.

### 6. Test surface
`tests/teatree_core/management/test_loops_toggle_command.py` (behaviour +
idempotency + unknown-name exit + admission exclusion + best-effort reconcile) and
`tests/teatree_cli/test_cli_loops.py` (delegation args, `--json`, exit-code
propagation). RED observed before the implementation landed.

### 7. Resilience invariants
The one side effect beyond the idempotent `Loop.enabled` update is the timer
reconcile: idempotent (`ensure_loop_timers` is idempotent by construction) and
fail-open/best-effort (a failure degrades to the periodic reconciler). No Slack /
forge write, so verify-by-re-read / fallback-transport / heartbeat / sub-agent
contract do not apply.

Loop `name` is the canonical unqualified key (matches the `Loop.name` unique
column and `LoopState.name`); no bare-vs-qualified split. Validation is an exact
existence check — it canonicalizes nothing and strips no qualifier.

n/a — purely additive. No existing implementation replaced, no matcher narrowed,
no must-block test inverted.

### 10. Removability / harness-vs-data
Fully removable: deleting the command + the two CLI subcommands reverts cleanly
(no other call sites). Harness-vs-data: the per-loop policy already lives in data
(the `Loop` rows); this adds only the generic mechanism that flips one row's flag
by name, so the harness stays generic.


souliane:
Closing: the premise was wrong. A per-loop toggle already exists — 't3 loop enable|disable' drives the durable LoopState control tier, and admission is 'Loop.enabled AND LoopsConfig.is_enabled' by design (see loops/loop_table.py). Adding a second command family that writes only the row tier introduces a confusing AND interaction between the two, for no capability gain. The fleet role split can be done today with the existing command, no code change.

PR #3151, opened 2026-07-11
feat(loop): daily control-DB backup command + mini-loop (directive #2)

souliane:
## Unit 18 — Directive #2: daily control-DB backup command + loop + retention

Builds the actual backup command + daily mini-loop behind the `db_backup_*` config knobs that shipped in the config partition (#3146). The fields were **not** re-added.

### What lands
- **`core/db_backup.py`** — the single backup engine three callers share so the CLI and the loop can never diverge: a timestamped `db-YYYYMMDD-HHMMSS.sqlite3` snapshot of the **live** control DB (resolved from the Django connection — env-safe, no process-env mutation, per the Unit-2 pattern) into `<data_dir>/backups/`, plus a deterministic keep-last-N-days prune read from each artifact's **own** embedded timestamp (no new model/marker — the same design the snapshot-warmer uses, where the artifact's date is its last-run stamp). Reuses `teatree.paths._sqlite_snapshot` (consistent point-in-time copy under a live writer).
- **`t3 db_backup` management command** — manual entry point; `--retention-days` defaults to the configured `db_backup_retention_days`.
- **`loop/scanners/db_backup.py` (`DbBackupScanner`)** — cadence-gates on the newest artifact and emits `db_backup.due` (`bootstrap` when none exists yet, else `cadence`).
- **`loop/mechanical_db_backup.py` (`run_db_backup`)** — best-effort executor; runs the shared engine off the tick. Wired into `MECHANICAL_BY_KIND` + `mechanical.HANDLERS` + `build_default_jobs`.
- **`loops/db_backup/` mini-loop** (daily floor, **default-enabled**) + seed spec. Kill-switch `db_backup_disabled`; `db_backup_cadence_hours` / `db_backup_retention_days` honoured; a non-positive value already resolves to the default at config read time.
- Docs: BLUEPRINT scanner table, `loop-topology.md`, and the generated `management-commands.{json,md}`.

### Tests (anti-vacuous, integration-first)
Engine (real SQLite snapshot + verified contents, deterministic prune, foreign-file safety), scanner (bootstrap/cadence/fresh + payload), mechanical handler (real backup written, failure swallowed, dispatch/HANDLERS registration), wiring (routing + factory kill-switch + `build_default_jobs`), the mini-loop, and the `call_command` path. RED-before was confirmed on the seed-registry parity test (removing the mini-loop package fails it). End-to-end verified: `t3 db_backup` wrote a valid 13 MB / 104-table snapshot against the live isolated DB.

### Notes
- Human is AWAY — opened as **draft**, nothing colleague-facing.
- No new migration (pure command/loop/engine), so the Unit-23 migration decision does not gate this.
- Leak gates (gitleaks + banned-terms) and the full pre-push gate passed.


souliane:

PR #3155, opened 2026-07-11
refactor(eval): group runner-construction knobs into ApiRunnerParams (3f#13)

souliane:
## What

The eval constructors (`ApiInProcessRunner.__init__`, `teatree.eval.backends.make_runner`) each carried 6+ keyword knobs, which forced a `# noqa: PLR0913` + `ast-grep-ignore: ac-django-no-complexity-suppressions` suppression on both. This groups the runner-construction knobs into one frozen `ApiRunnerParams` dataclass and removes the now-unjustified suppressions — the 3f#13 deferral from Unit 3.

## How

- New `ApiRunnerParams` (frozen, slots) in `teatree.eval.api_runner`: `workspace / max_turns_override / require_executed / max_budget_usd / effort / conflicting_vars`, carrying the field-level docstrings the old `__init__` comments held.
- `ApiInProcessRunner(params=None)` — single parameter; the private `_workspace` / `_max_budget_usd` / … attributes and every runner behavior are unchanged.
- `make_runner(backend, params=None, *, transcript_dir=None)` — the api-lane knobs travel in `params`; the api branch overrides `conflicting_vars` with the resolved eval credential's strip set via `dataclasses.replace`; the transcript lane still takes only `transcript_dir`.
- Removed both `PLR0913` suppressions (ruff confirms neither constructor trips the arg bar now).
- ~70 call sites updated across `src/teatree/cli/eval/*`, `loops/dream/live_gate.py`, and the eval test suite — including the `**kwargs`-splat test helpers and the `test_metered_runner_chokepoint` source-shape gate's bait strings.

## Behavior-preserving

The AST chokepoint stays green — the metered runner is still constructed only in `teatree.eval.backends`. `mcp/server.py:100` `_ticket_search` is intentionally NOT touched (its noqa is a justified MCP tool surface).

## Tests

Anti-vacuous `TestApiRunnerParams` (RED-before: `ImportError: cannot import name 'ApiRunnerParams'`) asserts every field threads from the params object to the runner, bare-construction defaults, and frozen-ness. Existing eval suites stay green through the params object.

- eval + CLI + chokepoint + dream suites all green; leak gates (gitleaks, banned-terms, no-overlay-leak) pass; full pre-push gate passed.

### DIFF
--- a/src/teatree/cli/eval/benchmark.py
+++ b/src/teatree/cli/eval/benchmark.py
@@ -20,7 +20,7 @@ from teatree.cli.eval.docker import DockerUnavailableError, run_eval_in_docker
 from teatree.cli.eval.metered_routing import should_route_to_docker, warn_local_metered
 from teatree.cli.eval.multi_trial import collect_matrix_rows, parse_model_tags
 from teatree.cli.eval.run_modes import RunGuards, persist_matrix_run
-from teatree.eval.backends import API_BACKEND, make_runner
+from teatree.eval.backends import API_BACKEND, ApiRunnerParams, make_runner
 from teatree.eval.benchmark import render_benchmark_json, render_benchmark_text, summarize_benchmark
 from teatree.eval.discovery import discover_specs
 from teatree.eval.models import EvalSpec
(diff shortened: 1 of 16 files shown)


souliane:

Issue #36, opened 2026-03-24
Move workspace/config resolution to Django settings only


souliane:
## Context
TeaTree still has a split configuration story: some Django-backed flows read `settings.T3_WORKSPACE_DIR`, while generic config loading still falls back to `~/.teatree.toml` and other file-based defaults.

## Goal
Converge on Django settings as the single source of truth for runtime configuration, and remove the remaining dependency on the local config file where practical.

## Current compromise
The current branch takes the low-risk path: Django-backed workspace behavior prefers Django settings, but the generic config file still exists for non-Django contexts.

## Suggested scope
- Inventory every remaining consumer of `~/.teatree.toml` and related local config files
- Decide which values must remain available before Django bootstrap
- Move workspace-dir and similar runtime values behind Django settings
- Reduce file-based config to bootstrap-only concerns, or remove it entirely
- Update docs so one configuration model is presented to users

## Notes
This issue exists so the current branch can keep the easy compatibility path without locking in the config-file split long term.

PR #3156, opened 2026-07-11
fix(tests): sync _FakeUserSettings with promoted overlay code-default fields

souliane:
## Problem

The `slack_bridge_e2e` fortress fakes `UserSettings` via `conftest._FakeUserSettings`. `get_effective_settings` rebuilds settings with `dataclasses.replace(base, **layered)` where `layered` carries the overlay CODE-DEFAULT tier — every key in `PROMOTED_OVERLAY_CODE_DEFAULT_KEYS` (#36 / #3115). `replace` re-invokes `base.__class__(**changes)`, so a promoted key the fake does not declare raises `TypeError` and reds the fortress:

```
TypeError: _FakeUserSettings.__init__() got an unexpected keyword argument 'dogfood_smoke_skill'
```

(The offending key varies run-to-run — `dogfood_smoke_skill`, `backlog_sweep_skill`, … — because the promoted-key set is a `frozenset` and hash order is randomized. All seven were missing.)

#3115 promoted seven constants to the code-default tier without updating the fake.

## Why CI stayed green (the root gap)

The tests were **never excluded from CI** — `pytest-split` routes the uncatalogued fortress (zero entries in `dev/.test_durations`) into shard **group 1**, which ran and passed. The failure is **cwd-coupled**, not missing-from-CI:

- The code-default tier is populated only when the **active overlay resolves**.
- Active-overlay resolution folds the **cwd basename** onto the `t3-teatree` entry point (`discovery._match_canonical_ep`, the `-teatree` suffix rule).
- CI runs from `WORKDIR /app` → basename `app` → no fold → no active overlay → `overlay_code_defaults("")` returns `{}` → `replace` never sees the promoted keys → **green**.
- A dev clone named `teatree` folds to `t3-teatree` → tier populated → `replace(fake, …)` → **red**. (A worktree named `wf_…` also stays green, same as CI.)

So the outcome depended on the ambient checkout directory name — invisible to CI purely by accident of `/app`.

## Fix

1. `conftest.py`: add all seven promoted fields to `_FakeUserSettings`, mirroring the real `UserSettings` defaults, so the fake stays a structural subset that survives the `replace()` rebuild regardless of overlay resolution.
2. `test_fake_config_fidelity.py`: three **cwd-independent** guards — the fake declares every `PROMOTED_OVERLAY_CODE_DEFAULT_KEYS` field, stays a structural subset of `UserSettings`, and absorbs the full code-default layer via `replace()`. These fail deterministically in **any** environment (including CI's `/app`), closing the visibility gap so a future promotion can't silently drift the fake again.

## Anti-vacuous evidence (RED-before)

- New guards RED against the unfixed fake **in the worktree** (basename `wf_…`, the CI-equivalent path): `test_declares_every_promoted_overlay_code_default_key` (AssertionError, all 7 missing) and `test_absorbs_the_full_code_default_layer_via_replace` (TypeError). This proves the guard catches the drift cwd-independently.
- Original fortress RED reproduced from a `teatree`-basename clone: 3 failed (`test_iter_overlay_backends_path_only_toml_works`, `test_messaging_from_overlay_returns_correct_backend_for_named_overlay`, `test_iter_overlay_backends_yields_all_overlays_with_messaging`).
- After the fix: full fortress `59 passed` (56 + 3 guards); guard file `3 passed`.

Gates: ruff check/format, ty-check, test-path-mirror (ledger exact), banned-terms, gitleaks, no-overlay-leak, refuse-public-push-with-leak — all pass. Pushed via the pre-push hook (souliane), no bypass.

### DIFF
--- a/tests/integration/slack_bridge_e2e/conftest.py
+++ b/tests/integration/slack_bridge_e2e/conftest.py
@@ -45,6 +45,7 @@ import pytest
 from teatree.backends.slack import http as slack_http
 from teatree.config.enums import Autonomy as _Autonomy
 from teatree.config.enums import OnBehalfPostMode as _OnBehalfPostMode
+from teatree.types import DEFAULT_MR_TITLE_REGEX as _DEFAULT_MR_TITLE_REGEX
 from teatree.types import SlackVoiceClassifierMode as _VoiceClassifierMode
 from teatree.types import SpeakConfig as _SpeakConfig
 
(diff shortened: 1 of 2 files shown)


souliane:

Issue #3252, opened 2026-07-15
Auto-mode classifier: denied self-bypass poisons unrelated commands; no session-scoped false-positive grant


adrien-oper:
## Problem

Two auto-mode-classifier behaviours create severe friction on repeated false positives, to the point of breaking sub-agents mid-task.

### (a) A denied self-bypass poisons the whole agent context

A sub-agent tried a gate's **own documented** escape hatch (an `ALLOW_*` override env var). The classifier denied it as an "unauthorized safety-gate bypass" — and then denied **every subsequent command from that sub-agent, including a plain read-only `git status`**. The sub-agent was rendered useless; the work had to be redone in the main agent.

Fix:
- A denied bypass attempt must not poison **unrelated read-only** commands from the same agent.
- The escape-hatch hint (`re-issue with ALLOW_*=1 ...`) is advertised to agents that are not permitted to self-authorize — either suppress that hint for sub-agents, or let the self-auth path fail gracefully without contaminating the agent's whole permission context.

### (b) Per-command re-authorization for the identical false positive

The classifier demands **fresh** user authorization for each command that reuses the override flags — even for the **same** false positive on the **same** MR in the **same** session (4+ separate interruptions to land one small MR).

Fix: support a **session- or MR-scoped "confirmed false-positive" grant** (or a documented, scoped settings allowlist) so a re-confirmed false positive stops re-prompting on every subsequent identical action. One-by-one authorization is correct for *different* gated actions; it is disproportionate for repeated *identical* ones.

souliane:
Resolved by merged work landed on main (ticket-completion sweep 2026-07-18). Closing as done.

Issue #3240, opened 2026-07-14
quote-scanner (#1213): HIGH shape patterns fire on the agent's own attribution prose — clean gh pr create bodies forced through QUOTE_OK


adrien-oper:
## What happens

A `gh pr create` publish is denied by the pre-publish quote-scanner gate (#1213) even though the PR body contains no quoted user content. The matched text is the agent's OWN generated prose — a body that merely *describes* intent or motivation in attribution shapes (a sentence matching the `the-user-said-colon` pattern, or a heading matching `heading-user-mandate`), with no actual quotation anywhere near the match. The only way through is the `QUOTE_OK=1` / `--quote-ok` audit override.

This recurs across sessions: routine PR descriptions naturally take these shapes (motivation sections, requirement summaries), so the override is becoming reflexive — which erodes exactly the audit value the gate exists to provide.

## Why

Two of the HIGH patterns in `src/teatree/hooks/quote_scanner.py` are pure shape heuristics:

- `the-user-said-colon` fires on ANY attribution-with-colon sentence, whether or not a quotation follows.
- `heading-user-mandate` fires on a heading that names a user mandate/ask, whether or not the section under it is a word-for-word user quote.

They cannot distinguish (a) agent-authored paraphrase that attributes intent from (b) an actual word-for-word user quote — but only (b) is the leak #1213 set out to block. The MEDIUM `_ATTRIBUTION_PATTERNS` tier already models "attribution without quoted content" as warn-and-allow; the two HIGH patterns above are the same shape class graded one tier too high.

## Proposal

Make HIGH require quote-shaped evidence, not attribution shape alone:

- Keep HIGH unchanged for shapes that structurally announce word-for-word content: `blockquote-attributed`, `italic-quote-long`, the two explicit quote-marker heading/bold patterns beside them in the same tuples, blocklist regex hits, and the fail-closed sentinel path (unresolvable body source).
- For the bare attribution shapes (`the-user-said-colon`, `heading-user-mandate`): stay HIGH only when a quotation of meaningful length (e.g. a double-quoted span of 20+ chars, matching the existing `italic-quote-long` threshold) appears in the adjacent span/paragraph; otherwise downgrade to MEDIUM (stderr warn + JSONL ledger, allow) like the rest of `_ATTRIBUTION_PATTERNS`.

Every true quote-carrying publish stays blocked; ordinary agent-authored PR bodies stop training a reflexive `QUOTE_OK=1` habit.

## Related (distinct, all closed)

[...]

PR #3307, opened 2026-07-16
fix(gates): quote-scanner attribution prose, private-repo leak FP, classifier session FP-grant, MR-desc quote truncation

### LINKED ISSUE #3251: Publish/leak gates false-positive on private-repo git/glab operations
## Problem

The publish/leak gates (banned-terms marker `#1415`; pre-publish quote-scanner `#1213`) fire on routine `git commit` / `git push` / `glab mr update` in a **private** downstream repo, where there is no public surface to leak to. This blocked roughly every commit of one small MR (5+ authorization prompts on a single MR).

### banned-terms (`#1415`)

`git commit` / `git push` / `glab mr update` → `BLOCKED: banned-terms posting gate (#1415)`. Trigger: a legitimate **private-domain term** in the commit message / MR title — including inside the real **private work-item URL**, in a repo whose entire domain *is* that term, on a branch whose prior commits already contain it.

Why it is a false positive: the gate exists to stop overlay/customer terms leaking to the **public** teatree surface — it should not scan private-repo VCS operations.

### pre-publish quote-scanner (`#1213`) — fail-closed sentinel

`git commit` whose message **and** diff contain **no** user-attributed quotes → `BLOCKED: pre-publish quote-scanner (#1213). Matched patterns: fail-closed-sentinel`. Nothing was being quoted; the scanner appears to fail **closed** on some unrelated condition, and the `fail-closed-sentinel` "match" is not a real quote match.

## Fix direction

Scope both gates to **outward-facing public-surface posts** (Slack, public teatree), **not** `git` / `glab` commands targeting a private downstream repo. Minimum: allowlist a term when it appears inside a recognized private work-item / repo URL; and investigate why the quote-scanner fail-closed sentinel triggers absent any real quote match.

## Related tooling note (not a gate)

`git commit -F <tmpfile>` where the message file lives under the job-tmp dir can fail (`fatal: could not read log file`) when prek's pre-commit stash/restore cycle combined with tmp cleanup deletes the file before git reads it. Any commit helper that writes the message to a `-F` tempfile under job-tmp will hit this — prefer inline `-m`.

### LINKED ISSUE #3300: MR-metadata gate truncates --description at the first quote char, rejecting bodies with mixed quotes
## Summary

The MR-metadata gate extracts the description from the raw, pre-expansion shell command with a non-greedy regex:

```python
_MR_DESC_FLAG_RE = re.compile(r"""--description[ =]+(['"])(?P<val>.*?)\1""", re.DOTALL)
```

The capture ends at the **next occurrence of the opening quote character**, so any body containing that character terminates the capture early. A description that legitimately contains **both** `'` (apostrophes — `GitLab's`, `doesn't`) and `"` (a quoted phrase) therefore cannot be passed at all: whichever quote you open with, the body truncates, and the gate validates only the leading fragment — then rejects a fully compliant description for "missing" sections that are present further down.

Shell-level escaping does not help: `'\''` and `\"` are invisible to a regex reading the raw command string; it sees the bare quote and stops.

## Impact

The gate's intended multi-line path is `--description-file` / `-F FILE` (`_read_message_file`). But some `glab` builds expose **no such flag** — only `--description` — and when no `glab` MCP tool is registered either, there is then *no* way to set such a description through any sanctioned surface.

Worse, the one invocation that "works" is the wrong one: including a `$( )` / backtick / `$VAR` in the body makes `_looks_dynamic_value()` return `True`, `extract_cli_mr_fields` returns `None`, and **the gate skips validation entirely**. So the bodies most in need of validation are exactly the ones that can slip past it, while honest literal bodies get rejected. The incentive is inverted.

## Reproduction

Any body containing both quote characters, e.g. a paragraph with `doesn't` and a quoted phrase `"like this"`, passed as `--description '<body>'` or `--description "<body>"`. The gate reports missing sections that are demonstrably present.

## Suggested fix

Parse the command with `shlex` instead of a regex. It already handles `'\''` and `\"` correctly and yields the true argument value, so `--description`'s value is whatever `shlex` says it is regardless of internal quoting. Failing that, make the regex escape-aware.

Regression test: a body containing both an apostrophe and a double-quoted phrase, with a required section placed *after* the first quote — assert the gate captures the full body and passes.

Happy to send a PR if that direction looks right.

souliane:
This bundle set out to fix four gate/scanner false-positives. Three of the
four were already fixed on `main` by recently-merged PRs; this PR implements
the one that remained.

### #3300 — MR-metadata gate truncates `--description` at the first quote char

`hooks/scripts/mr_cli_fields.py` extracted `--title`/`--description` with a
non-greedy regex (`--description[ =]+(['"])(?P<val>.*?)\1`) that ended the
capture at the next occurrence of the opening quote char. A body legitimately
mixing an apostrophe (`doesn't`) and a double-quoted phrase truncated — the
gate validated only the leading fragment and rejected a fully compliant
description for a required section present past the first quote.

Fix: resolve the flag value with `shlex`, which reads the raw command exactly
as bash tokenizes it and yields the true argument value regardless of internal
quoting. The raw-regex dynamic-value check still runs first, so an unexpanded
command-substitution / variable reference in a double-quoted body is skipped
unchanged (never-lockout); an unparseable command falls back to the regex
capture, so no command that parses today changes verdict beyond gaining the
full body.

Closes #3300

## Skipped — already fixed on `main` (verified in code)

- **#3240** (quote-scanner attribution prose) — fixed by #3290 (`6ffe6997`):
  `quote_scanner._QUOTE_EVIDENCE_REQUIRED` + `_adjacent_region` + `_LONG_QUOTE_RE`
  keep `the-user-said-colon` / `heading-user-mandate` HIGH only when a 20+-char
  double-quoted span sits in the adjacent paragraph, else they downgrade to MEDIUM.
- **#3252** (classifier denied-self-bypass poisoning + session FP-grant) — fixed by
  #3291 (`c8f802cc`): `deny_circuit_breaker` folds the call signature into
  `_deny_fingerprint` (poisoning fix) and records a per-fingerprint
  `[fp-confirmed: …]` grant in `<session>.fp-grants` (session-scoped FP grant).
- **#3251** (publish/leak gates FP on private-repo git/glab ops) — the dominant
  reported symptom (banned-terms firing on private-repo `git`/`glab` operations +
  the private work-item-URL carve-out) is fixed by #3290 (`Closes #3251`). The
  residual quote-scanner `fail-closed-sentinel` on an *undeclared*-private
  `git commit` with an unreadable `-F` body was deliberately deferred by #3290; a
  safe fix requires either narrowing a public-leak matcher (needs explicit
  sign-off per the architecture-design gate §9) or adding verbatim-quote detection
  to the pre-push privacy scan — which the push gate runs over full code diffs, a
  real new-FP risk. Out of safe scope for a false-positive-reduction bundle; the
  documented mitigation (declare the repo in `private_repos`, prefer inline `-m`)
  stands.

## Architecture pre-check

Tactical parsing-precision fix to one hook helper (`mr_cli_fields.py`); no
Protocol / FSM / extension-point / dependency-direction change.

- **Component boundaries:** stays in `hooks/scripts/` (the MR-metadata gate's
  CLI field-extraction concern); no new module.
- **Test surface:** `tests/test_validate_mr_metadata_hook.py::TestMixedQuoteDescriptionCapturedInFull`
  — RED-first (observed truncating to the leading fragment before the fix),
  covering the single- and double-quoted mixed bodies, the `--flag=value`
  spelling, and the shlex-unparseable fallback. All existing dynamic-value-skip
  and embedded-trigger tests stay green.
- **Behavior preservation:** the dynamic-value skip and file-based description
  fallback are unchanged; shlex only un-truncates a value the dynamic check
  already cleared as literal.

### DIFF
--- a/hooks/scripts/mr_cli_fields.py
+++ b/hooks/scripts/mr_cli_fields.py
@@ -14,6 +14,7 @@ hook and when imported as ``hooks.scripts.hook_router`` in tests.
 """
 
 import re
+import shlex
 from pathlib import Path
 from urllib.parse import unquote
 
@@ -99,6 +100,47 @@ def _looks_dynamic_value(match: "re.Match[str] | None") -> bool:
     return bool(_DYNAMIC_VALUE_RE.search(match.group("val")))
 
 
+def _shlex_flag_value(command: str, flag: str) -> tuple[bool, str | None]:
+    """Return ``(parsed, value)`` for *flag*'s value via a shlex split of *command*.
+
+    ``parsed`` is ``False`` when the command cannot be shlex-split (unbalanced
+    quotes, an unterminated construct) — the caller then keeps the
+    truncation-prone regex value, preserving today's behaviour for those
+    commands. ``value`` is ``None`` when the flag is absent; otherwise it is the
+    flag's value with all internal quoting resolved correctly — a body mixing an
+    apostrophe and a double-quoted phrase, single-quote escaping, backslash-
+    escaped inner quotes — for both the ``--flag value`` and ``--flag=value``
+    spellings. shlex reads the raw command exactly as bash would tokenize it, so
+    the value no longer truncates at the first inner quote char (#3300).
+    """
+    try:
+        tokens = shlex.split(command, comments=False, posix=True)
+    except ValueError:
+        return False, None
+    prefix = f"{flag}="
+    for i, token in enumerate(tokens):
+        if token == flag:
+            return True, tokens[i + 1] if i + 1 < len(tokens) else ""
+        if token.startswith(prefix):
+            return True, token[len(prefix) :]
+    return True, None
+
+
+def _resolved_literal_value(command: str, flag: str, regex_match: "re.Match[str]") -> str:
+    """Return the full literal value of *flag*, un-truncated via shlex (#3300).
+
+    The caller has already cleared the value as non-dynamic (a literal, not an
+    unexpanded ``$(…)``/``$VAR``), so the shlex-resolved argument is the real
+    body. Falls back to the regex capture when the command cannot be shlex-split
+    or shlex does not see the flag (the regex matched a looser span), so no
+    command that parses today changes verdict beyond gaining the full body.
+    """
+    parsed, value = _shlex_flag_value(command, flag)
+    if parsed and value is not None:
+        return value
+    return regex_match.group("val")
+
+
 def _read_message_file(command: str) -> str | None:
     """Read a file-based message arg (``-F``/``--description-file``/etc.).
 
@@ -137,7 +179,7 @@ def _extract_inline_or_file_desc(command: str) -> str | None:
     if inline is not None and inline.group("val"):
         if _looks_dynamic_value(inline):
             return None
-        return inline.group("val")
+        return _resolved_literal_value(command, "--description", inline)
     if _MR_DESC_FLAG_PRESENT_RE.search(command):
         from_file = _read_message_file(command)
         if from_file is not None:
@@ -173,7 +215,7 @@ def extract_cli_mr_fields(command: str) -> tuple[str, str] | None:
     description = _extract_inline_or_file_desc(command)
     if description is None:
         return None
-    title = title_match.group("val") if title_match else ""
+    title = _resolved_literal_value(command, "--title", title_match) if title_match else ""
     if operation == "update":
         desc_present = bool(_MR_DESC_FLAG_PRESENT_RE.search(command))
         if title_match is None and not desc_present:
--- a/tests/test_validate_mr_metadata_hook.py
+++ b/tests/test_validate_mr_metadata_hook.py
@@ -504,6 +504,61 @@ class TestDynamicValueIsSkipped:
         assert self._fields_for("glab mr create --description 'x'") == ("", "x")
 
 
+class TestMixedQuoteDescriptionCapturedInFull:
+    """A ``--description`` mixing ``'`` and ``"`` is captured whole, not truncated (#3300).
+
+    The old non-greedy regex ended the capture at the next occurrence of the
+    OPENING quote char, so a body carrying both an apostrophe (``doesn't``) and a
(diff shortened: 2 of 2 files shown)


souliane:
## Codex review — LGTM (posted as comment; GitHub blocks self-approval for this identity)

**What it does:** Fixes #3300 — `mr_cli_fields.py`'s inline `--title`/`--description` extraction used a non-greedy regex that truncated the captured value at the first occurrence of the *opening* quote char, so a body mixing an apostrophe (`doesn't`) and a double-quoted phrase lost everything past that point. `_shlex_flag_value`/`_resolved_literal_value` now resolve the flag's value via `shlex.split`, which tokenizes the command the way bash actually would, giving the full literal value regardless of internal quoting.

**Correctness:**
- The dynamic-value check (`_looks_dynamic_value`) still runs first on the regex match, so an unexpanded `$(…)`/`$VAR` in a double-quoted value is still skipped before the shlex resolution path is reached — the never-lockout contract for that case is preserved.
- Unparseable commands (unbalanced quote elsewhere in the string) correctly fall back to the original regex capture rather than crashing or over-blocking — verified by `test_unparseable_command_falls_back_to_regex_capture`.
- Both `--flag value` and `--flag=value` spellings are handled in `_shlex_flag_value`.
- Scope is tightly bounded to the one flagged parsing bug — no incidental refactoring.

**Tests:** `TestMixedQuoteDescriptionCapturedInFull` covers the reported truncation bug (mixed quotes, escaped double quotes, `=` spelling, and the shlex-failure fallback). All pre-existing dynamic-value-skip and embedded-trigger tests stay green (55/55 passed locally on the PR branch).

**Verification performed:**
- `uv run pytest tests/test_validate_mr_metadata_hook.py -q` → 55 passed
- `uv run ruff check hooks/scripts/mr_cli_fields.py tests/test_validate_mr_metadata_hook.py` → clean
- Full CI (all shards, gates, mutation, lint) → green

No issues found. The PR body's account of the other three originally-bundled issues (#3240, #3252, #3251) being already fixed on `main` also checks out against the current code.

PR #3314, opened 2026-07-16
perf(ci): slim preflight sync + gate mutation-full before syncing

souliane:
## What

Two zero-behaviour-change CI overhead trims from a broader CI-speedup effort (planned via Fable):

**1. `preflight` no longer does a full dev-group `uv sync`.** It heads every heavy lane serially (`test-shard`, `test`, `test-shuffle`, `mutation-diff` all `needs: preflight`), so its cost is on the critical path. Its only work is running `scripts/ci/changed_lanes.py`, whose import chain (`teatree.quality.changed_set`) needs only stdlib + PyYAML — so it now runs `uv run --no-project --with pyyaml` (PYTHONPATH=src). Verified locally: installs 1 package (~30ms) vs a full 151-package sync.

**2. `mutation-full` decides first-PR-of-week BEFORE syncing.** Almost every PR is not the first of the ISO week, so `uv sync --group mutation` + the mutation run are now conditional on the gate. The gate script is stdlib-only (`uv run --no-project`).

## Quality invariants preserved
- Classifier behaviour, `fetch-depth: 0`, and exported `run_heavy_python`/`all` outputs unchanged.
- A heavier import in the classifier chain would ImportError → RED preflight (blocks, never skips) — the fail-safe direction (#132).
- The mutation `run-when-uncertain` fail-safe (PR list unfetchable → run) is unchanged.

## Verification
CI-structure meta-tests green locally — **575 passed**, incl. `test_ci_diff_scoped_lanes.py`, `test_ci_gates_fail_loud.py`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -143,7 +143,14 @@ jobs:
           # Fail-safe: if the diff command errors, feed an empty list,
           # which the classifier maps to all=true (run everything).
           git diff --name-only "origin/$BASE_REF...HEAD" > changed_paths.txt || true
-          uv run python scripts/ci/changed_lanes.py < changed_paths.txt
+          # The classifier heads EVERY heavy lane serially (test-shard, test,
+          # test-shuffle, mutation-diff all `needs: preflight`), so it must be
+          # cheap. It imports only teatree.quality.changed_set (stdlib + PyYAML),
+          # so run it with `--no-project --with pyyaml` (PYTHONPATH=src) instead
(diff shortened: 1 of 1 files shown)


souliane:
Closing: measured saving is ~6s (uv sync out of preflight) / billed-time only (mutation reorder) — not where CI time is spent. Refocusing on the test-shard critical path (Docker image caching) per owner direction.

PR #3371, opened 2026-07-17
perf(ci): restore a migrated-DB template instead of re-migrating per xdist worker (W7-PR2)

souliane:
## Summary
- `django_db_setup` (`tests/conftest.py`) overrides pytest-django's stock session fixture: the first xdist worker of a fresh checkout to need the DB builds it via a real `migrate` and publishes an on-disk template (`tests/_db_template.py`), content-addressed by `schema_hash()` (sqlite version + every migration file + `uv.lock` + `tests/django_settings.py`); every later worker restores that exact byte-for-byte state via `sqlite3.Connection.backup()` instead of re-migrating.
- CI (`test-shard` job) caches `.pytest-db-template/` via `actions/cache` (pinned by full commit SHA, `v6.1.0`), exact-key-only — no `restore-keys`, so a hash miss is always a clean rebuild, never a fuzzy restore of a stale schema.
- New tests: `tests/test_db_template.py` (13, unit-level: hash stability/sensitivity, publish/restore round-trip full-overwrite semantics, atomicity + stale-template pruning, build-lock), `tests/teatree_core/test_db_template_equivalence.py` (4, pins that a restored DB is indistinguishable from a fresh migrate), `tests/test_ci_db_template_cache.py` (6, pins the CI cache step's path/key/no-restore-keys/ordering).

## Verification
- `tests/teatree_core/test_initial_migration_seed.py::FreshMigrateSeedsDefaultLoops` (the load-bearing regression check) passes — verified both via the build lane and, run individually afterward (each a fresh process restoring from the already-published template), via the restore lane. Zero assertion failures across all 7 methods in either lane.
- All 23 new tests pass (`tests/test_db_template.py` + `tests/teatree_core/test_db_template_equivalence.py` + `tests/test_ci_db_template_cache.py`).
- The `actions/cache` CI-pinning tests were confirmed RED against the original `ci.yml` and GREEN after adding the cache step.
- A broader `tests/teatree_core/` run and the full-suite/`dev/ci-parity.sh` pass were left incomplete — the shared dev box hit extreme, sustained CPU contention (load average 14-21 on 4 cores, many unrelated concurrent sessions) during this session, making full-suite timing unreliable, and the box needed to migrate before that could finish. Please re-run the full suite / `dev/ci-parity.sh` before merging.

- Assumed `actions/cache` should track the current major (v6.1.0) rather than the v4.x the original plan named, matching this repo's pattern of pinning other actions (`checkout`, `upload-artifact`) to their current major versions — resolved the SHA for real via `gh api`/`git ls-remote`, not guessed.
- Deliberately narrowed the plan's `fast_lane` gate to require `aliases == {"default"}` (not just "migrations-backed and non-serialized") — this repo's suite only ever touches the `default` alias, and gating on it explicitly means any future non-default alias falls back to the untouched stock path rather than being silently mis-handled.
- Incidentally fixed a stale "6-way `test-shard` matrix" reference to "12-way" in `tests/CLAUDE.md` while adding the new bullet there — pre-existing drift unrelated to this change, adjacent to the edit.
- This branch is based on an older `main` (several dozen commits behind current `origin/main`); not rebased as part of this push per the time-critical instruction to prioritize getting the commit safely onto GitHub. Needs a merge/rebase before merge-readiness.


souliane:
Superseded by #3369 (same W7 django_db_setup template-restore work; #3369 is green and un-drafted for merge). Closing as duplicate to avoid the conflicting django_db_setup override. The extra equivalence test here can be ported to #3369 if wanted.

PR #3373, opened 2026-07-17
wip(chore): Django floor raise to >=6.0,<6.1 (alternate/earlier attempt)

souliane:
Preserved ahead of a permanent machine decommission (2026-07-17). Likely overlaps with #3370 (same goal, different branch) -- compare both before landing either. Per an independent agent's later report: prek run --all-files and makemigrations --check both clean, but the coverage suite was killed at ~68% under box CPU contention with one unidentified failing test around the 5-6% mark -- needs that failure triaged and the coverage run finished before this is trustworthy.


souliane:
Superseded by #3370 (same Django floor raise; alternate approach on an older base with failing shards). Closing as duplicate.

PR #3461, opened 2026-07-19
fix(deploy): freshbox bootstrap — all bind dirs, deterministic UID, digest-pinned base (closes #3439, closes #3438, closes #3452)

### LINKED ISSUE #3438: [deploy] Fresh-box container/host UID mismatch (1001 vs 1000) breaks every bind mount
**Category:** fresh-box bootstrap blocker (substrate-agnostic). Invisible on the current box; breaks the very-first zero-touch deploy on any fresh box.

## Problem
`deploy/Dockerfile:37` runs `useradd` on `ubuntu:24.04`, which already ships a default `ubuntu` user at **uid 1000**, so the container's `teatree` user gets **uid 1001** (verified: container uid=1001). On a fresh box the documented bootstrap `adduser teatree` (`deploy/README.md:139`) — and the default first non-root user on a fresh Ubuntu image — is **uid 1000**. Every host bind mount (`deploy/docker-compose.yml:25-63`) and the dirs `deploy/deploy.sh:57-60` pre-creates are then owned 1000 while the container writes as 1001 → `init` fails EACCES on the DB and credential dirs.

The current production box masks this only because its host user coincidentally landed on 1001.

## Suggested fix
Make the container UID deterministic and reconcilable with the host: in `deploy/Dockerfile`, `userdel -r ubuntu && useradd -u 1000 …`, or take a `TEATREE_UID` build arg passed from the host deploy-user's real uid. Document the "container UID == host deploy-user UID" invariant in `deploy/README.md`.

### LINKED ISSUE #3439: [deploy] deploy.sh pre-creates only 3 of 5 bind-source dirs — root-owned data dirs on a fresh box
**Category:** fresh-box bootstrap blocker (substrate-agnostic).

## Problem
`deploy/deploy.sh:53-60` pre-creates only the credential-plane and transcript dirs (`~/.password-store`, `~/.gnupg`, `~/.claude/projects`) owned by the deploy user, with the correct rationale in-comment: *"dockerd would otherwise create a missing source ROOT-owned, locking the user out."* But the **data-plane** bind sources — `/home/teatree/.local/share/teatree`, `/home/teatree/.local/share/teatree-worktrees`, `/home/teatree/workspace/t3-workspaces` (`deploy/docker-compose.yml:25-33`) — are **not** pre-created. On a fresh box these don't exist, so dockerd creates them **root-owned** and `init` cannot write `db.sqlite3`.

Masked on the current box only because those dirs already exist from prior runs.

## Suggested fix
Extend `deploy/deploy.sh:57-60` to `install -d` all five bind-source dirs (one added line), so the "pre-create owned by the deploy user" guarantee that already protects the credential plane also protects the data plane.

### LINKED ISSUE #3452: [deploy] Pin deploy/Dockerfile base image by digest (as dev/Dockerfile.test already does)
Part of the self-contained-deployment epic #3445 (small rider).

## Problem
`deploy/Dockerfile:5` is `FROM ubuntu:24.04` — a floating tag. The CI toolchain image it conceptually extends, `dev/Dockerfile.test:1-6`, is already **digest-pinned** to `ubuntu:24.04` precisely to keep the build reproducible and avoid silent base drift. The production deploy image should do the same.

## Why it matters
A self-contained, reproducible image (see #3445) should not depend on whatever `ubuntu:24.04` resolves to on a given build day. Pinning by digest is a one-line change that directly serves the "contains everything, builds identically" goal.

## Notes / scope
- **Stay on 24.04 for now.** The base is toolchain-only — the teatree source is not baked and Python is uv-managed at `3.13` (`.python-version`, `deploy/entrypoint.sh:347,351`), so the Ubuntu release contributes only glibc/openssl + git/gpg/pass/ttyd/jq/sqlite3 + apt plumbing. The base version barely affects behavior.
- A later bump to the newer Ubuntu LTS is mechanical (the third-party apt repos — nodesource `setup_24.x` uses `nodistro`, gh uses the literal `stable` suite, and download.docker.com already publishes for the newer codename), **but** `deploy/Dockerfile` and `dev/Dockerfile.test` must move together or production runs on a base CI never exercises. Defer that to a deliberate toolchain refresh; this ticket is just the digest pin.

Category: reproducible image / low-risk hygiene.

souliane:
## What

Fresh-box bootstrap fixes so a first deploy onto a clean host comes up writable. Three deploy-substrate issues, all under `deploy/`:

- **Closes #3439** — `deploy/deploy.sh` now pre-creates **every** host bind-mount source owned by the deploy user. It previously created only 3 of the sources; the data-plane dirs were left for dockerd to auto-create **root-owned**, which locks the non-root container out of its own writes. Enumerated from the compose file there are **six** bind sources (the issue said five — `workspace/t3-workspaces` was also missing): the DB dir, worktrees, workspaces, pass store, GPG home, and the Claude session `projects` dir.
- **Closes #3438** — `deploy/Dockerfile` pins the container user to a **deterministic UID** (`TEATREE_UID` build arg, default **1000**). Ubuntu 24.04 ships a stock `ubuntu` user already holding UID 1000, so it is removed and `teatree` is renumbered onto 1000 (`usermod`/`groupmod`), re-owning its home tree. The invariant "container UID must equal the host deploy-user UID" is documented in `deploy/README.md`.
- **Closes #3452** — `deploy/Dockerfile` base image is digest-pinned to the same `ubuntu:24.04` manifest `dev/Dockerfile.test` pins (`sha256:4fbb8e6a...2d90`), so a retag of the floating tag can never silently change the deploy toolchain.

## Migration note — BREAKING for the running deployment

**The live box currently runs the container as UID 1001**, and every host bind-mount source is currently owned by 1001. This change makes the container UID **1000**. When it lands and the image rebuilds, the pre-existing bind-mount data (owned by 1001) becomes **unwritable** by the container (now 1000), so the operator MUST, at deploy time (stack stopped), either:

- **chown the bind sources to 1000**:
  ```
  docker compose -p teatree -f deploy/docker-compose.yml down
  sudo chown -R 1000:1000 \
      ~/.local/share/teatree ~/.local/share/teatree-worktrees \
      ~/workspace/t3-workspaces ~/.password-store ~/.gnupg ~/.claude/projects
  ```
- **or** keep the box on 1001 by building with `--build-arg TEATREE_UID=1001` (matches the current host deploy user).

Do not merge-and-forget: pick one of the above as part of rolling this out.

## Verification

- Built a minimal image from the pinned base running the exact user-alignment block: result is `uid=1000(teatree) gid=1000(teatree)`, home tree + `/opt/teatree` owned `1000:1000`, stock `ubuntu` user absent.
- New smoke test `tests/test_deploy_freshbox_bootstrap.py` pins all three fixes against the deploy files (bind-source pre-creation vs the compose sources, deterministic UID renumber, digest match with `dev/Dockerfile.test`).
- Local gates green: `prek`, `ruff check`/`format`, `t3 tool test-path-mirror`, the deploy tests, and the full pre-push CI-critical parity gate.

Substrate change — held for owner authorization; do not merge without the UID migration decided.

### DIFF
--- a/deploy/Dockerfile
+++ b/deploy/Dockerfile
@@ -2,7 +2,10 @@
 # selected by $TEATREE_ROLE. Extends the dev/Dockerfile.test toolchain and adds
 # gh. The teatree source is NOT baked: it is cloned onto a volume at runtime and
 # installed editable by the init role, so the image is just the toolchain.
-FROM ubuntu:24.04
+#
+# Digest-pinned to the same ubuntu:24.04 manifest dev/Dockerfile.test pins, so a
+# retag of the floating tag can never silently change the deploy toolchain.
+FROM ubuntu:24.04@sha256:4fbb8e6a8395de5a7550b33509421a2bafbc0aab6c06ba2cef9ebffbc7092d90
 
 # Toolchain layer — ca-certificates, curl, git, jq, sqlite3, direnv, node 24, the
 # claude CLI (as in dev/Dockerfile.test) plus gh, plus pass/gnupg so the box can
@@ -65,6 +68,22 @@ RUN mkdir -p \
     chown -R teatree:teatree /home/teatree /opt/teatree
 
 COPY --chmod=0755 deploy/entrypoint.sh /usr/local/bin/entrypoint.sh
+
+# Align the runtime user to a DETERMINISTIC UID (the TEATREE_UID build arg,
+# default 1000). Every host bind mount is at path identity (see
+# deploy/README.md § UID invariant), so the container UID MUST equal the host
+# deploy user's UID or every bind mount is unwritable and init crash-loops on
+# its first write. Ubuntu 24.04 ships a stock `ubuntu` user already holding UID
+# 1000, so it is removed to free the id; the teatree user created above (at the
+# next free id) is renumbered onto TEATREE_UID and its already-populated home
+# tree is re-owned to the new id. A fresh box's first `adduser` is UID 1000, so
+# the default lines up with no extra step; override with
+# `--build-arg TEATREE_UID=<n>` when the deploy user is not 1000.
+ARG TEATREE_UID=1000
+RUN userdel -r ubuntu 2>/dev/null || true; \
+    usermod -u "${TEATREE_UID}" teatree && \
+    groupmod -g "${TEATREE_UID}" teatree && \
+    chown -R teatree:teatree /home/teatree /opt/teatree  # privacy-scan:allow
 
 # The deploy-managed default ~/.claude/settings.json template (#3359). The init
 # role generates the container's real settings.json from this (plus env overrides)
--- a/deploy/README.md
+++ b/deploy/README.md
@@ -68,6 +68,34 @@ from the operator's real DB.
 `teatree_src` and `teatree_uv` stay Docker-managed named volumes for now (later
 PRs handle code-mount modes).
 
+### UID invariant — the container user must equal the host deploy user
+
+Every state, credential, and session mount in the table above is a **host bind
+mount at path identity**, so each file on the host disk carries the host deploy
+user's numeric UID while the container writes it as the image's `teatree` user.
+Those two must be the **same UID**, or every bind mount is unwritable from inside
+the container and `init` crash-loops on its first write.
+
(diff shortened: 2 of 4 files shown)


souliane:
Superseded by #3470 (UID host-derived + 1001-safe, rebased onto latest main).

Issue #3245, opened 2026-07-15
Bot-authored housekeeping PRs never get CI (GITHUB_TOKEN) → unmergeable backlog


souliane:
## Problem
Bot-authored housekeeping PRs (weekly `uv.lock` upgrade, docs refresh, BLUEPRINT/README sync) are opened by the workflow's `GITHUB_TOKEN`. GitHub does not trigger workflows on PRs authored by the same `GITHUB_TOKEN` (recursion guard), so:

- 0 check runs ever attach to their head SHA.
- Branch-protection required checks stay `pending` forever.
- The merge keystone correctly refuses (`required checks pending, not green`).
- These PRs pile up in the backlog and can never merge through the normal path.

Confirmed on #3171, #3218, #3196 (all `total_count: 0` check-runs). Closed as stale — this issue tracks the root fix.

## Fix (pick one)
1. Open the housekeeping PRs with a dedicated PAT (not `GITHUB_TOKEN`) so CI triggers, then they flow through the normal green→auto-merge path once same-repo auto-merge is live.
2. OR have the housekeeping workflows enable GitHub auto-merge on creation (`gh pr merge --auto`) so they merge themselves when green.

## Acceptance
- A newly created weekly bot PR gets CI runs on its head SHA.
- It merges automatically once green (same-repo trusted; see the fork-provenance auto-merge work), with no human step and no backlog accumulation.

PR #3466, opened 2026-07-19
chore(deps): weekly uv.lock upgrade

souliane:
Automated weekly lockfile refresh via `uv lock --upgrade`.\n\nDependabot's `pip` ecosystem updates `pyproject.toml` constraints but does not regenerate `uv.lock`. This PR closes that gap by re-resolving all hashes after any constraint bumps that landed during the week.\n\nOpened with a PAT so CI runs; GitHub auto-merge is enabled so it self-merges once green (#3245).


souliane:
Closing: this weekly uv.lock bump has 10 failing required checks (docs-drift, sbom, lint, test-shard 6/7, test) — the dependency upgrade breaks the tree and cannot merge as-is. The next scheduled weekly lock refresh regenerates a clean bump against current main; fixing this stale one is not worth the churn.

PR #3529, opened 2026-07-21
feat(eval): refuse to publish benchmark shards not backed by metered spend

souliane:
## Problem

The weekly benchmark runs against a subscription OAuth window. When that window
is exhausted mid-run the remaining shards still "complete": the API calls return
without doing any metered work, every scenario force-FAILs, and the shard reports
`$0.0000` for the affected model(s). The publish job then committed those all-red
zero-cost matrices into `docs/evals/benchmark/*.html` as if they were real results.

Confirmed, not hypothetical — run `29760291395`:

- `under_load-2-5`, `under_load-4-5`, `under_load-5-5` — `$0.0000` for all three models, every scenario FAIL
- `under_load-3-5` — **PARTIAL**: opus metered `$0.4746` with real pass/FAIL verdicts, while sonnet and haiku were both `$0.0000` with every scenario FAIL

The partial case is the one that matters: a filter keyed on *all* models being
zero-cost silently lets it through.

## The seam

The publish job only ever holds the rendered matrix HTML — the shard jobs upload
nothing else — so that is the only substrate a publish-time guard can read.

- `teatree.eval.matrix_html_tally` — the reader counterpart to `render_matrix_html`, parsing the per-model tally table back out (located structurally by its header row, not by position; a document with no tally table is a fail-loud `MatrixTallyError`).
- `teatree.eval.benchmark_publish_guard` — the policy. A shard is contaminated when **any single model** records graded verdicts against zero metered cost.
- `t3 eval verify-benchmark-publish <dir>` — the CI primitive, wired into `eval-weekly-reusable.yml`'s publish job **before** the dashboard dir is written, so nothing contaminated reaches the working tree.

Deliberately not in workflow YAML, and deliberately not a render-time refusal: a
render-time failure would drop the shard's artifact and let publish commit the
rest as a complete-looking dashboard, which is the exact failure mode being
prevented. One contaminated shard fails the **whole** publish, naming the shard
and its zero-cost models.

`skipped` and `errored` cells are excluded from the verdict count, matching the
matrix's own pass-rate semantics — a shard with no graded verdicts (empty, or
all-skipped) owed no spend and is not contaminated.

## Tests

Integration-leaning: every fixture is a real `render_matrix_html` artifact written
to `tmp_path`, so the guard is exercised against the exact bytes the workflow
publishes.

- all-models-zero-cost with verdicts, refused
- **partial** (one model metered, two zeroed, all with verdicts), refused
- fully metered shard, published unchanged
- shard with no scenarios, not contaminated
- all-skipped shard, not contaminated
- one bad shard among good ones, whole publish refused
- CLI exit codes through the real typer app; workflow step ordering (guard before collect before commit)

**Anti-vacuity:** weakening `contaminated_shards` to the all-zero filter turns
`test_partial_contamination_is_refused` and `test_names_every_zero_cost_model_once`
RED (2 failed, 6 passed). Removing the detection entirely turns 4 of 8 RED. Both
restore to 8 passed.

## Architecture pre-check

**1. BLUEPRINT section** — "Reusable eval-CI primitives", updated in this PR with the guard.
**2. FSM** — n/a, no `Ticket`/`Worktree` transition.
**3. Extension points** — n/a; no `OverlayBase`/scanner/Protocol surface touched. The reusable workflow is overlay-consumable, so overlay callers inherit the guard for free.
**4. Component boundaries** — parse (`matrix_html_tally`) split from policy (`benchmark_publish_guard`); CLI is a thin exit-code wrapper.
**5. Dependency direction** — `teatree.eval` to `teatree.eval` only; `uv run tach check` green.
**6. Test surface** — per bullet list above; each behaviour has a named failing assertion.
**7. Resilience** — read-only over local files; the external write it gates (the dashboard commit) is unchanged. Idempotent (pure function of the dir). Fails loud, never silent-empty: a shard with no tally table raises rather than reading as clean.
**8. Identity/normalization** — shard identity is the artifact stem minus the `eval-benchmark-` prefix; no qualifier stripping to force a match.
**9. Behavior preservation** — purely additive; no matcher narrowed, no test inverted. `render_matrix_html` is untouched (the trailing-newline change from the shard-subset PR is preserved; this branch is rebased onto it).
**10. Removability / harness-vs-data** — one workflow step plus two self-contained modules; deleting them reverts to the prior behaviour with no cascade. Harness side is correct: the rule is a fixed invariant of the publish path, not per-shard policy data.

## Open questions and assumptions

- **Assumption:** a genuine metered 3-scenario shard never costs under `$0.0001`. The renderer formats cost as `$%.4f`, so anything below that is indistinguishable from `$0.0000` in the artifact and is treated as unmetered (`RENDERED_COST_RESOLUTION_USD`).
- **Assumption:** `errored` cells are not graded verdicts, so an all-errored zero-cost shard is not flagged. This follows the matrix's own pass-rate semantics; say so if the desired policy is stricter.


souliane:
Codex review pass (automated).

**Reviewed:** `benchmark_publish_guard.py` + `matrix_html_tally.py` (new) + CLI wrapper + `eval-weekly-reusable.yml` wiring + tests.

No correctness bugs found. Notes:
- Partial-contamination detection (per-model, not all-model) correctly matches the confirmed failure mode from run `29760291395`.
- Guard runs before the dashboard copy/commit steps (pinned by `TestPublishGuard.test_guard_runs_before_the_dashboard_is_written_or_committed`).
- Parser is structural (locates the tally table by header row) and fails loud (`MatrixTallyError`) rather than silently returning empty — good.
- `errored`/`skipped` correctly excluded from the verdict count, matching the matrix's own pass-rate semantics.
- Anti-vacuity check documented and verified in the PR description.

CI: `test-shard (3.13, 2)` and the downstream `test (3.13)` gate failed on the known intermittent `PytestUnraisableExceptionWarning` sqlite-leak flake (unrelated to this diff — landed on `test_retro_gate_failures.py`, a random victim in the xdist worker). Re-ran the failed jobs.

souliane:
Re-verified after the latest merge from `main` (43bbefd2): that merge touched only unrelated files, none of the guard/tally/CLI/workflow files reviewed earlier. Prior codex review (no correctness bugs found) still stands. All CI checks green, PR is CLEAN/MERGEABLE — ready to merge.

Issue #3535, opened 2026-07-21
main is RED: leaked sqlite connection fails a random shard victim (`PytestUnraisableExceptionWarning`)


souliane:
`main` is currently failing CI, and so is every PR branched from it. This is not branch-specific.

## Signature

A shard fails with an unraisable `ResourceWarning` for a leaked sqlite connection:

```
E   ResourceWarning: unclosed database in <sqlite3.Connection object at 0x...>
E   pytest.PytestUnraisableExceptionWarning: Exception ignored in: <sqlite3.Connection object at 0x...>
```

`pyproject.toml` sets `filterwarnings = ["error"]`, so the warning is a hard failure.

## Why it looks nondeterministic (it isn't)

The connection is leaked by one test but finalised by the garbage collector during a *later* test, so **the test that fails is an innocent bystander** and the name changes every run. Observed victims so far, all the same root cause:

| Run | Shard | Victim blamed |
|---|---|---|
| main `9ce01693` | 9 | `tests/teatree_core/management_commands/test_run.py::TestRunTests::test_calls_overlay_test_command` |
| PR #3531 | 10 | `tests/test_cli_tools.py::TestRepoModeCommand::test_plain_output_is_the_bare_mode` |
| PR #3531 (rerun) | 10 | `tests/teatree_core/management_commands/test_tool.py::TestToolList::test_no_tools` |

The *shard* is stable per branch (it contains the polluter); the *victim* is not. Which shard is affected depends on `pytest-split` packing, so adding or removing any test can move it — which makes it look like an unrelated PR "caused" it.

## Reproduction notes

- Reproduces only in the CI invocation. Running `--splits 12 --group 10` locally **passes** (2462 passed).
- The differentiator appears to be `--cov --cov-branch`: coverage tracing holds references and changes when GC runs, which decides whether the finaliser lands inside a test's window.
- `dev/ci-parity.sh` cannot catch this — it runs one process with a different flag set, so a locally-green suite is genuinely compatible with a red shard.

## Lead on the polluter

Main's traceback shows the finaliser firing at `click/core.py:286`, i.e. during a `CliRunner` invocation. That points at a CLI-command test whose command opens a sqlite connection (the config/cold-DB path) without closing it.

[...]

PR #3536, opened 2026-07-21
fix(db): close every worker thread's raw DB handle, not just the scanner pool

souliane:
## Why

`main` is RED on a non-deterministic failure that blocks every open PR:

```
pytest.PytestUnraisableExceptionWarning: Exception ignored in: <sqlite3.Connection object ...>
```

It lands on a different, unrelated test each time — `test_run.py` on `main`'s shard 9
([run 29815292183](https://github.com/souliane/teatree/actions/runs/29815292183)),
`test_retro_gate_failures.py` on [#3529](https://github.com/souliane/teatree/pull/3529)'s shard 2.

`pyproject.toml:326-331` sets `filterwarnings = ["error", ...]`, so the
`ResourceWarning: unclosed database` that CPython emits when it finalizes a stranded
`sqlite3.Connection` becomes a hard failure — attributed to whichever test happened to be
running in that xdist worker when the GC fired. **The reported test is always a victim,
never the culprit.** That is the whole reason it looks random.

## What leaks

Any thread that touches the ORM and exits without closing its thread-local Django
connection.

The subtlety: **`connections.close_all()` does not fix this.** Django's sqlite backend
makes `close()` a deliberate no-op for an in-memory database, because closing the last
handle would discard the database — `django/db/backends/sqlite3/base.py:221-227`:

```python
def close(self):
    self.validate_thread_sharing()
    # If database is in memory, closing the connection destroys the
    # database. To prevent accidental data loss, ignore close requests on
    # an in-memory db.
    if not self.is_in_memory_db():
        BaseDatabaseWrapper.close(self)
```

So the raw DB-API handle has to be closed directly.

## Note on `8a35ea35`

An earlier fix for this (`8a35ea35`, "close scanner pool workers' DB connections") is
**already on `main`** — it rode in with the `#3520` squash, which is `9ce01693`, the very
commit that went red. It was correct but incomplete: it fixed the scanner pool inline and
left four other worker sites leaking. This PR generalises it. Nothing here re-lands it.

Its commit message cited `src/teatree/loops/worker.py` as the "precedent" for per-thread
hygiene. That site was not a precedent — it called `connections.close_all()`, i.e. exactly
the ineffective form above, so it was itself a leak site.

## What changed

New shared primitive `src/teatree/utils/thread_db.py` —
`close_thread_db_connections()` closes the raw DB-API handle and dereferences the wrapper
so Django reopens cleanly. It refuses to run on the main thread, so it can never close the
connection a `TestCase` wraps its transaction in.

Wired into every worker site:

| Site | Before |
|---|---|
| `loop/phases/scan.py` | inline duplicate from `8a35ea35` → replaced by the shared helper |
| `loops/worker.py` | `connections.close_all()` — ineffective, real leak |
| `core/worktree/readiness.py` (`run_probes`) | no hygiene at all |
| `_workspace/provision_parallel.py` | no hygiene at all |
| `core/provision/provision_timebox.py` (callable worker thread) | no hygiene at all |

`core/provision/step_runner.py:300-322` is already covered — it hoists the ORM read onto
the caller thread and documents this exact ResourceWarning. Left alone.

No `filterwarnings` ignore, no teardown hack in the victim tests, no skip/xfail.

## Tests

A regression guard per site asserts the worker's raw handle is closed, plus
`tests/teatree_utils/test_thread_db.py` for the primitive itself — including a pin that
`close_all()` really does leave an in-memory handle open, which is direct proof that
`loops/worker.py` was leaking.

**Anti-vacuity.** Neutering the helper with an early `return` (removing hygiene at all five
sites at once) turns every guard RED:

```
FAILED tests/teatree_loop/phases/test_scan.py::TestScanPhaseConnectionHygiene::... - DID NOT RAISE
FAILED tests/teatree_core/worktree/test_readiness.py::TestRunProbesConnectionHygiene::... - DID NOT RAISE
FAILED .../test_workspace_provision_parallel.py::TestProvisionPoolConnectionHygiene::... - DID NOT RAISE
FAILED tests/teatree_loops/test_worker.py::TestExecutorThreadConnectionHygiene::... - DID NOT RAISE
FAILED tests/teatree_utils/test_thread_db.py::TestCloseThreadDbConnections::... (x2)
FAILED tests/teatree_loop/phases/test_scan.py::test_scan_phase_worker_pool_is_bounded
       - pytest.PytestUnraisableExceptionWarning: Exception ignored in: <sqlite3.Con...
```

That last line is the production bug itself, reproduced locally on an innocent bystander
test — which is the strongest evidence that this is the right root cause.

**Repeat runs.** Because the failure is nondeterministic, the guard set plus both victim
files (`test_run.py`, `test_retro_gate_failures.py`) were run under xdist **6 consecutive
times**, 92 passed each time. Also green: `ruff check`, `ruff format --check`,
`prek run ty-check --all-files`, `t3 tool test-path-mirror`, and the full pre-commit and
pre-push gate suites.

`dev/ci-parity.sh` was not run locally — CI is the check for that.

## Follow-up (not fixed here)

`backends/slack/receiver.py:207` spawns long-lived listener threads with no connection
hygiene. Production-only, not exercised under a `TestCase`, so it cannot be causing this
red — but it is the next place to look if the flake recurs.


souliane:
Two observations from chasing the same `ResourceWarning` from a different branch — offered in case they help, not as a review.

**1. The bisect is a single commit.** On `main`, the last green run is `1be4134c` and the first red is `9ce01693`, and there is exactly one commit between them:

```
$ git log --oneline 1be4134c..9ce01693
9ce01693 refactor(core): merge availability + loop presets into one Mode concept (#3520)
```

So the whole leak class arrived with #3520, which matches this PR's framing (worker threads opening Django connections). Useful as a scope bound: whatever handle is still unclosed should also be reachable from that commit's thread additions, not from older code.

**2. This branch is still red in CI at its current head.** `1b54514d` fails `test-shard (3.13, 9)` with the same signature:

```
E   ResourceWarning: unclosed database in <sqlite3.Connection object at 0x7fbd8b66c8b0>
FAILED tests/teatree_core/management_commands/test_pr.py::TestPrPostTestPlan::test_without_code_host_returns_error
  - pytest.PytestUnraisableExceptionWarning
```

Flagging it because the branch notes record 10/10 green repeat runs — which I believe, and which is compatible with CI still failing. The differentiator is `--cov --cov-branch`: coverage tracing holds references and shifts when the collector runs, so a locally-green loop and a red shard are not contradictory. A clean local shard 10 passed for me in ~5m40s and only failed once instrumentation slowed it to ~12m. So local repeat-runs unfortunately cannot confirm this one is closed; the shard job is the only authority.

The failing test is a bystander (it uses only the ORM and `call_command`) — the victim name changes every run while the shard stays stable, so it should not be read as pointing at `test_pr.py`.

**Suggestion that would shorten the remaining search:** set `PYTHONTRACEMALLOC` on the shard job. The failure output itself advises it, and with it the `ResourceWarning` carries the traceback of the line that opened the connection — naming the remaining handle directly instead of by elimination.

One trap if you are instrumenting: `sqlite3.connect` and `sqlite3.dbapi2.connect` are the same function under two names, and Django's backend calls the second (`django/db/backends/sqlite3/base.py:205`). Patching only the first reports zero connections while one is actively leaking — it cost me a full run. More detail in #3535.

Also noting `src/teatree/core/provision/step_runner.py:320` already hoists its ORM read to the caller thread with a comment describing this exact failure mode; it is not in this diff and appears to need no change — mentioning only so it is not re-investigated.

souliane:
**Codex review — verdict: merge-safe**

Reviewed `5ba3ff83` (all 6 commits) against `main`.

**Root cause & fix.** `close_thread_db_connections()` (new `utils/thread_db.py`) closes the raw DB-API handle directly instead of `connection.close()`/`connections.close_all()`, both of which are documented no-ops on Django's in-memory sqlite backend. Refuses to run on the main thread, so it can't tear down a `TestCase`'s wrapping transaction. Correctly generalizes the incomplete `8a35ea35` fix to 10 worker sites total (headless heartbeat/watchdog, provision pool/timebox/step-runner, speak, worktree readiness probes, eval parallel runner, scan phase, loop worker) — checked each site's diff; every one wraps the ORM-touching call in `try/finally: close_thread_db_connections()`, and the `eval/parallel.py` refactor (`try/except/else/finally`) preserves the original success-path semantics while adding the `finally` cleanup.

**Tests.** `test_thread_db.py` is genuinely load-bearing, not just plumbing: it pins that `connections.close_all()` leaves the in-memory handle open (the exact behavior this PR works around) and separately proves the helper is a no-op on the main thread. Per-site regression guards assert each worker's raw handle closes.

**CI.** All 12 `test-shard` + 4 `test-shuffle` jobs green on this run (`29829050143`) — the multi-shard/shuffle spread is exactly what would have caught the original nondeterministic flake. Local `t3 tool verify-gates` (commit + push stage) also green, exit 0.

No concerns. Approving for merge.

PR #3537, opened 2026-07-21
fix(setup): install git hooks in every checkout, and check it

souliane:
fix(setup): install git hooks in every checkout, and check it

`t3 setup` installed the prek hooks into the clone it was invoked from and
nowhere else. On a containerized deploy that clone is the container's, while
every commit and push comes from the host checkout — whose `.git/hooks` held
only `*.sample` files and whose `core.hooksPath` was unset. Every worktree
sharing that git dir was ungated too, so the leak gate, the banned-terms gate
and `dev/push-gate.sh` never fired on a host push. Nothing errored; CI's
tree-wide jobs still passed, so the gap was invisible.

`teatree.core.git_checkouts.discover_checkouts` enumerates every checkout
teatree commits from — the installed clone, each tracked worktree's on-disk
path, every checkout under the auto-isolated worktrees root, and the owning
clone behind each. `git_hooks_preflight.probe_checkouts` returns one verdict
per git hooks dir. Setup and the new doctor check consume the same discovery,
so the repair and the verification cannot disagree; the doctor fails per
unprotected checkout, naming the path, the missing hooks, the gates each
carries, and the remediation. An explicitly set `core.hooksPath` is reported
and left untouched.

Install is delegated to the existing `prek_hook.install`, which is idempotent
by overwrite. Same install-completeness class as the PAT-scope gap in #3523.


souliane:
Codex review — `t3 tool verify-gates` (commit + push stage): **exit 0**, all green.

**Design/code:** `git_checkouts.discover_checkouts` + `git_hooks_preflight.probe_checkouts` are well-scoped (crash-proof per-source, dedup by resolved path, worktrees collapse onto their owning clone's git-common-dir verdict) and the installer/doctor check consume the same discovery so repair and verification can't disagree. `core.hooksPath` override handling is correct (reported, never stomped). Test coverage is thorough — real-git integration tests for discovery, probing, install, and the doctor aggregation.

**Found (now fixed):** the new hard-FAIL gate ran unmocked inside `tests/cli_doctor/test_doctor_check_command.py`'s aggregation tests, which probes the *real* checkout (`/app` in CI, samples-only `.git/hooks`) and flipped `t3 doctor check`'s exit code — turned 7 pre-existing tests red across every shard/shuffle job. Confirmed by diffing CI logs against the pinned `_isolate_environment_dependent_gates` fixture. This was independently fixed in `f9e886e5` (pinning the gate alongside the other environment-dependent ones) while I was validating the same root cause locally — no further action needed there.

**Remaining CI red:** `test-shard (3.13, 12)` fails on `test_initial_migration_seed.py::FreshMigrateSeedsDefaultLoops::test_seeds_every_default_loop_once` with a `sqlite3.Connection` unraisable-exception warning — this matches the known intermittent leaked-scanner-pool-connection flake unrelated to this PR's diff (the named test is a victim, not the cause). Worth a CI re-run; not a blocker on the substance of this change.

souliane:
Codex review (follow-up) — re-review after CI red on `f1a4c090`.

**Found + fixed:** `test-shard (3.13, 2)` failed real: `TestSetupBootstrapsDjangoBeforeDmProvisioning::test_run_does_not_raise_improperly_configured` — the new `GitHooksInstaller(repo).install(...)` call in `t3 setup`'s `run()` executes *before* `ensure_django()`. `prek_hook.install()` routes through `run_step()`'s optional time-box, which imports `provision_timebox` -> `notify` -> `teatree.core.models` at call time; on a fresh pre-bootstrap shell that raises `ImproperlyConfigured` (the exact invariant this pre-existing test pins). Not the sqlite flake — a genuine ordering bug introduced by this PR's diff.

**Fix (`f1a4c090`):** hoist the existing (idempotent) `ensure_django()` call to run immediately before `GitHooksInstaller(repo).install(...)`, so Django is configured before any transitive ORM-import in the git-hooks install path. Verified: `tests/teatree_cli/setup/`, `tests/teatree_core/test_prek_hook.py`, `test_step_runner.py`, `provision/test_provision_timebox.py`, `gates/test_git_checkouts.py`, `gates/test_git_hooks_preflight.py` — 197 passed. `t3 tool verify-gates` (commit + push stage): exit 0, all green.

Re-running full CI on the new head.

souliane:
Codex review (final) — full CI re-run on head `f1a4c090` is **all green** (every shard/shuffle job, no sqlite-unraisable flake this time). Local `t3 tool verify-gates` (commit + push stage): exit 0.

Design and fix from the prior two review passes hold: Django bootstrap now runs before `GitHooksInstaller.install()`, and the new gate is correctly pinned in the doctor aggregation fixture. Nothing new found on this pass.

**merge_safe.**

PR #3542, opened 2026-07-21
chore(handover): fast-push checkpoint before termination (agent-a1b97217e7f1a1c23)

souliane:
chore(handover): fast-push checkpoint before termination (agent-a1b97217e7f1a1c23)

### DIFF
--- a/src/teatree/core/availability.py
+++ b/src/teatree/core/availability.py
@@ -368,9 +368,10 @@ def write_override(
 ) -> Path:
     """Write the override atomically via ``tmp.replace``.
 
-    ``mode`` must be one of ``"present"`` / ``"away"``. ``until`` is an
-    optional aware-datetime; ``None`` means the override never expires
-    on its own (it is cleared explicitly with :func:`clear_override`).
+    ``mode`` must be one of :data:`VALID_MODES` — ``"present"`` / ``"away"`` /
+    ``"autonomous_away"``. ``until`` is an optional aware-datetime; ``None``
+    means the override never expires on its own (it is cleared explicitly with
+    :func:`clear_override`).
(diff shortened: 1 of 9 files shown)


souliane:
Checked for supersession before doing review work on this stale `fast-push checkpoint` branch (same batch as #3542-3548, see prior investigation on #3546/#3544/#3543/#3547).

Diffed each touched file's own PR-diff (from merge-base `bb14551`) against current `main`:
- `src/teatree/core/notify.py` (`_deliver_dm` — sidecar fires only after confirmed delivery, keyed on `posted_ts`, empty `initial_comment`, F4.4) — byte-identical on main, including comments.
- `src/teatree/core/provision/provision_timebox.py` (`alert_provision_user` — `OWNER_ESCALATION` audience, F4.2) — byte-identical on main.
- `src/teatree/core/speak.py` (`deliver_user_dm_sidecar`/`_maybe_post_with_audio` `initial_comment` threading, `_serial_speaker` docstring) — identical; remaining diff vs main is only an unrelated later refactor (regex helpers moved to `speak_cleaning.py`, `mode_resolution` import).
- `src/teatree/core/availability.py` (`write_override` symmetric away/autonomous_away handling) — the whole legacy `resolve_mode`/`Schedule`/`Override`/`write_override` substrate this touched has since been replaced by the unified `teatree.core.mode_resolution` (`Mode` model). The underlying bug (asymmetric handling of deferring modes on return-to-reachable) is fixed there more generally via `_drain_if_returned(before_defers: bool, ...)`, which flips on any `defers_questions` transition rather than hardcoding a mode set.

Zero unique work remains. Closing as redundant.

Issue #706, opened 2026-05-15
execute_teardown can destroy a worktree with unpushed commits (data loss)


souliane:
## Problem

`execute_teardown` removes a worktree without checking whether that worktree holds commits that are not present on any remote. If a branch was never pushed (no upstream, no matching remote ref), tearing down the worktree and deleting its branch ref destroys the only copy of the work.

## Evidence

Concrete exposure observed during a rescue task: queued teardown for tickets #637 and #643. Both worktrees held real commits with no upstream and no matching `origin` branch:

- `#637` branch `ac-teatree-637-...`: HEAD `c656fb6` (+ parent `090439b`), no upstream, not on origin.
- `#643` branch `ac-teatree-643-...`: HEAD `faebb32` (+ parent `5658e2e`), no upstream, not on origin.

While preparing to back them up, the teardown ran and **both worktree working directories were deleted** (parent dirs left empty). The commits were only recoverable because they still existed as loose objects in the shared main-clone object store, and could be force-named onto new `origin` branches by SHA. Had the objects been GC'd, the work would have been permanently lost.

## Proposed direction

`execute_teardown` should refuse to remove a worktree (or require an explicit `--force`) when the target worktree has commits absent from all remotes — e.g. for the worktree's branch, check `git rev-list <branch> --not --remotes` is empty before destroying it. Surface a clear actionable message ("branch X has N unpushed commits; push or pass --force"). This is the priority follow-up: it is a data-loss bug, not a usability nit.

PR #3802, opened 2026-07-28
fix(retention): prune limit-park rows on their own criteria; narrow dead-row release

souliane:
## What

Two related fixes for cleanup paths whose prescribed remedy could not act on what it flagged.

### 1. Retention could not prune the rows that caused the bloat

`t3 doctor` alarms on `TaskAttempt` at 339,944 rows (ceiling 100,000) and prescribes
`t3 <overlay> retention prune`. On the live control DB that command reports:

```
{"applied": false, "total_rows": 0, "tables": [{"table": "TaskAttempt", ... "rows": 0 ...}]}
```

Zero, against a table that is 97.3% park-audit rows. The cause is structural, not a
tuning miss: `prunable()` requires the owning task to be terminal, but a limit-park
*returns* its task to the queue PENDING (`usage_window._record_park` → `Task.park`).
A park row's task is therefore non-terminal by construction, so the terminal-owned
guard excludes every one of the 330,826 park rows — permanently, at any window.

Retention now runs a second lane over `TaskAttempt` with the park's own criteria.
The 30-day terminal-owned rule is untouched and still protects real attempts.

**What identifies a prunable park row.** The `limit_parked:` prefix — the one
canonical marker, written by the single `_record_park` chokepoint, and already the
definition `TaskAttemptQuerySet._coalesce_repeated_park`, `_stamp_repair_loop_fields`,
`_check_park_spin` and `_junk_count` all share. Adding a second, differently-shaped
definition of "is a park" is exactly the drift to avoid. It is an exact prefix, not a
shape heuristic: `stuck_loop: lease lost … re-claimed by another worker` (319 rows) is
a genuine watchdog record of the duplicate-dispatch breach and is deliberately left
alone, as are the 7,300 pre-autorecovery `all configured … accounts are exhausted`
terminal failures.

**It never deletes a row carrying billed telemetry.** Measured on the live data, all
telemetry-carrying rows are `outcome=success` and zero park rows carry a token or
cost figure — so the marker alone already implies it. The guard is explicit anyway,
because those rows are the entire cost ledger every `t3 cost` / baseline / S-signal
reads, and no future writer or marker-string collision may put them in reach.

**It has its own window, measured on `ended_at`.** Default 7 days
(`park_attempt_retention_days`), separate from the 30-day rule. Age is the *last*
observation, not the first: #3680 folds a repeated park into ONE row whose `started_at`
stays ancient while `ended_at` refreshes each poll, so a `started_at` window would delete
exactly the row that says "still parked, N polls later" — the live signal the coalescer
and the 24h park-spin detector both read.

**It deletes in separately-committed batches.** A single `DELETE` over 330k rows holds
the SQLite write lock for its whole duration and can collide with a converging deploy's
migration step. Park rows are independently disposable and nothing references a
`TaskAttempt`, so partial progress is a correct outcome — unlike the terminal-owned lane,
whose all-or-nothing transaction is left as it was.

### 2. Releasing a dead worktree row required the whole clean-all blast radius

`t3 doctor` FAILs on 10 registered worktrees whose directory is not a git checkout and
prescribes `workspace clean-all` — a sweep that also prunes branches, drops databases,
reaps docker projects and removes directories. On a host running a live stack that is a
reason not to run the remedy at all.

`workspace release-dead-rows` is the narrow alternative. It applies the same protection
gates and the same classifier the sweep applies, and then deletes the `Worktree` row
alone. No directory removed, no branch deleted, no container, image, database or stash
touched. Rows it cannot positively clear are kept with their reason printed.

**"Narrow" describes the blast radius, never the safety bar** — and the first cut of this
command got that wrong. It called `classify_broken_checkout` directly, while its sibling
`reap_done_worktree` clears three gates *first*: `clean_ignore`, the ownership exclusion,
and the liveness guard. None of those is mooted by the directory dying —
`_db_liveness_reason` reads only the database, and a `clean_ignore` entry or a
`reaper_pinned` pin is the operator's decision about the ROW, not about its files. So a
pinned or mid-delivery row was KEPT by `clean-all` and RELEASED by
`release-dead-rows --apply`: two commands, two standards, and the destructive one was the
weaker.

The gates now live in one predicate, `core/cleanup/reap_pre_gates.reap_pre_gate`, which
both passes call. Re-deriving them per caller is what let the standards diverge in the
first place, so the fix is a single definition rather than a second copy: `worktree_done`
renders its verdict as SKIPPED / EXCLUDED / ACTIVE, the narrow release as a KEPT row with
disposition `protected`, and neither decides anything itself. The gate takes precedence
over the classifier's own verdict, matching the order the sweep already applies — it is
why the row is kept, so it is what the operator reads. Scope is unchanged: a row with a
live checkout is still not this pass's business and is not reported.

### 3. A stale-tracking-ref hole on that same classifier

`commits_absent_from_all_remotes` is a purely local graph query over `refs/remotes/*`. A
tracking ref goes stale the moment a branch is deleted upstream by anything other than
this clone — the ordinary forge auto-delete-on-merge — and against a stale ref it reports
commits that exist on NO remote as safely pushed. That is the misread that authorises
reaping the last copy of unmerged work.

`prune_branches` and `reap_orphan_raw_worktrees` already refresh first and fail closed.
The dead-checkout **row** path did not — and that is the one path this doctor finding
routes through. It now refreshes (memoised per clone, so ten rows cost one fetch) and
keeps the row when the refresh fails. A branch whose ref is already gone has no commits
to lose, so it is decided before the refresh and stays releasable offline.

## Tests

Proven RED before the fix, including two anti-vacuous controls:

- `test_terminal_owned_lane_cannot_reach_a_live_task_park_row` — passes before *and*
  after, establishing that the defect is structural rather than a window setting.
- `test_control_the_unrefreshed_probe_really_does_misread_the_stale_ref` — asserts the
  unrefreshed probe genuinely returns `[]` for a commit on no remote, so the freshness
  test cannot pass for the wrong reason. (This control also caught a bug in the test
  itself: a same-clone `push --delete` prunes the tracking ref, destroying the staleness
  under test — the delete has to happen on the bare remote.)

Guard tests each go RED if their guard is dropped: no park marker, carries cost, carries
output tokens, carries cache tokens, within the window, and the `ended_at`-vs-`started_at`
window semantics. `release-dead-rows` pins its narrowness directly — directory survives,
branch survives, live rows untouched, preview matches the live run.

`ProtectionGatesMatchTheSweepTest` pins each pre-gate in **both** directions, so no test
passes by the pass simply keeping everything:

| gate | protected → KEPT | absent → still released |
| --- | --- | --- |
| liveness (busy ticket) | live `Session` on the ticket | `ended_at` set |
| liveness (`reaper_pinned`) | `extra["reaper_pinned"] = True` | `False` |
| `clean_ignore` | glob matches the branch | glob does not match |
| ownership | colleague-authored tip on a product repo | owner-authored tip, same repo |

All six protected-direction assertions were observed RED against the pre-fix code
(4 × `assert not True`, 1 × `AttributeError: DeadRowDisposition has no attribute
PROTECTED`, 1 × a `releasable` verdict for a busy row); the four released-direction ones
were green before and stay green after, which is what makes the pair falsifiable.

## Note on scope

`workspace stamp-identity` and `workspace doctor` move their bodies into `_workspace`
siblings. This is the module's established pattern (see `clean_all.py`'s own docstring)
and was needed to keep `workspace.py` under the module-health LOC cap once the new
subcommand landed. Both are distinct concerns from worktree lifecycle.

## Rebase onto `main` — the conflict was semantic

The sole conflicting file was `src/teatree/core/management/commands/retention.py`, and
#3870 (`b602f486`) had added VACUUM wiring to the exact command body this branch rewrites,
so the branch diff showed that wiring as *removals*. Taking either side whole would have
deleted a merged fix. The resolution is a union: `vacuum_control_db()`, `_NOT_ATTEMPTED`,
`_VacuumRow` and the `vacuum` key on `RetentionReport` are kept from `main`; `PARK_TABLE`
and the `batches` field on `_TableRow` come from the branch; `_render` now emits the park
lane's own detail line *and* the VACUUM row. `git diff origin/main...HEAD` on that file
shows no `-` line containing `vacuum`.

`defaults.toml` needed only placement: `park_attempt_retention_days = 7` lands in
`[teatree.Infrastructure."Retention & sweeps"]` beside its siblings, under the nested
form #3866 introduced.

## Measured against the live control DB (2026-07-29)

Read-only counts, taken after the manual `TicketTransition` prune and VACUUM, so they
supersede the as-of-open figures:

| set | rows |
| --- | --- |
| `TaskAttempt` total | 340,052 |
| carrying the `limit_parked: ` marker | 330,827 |
| ...of those, carrying any token or cost figure | 0 |
| **reachable by the park lane at the default 7-day window** | **219,233** |

The terminal-owned lane still reaches 0 of them, which is the defect this PR is about.

**On the residue, correcting the earlier claim in this description.** The previous text
attributed the ~111k rows the 7-day window does not yet reach to an ongoing park storm at
"~16k rows/day". That is wrong. Park rows by `started_at`:

```
07-16  25,563     07-20 133,191     07-23  63,784
07-17  19,559     07-22  47,809     07-28       1
07-19  40,920
```

The storm ran in bursts and **stopped on 2026-07-23** — one park row since, while the loop
is otherwise alive. The residue is exactly 07-22 + 07-23 + 1, and it ages past the 7-day
window on 07-30 and 07-31, taking `TaskAttempt` to ~9,200 rows. So retention clears the
ceiling unaided within two days, and there is no upstream spin to chase.

**On disk.** The earlier text also said "no `VACUUM` exists anywhere in `src/`". True when
written, false since #3870 added `src/teatree/utils/django_db/vacuum.py` and made
`retention prune --apply` vacuum after the prune commits — which is precisely what made
this rebase semantic rather than mechanical. The two halves now compose: this lane frees
the pages, #3870's VACUUM hands them back.

The file is no longer the 1.12 GB the original text cited: the manual `TicketTransition`
prune and VACUUM took it to ~377 MB, and a zero freelist since means every page in it is
live data. It is also climbing again — 381 MiB at 09:20 UTC, 430 MiB at 10:40 — and the
climb is not `TaskAttempt`, which is flat at 92 MiB. It is
`django_tasks_database_dbtaskresult`: 224 MiB of table at 09:20, 252 MiB at 10:40, plus
~53 MiB of its own indexes. `DBTaskResult` has no retention lane at all, so it is the next
durable lever — a separate PR, not this one.

## Architecture pre-check

1. **BLUEPRINT §** — §4 (domain models / retention) and the worktree-cleanup passes. No
   contradiction with either; no BLUEPRINT edit needed.
2. **FSM phase boundaries** — none crossed. `release-dead-rows` deletes a `Worktree` row
   outright rather than transitioning it, because its checkout is provably gone and no
   teardown step can run against it.
3. **Extension-point contracts** — none. No `OverlayBase` hook, scanner registration or
   `*Backend` Protocol changes.
4. **Component boundaries** — `reap_pre_gates` sits in `core/cleanup/` beside the three
   guards it composes, not in `core/worktree/`, so the guards keep one home and both
   worktree-side callers reach them the same way.
5. **Dependency direction** — one new intra-`teatree.core` edge (`core.cleanup` →
   `core.worktree.clone_paths`) that `worktree_done` already had; `uv run tach check`
   passes as part of the pre-commit run.
6. **Test surface** — `tests/teatree_core/worktree/test_dead_row_release.py`
   (`ProtectionGatesMatchTheSweepTest`, both directions per gate) and the existing
   `tests/teatree_core/test_worktree_done.py` gate tests, whose patch targets moved with
   the symbols and are pinned by `tests/quality/test_patch_targets_resolve.py`.
7. **Resilience invariants** — the only external writes are row deletes: idempotent
   (a deleted row is simply not re-reported), preview-first, and batched with separate
   commits on the retention lane so partial progress is correct.
8. **Identity / key normalization** — n/a.
9. **Behavior preservation** — `worktree_done`'s three gate outcomes keep their exact
   labels, actions and emit records; the refactor moves the predicate, not the behaviour.
   No matcher narrowed, no must-block test inverted.
10. **Removability / harness-vs-data** — `reap_pre_gates` is a pure predicate any caller
    can stop consulting without a cascade; the per-branch and per-row policy it reads
    (`clean_ignore`, `colleague_repo_url_pattern`, `reaper_pinned`) stays in data, with
    the harness only evaluating it.


souliane:
## Cold review — verdict: `hold` at `0762cd3c`

CI green (40 SUCCESS / 3 SKIPPED / 0 fail). Merges into current `main` (`19858958`) with zero conflicts. `tomllib` parse-equality on the merged tree: 503 keys vs main's 502, the only delta is the new `park_attempt_retention_days` landing in the correct #3866 group, **zero value drift**. `makemigrations --check` clean on the merge result; `tests/config` + `tests/teatree_config` = 1960 passed. No root-level artifacts, no weakened gates.

Prune logic is correct and the row measurements reproduce **to the row**: total 340,029; park marker 330,827; park rows carrying telemetry **0**; park lane reachable at 7 d **219,233**; billed-telemetry rows 1,256 (grown from 1,252, still protected); `stuck_loop:` 329, untouched.

Anti-vacuity verified with three targeted mutations rather than a blanket revert — each guard is genuinely load-bearing:

| mutation | result |
|---|---|
| drop the stale-ref freshness guard (`broken_checkout.py`) | 4 RED / 23 pass, incl. `test_a_stale_tracking_ref_does_not_authorise_a_release` |
| drop the billed-telemetry filter from `prunable_parks` | 4 RED / 23 pass |
| swap the `ended_at` window for `started_at` | 1 RED / 26 pass — exactly the coalescer-row test |

## Blocking: `release-dead-rows` applies a weaker standard than `clean-all`

`src/teatree/core/worktree/dead_row_release.py:66` — `plan_dead_row_release` iterates `Worktree.objects` and calls `classify_broken_checkout` **directly**. The classifier's existing owner, `reap_done_worktree` (`worktree_done.py:407-414`), calls it only *after* three pre-gates:

1. `is_clean_ignored(branch, overlay=…)` — the operator's configured protect list
2. `is_excluded_by_ownership(...)` — a colleague's branch on a product repo
3. `worktree_liveness(...)` — the ACTIVE skip

The new command applies **none** of them, and this is not vacuous. `_db_liveness_reason` (`cleanup_liveness.py:152-166`) is purely DB-side — busy ticket (live session or active/claimed task), live external-delivery lease, recent E2E run, and the explicit **`reaper_pinned`** operator pin. None of those depend on the directory still being a checkout.

So a row whose directory got clobbered while its ticket is mid-delivery — or which the operator explicitly pinned, or listed in `clean_ignore` — is **KEPT by `clean-all` and RELEASED by `release-dead-rows --apply`**.

The PR body says "same classifier, same #706 data-loss standard". The classifier is the same; the standard around it is not. It is also untested in both directions: `test_a_live_checkout_row_is_left_alone` covers a live *checkout*, not a live *ticket*.

Blast radius is bounded — row only, git work provably absent, push-state guard intact — so this is state loss rather than source loss. But "unproven means keep" is the bar, and here liveness is not unproven, it is **unasked**. Fix is small: thread the same pre-gates, plus a test each way.

## Correction to the PR body: the residue attribution is wrong

The body attributes the ~111k residue to an ongoing park storm at ~16k/day. Park rows by `started_at`:

```
07-16  25,563
07-17  19,559
07-19  40,920
07-20 133,191
07-22  47,809
07-23  63,784
07-28       1
```

The storm ran in bursts and **stopped on 2026-07-23**. The loop is still alive (85 attempts on 07-28, latest 22:33) — it is simply not parking. The 111,594-row residue is exactly 07-22 + 07-23 + 1, and it crosses the 7-day window on **2026-07-30 and 07-31**, taking `TaskAttempt` to ~9,200 rows — well under the 100,000 ceiling.

So retention **does** clear the ceiling on its own within two days, and there is no upstream spin to name. Please correct the body; this needs no action beyond that.

## On the VACUUM split: keep it, but this PR is not the disk lever

The split is right — a whole-file exclusive lock does not belong in a lane whose design point is that it never holds one. But the accounting says this PR is ~10% of the achievable win:

| | file size |
|---|---|
| today | 1.199 GB (292,627 pages × 4096; freelist 140,757 = **48.1% free**, `auto_vacuum=0`) |
| this PR alone | 1.199 GB — **0 bytes** |
| VACUUM alone | ~622 MB |
| VACUUM + this PR | ~560 MB |

`teatree_taskattempt` is only **95.9 MB of 622 MB of live data** (282 B/row), so pruning 219,233 rows frees ~62 MB. Of ~638 MB recoverable, VACUUM contributes ~576 MB. Across 124 per-worktree copies: 137 GB → ~77 GB from the VACUUM, → ~69 GB with this PR on top.

Two notes for whoever lands the VACUUM: it writes a full copy, so it needs ~600 MB–1.2 GB of temp headroom (the box is at 8.6 GB free — fine, but not to be started casually). And the durable lever is neither PR: **`teatree_tickettransition` (2,956,578 rows, 354 MB with indexes) and `django_tasks_database_dbtaskresult` (358,691 rows, 161 MB) are 81% of the live data and have no retention lane at all** — `retention.py` covers only `TaskAttempt` and `IncomingEvent`. Without a lane there the file re-bloats and the next VACUUM is a recurring chore.

## Non-blocking

- **The command releases 0 of 10 rows on this host.** All ten return `unverifiable` — "the source clone … is unresolvable" — because `resolve_clone_path` finds no clone (the real one, `/home/teatree/teatree-deploy`, is outside the scanned roots). It fails closed, so it is safe, but `t3 doctor` now prescribes a remedy that no-ops on exactly the finding it is printed for — the same shape of defect this PR exists to fix.
- `plan_retention` can double-count a row that is both park-prunable and terminal-owned-prunable, inflating `total_rows`. Zero occurrences on live data today.
- No completeness gate on the `workspace` `DJANGO_GROUPS` catalogue (one exists for `ticket` in `tests/test_cli_overlay_groups.py`), so the group-entry omission fixed by hand in this branch can recur silently.

souliane:
## Second cold review — verdict: `hold` at `0762cd3c`

This is the **same SHA the previous cold review held**. The blocker was not addressed, and the PR has since developed a semantic merge conflict.

The park-prune logic itself is sound and independently re-verified — two fresh anti-vacuity mutations, not the maker's table: dropping the billed-telemetry filter from `prunable_parks` gives **4 RED / 42 pass**; dropping the stale-tracking-ref freshness guard from `_branch_verdict` gives **4 RED / 117 pass** (including `test_a_stale_tracking_ref_does_not_authorise_a_release`). Both guards are load-bearing. CI is green (40 pass / 3 skipping / 0 fail).

### Blocker 1 — `release-dead-rows` applies a weaker standard than `clean-all` (unchanged)

`src/teatree/core/worktree/dead_row_release.py:71` calls `classify_broken_checkout` **directly**. Its existing owner, `reap_done_worktree` (`worktree_done.py:406-414`), applies `is_clean_ignored`, then `_ownership_liveness_skip` (`is_excluded_by_ownership` + `worktree_liveness`) **first**.

`_db_liveness_reason` (`cleanup_liveness.py:152-166`) is purely DB-side — busy ticket, active delivery lease, recent E2E run, explicit `reaper_pinned` — so a dead directory does **not** moot any of it.

Net: a pinned or mid-delivery row is **KEPT by `clean-all` and RELEASED by `release-dead-rows --apply`**.

### Blocker 2 — untested in both directions

The 11 tests in `test_dead_row_release.py` cover a live **checkout** (`test_a_live_checkout_row_is_left_alone`, `test_a_live_checkout_is_not_reported_at_all`). None covers a live **ticket**, a `reaper_pinned` row, a `clean_ignore` branch, or an ownership-excluded branch on a dead directory. Those are exactly the cases blocker 1 opens.

### Major — the conflict is SEMANTIC, and a naive resolution deletes a merged fix

The sole conflicting file is `src/teatree/core/management/commands/retention.py`. #3870 (`b602f486`, merged) added VACUUM wiring to the exact command body this PR rewrites, and the branch diff shows that wiring as **removals**.

A resolver who takes "theirs" will silently delete #3870's VACUUM. The correct resolution unions both sides: keep `vacuum_control_db()` / `_NOT_ATTEMPTED` / the `vacuum` key in `RetentionReport`, **and** add `PARK_TABLE` plus the `batches` field on `_TableRow`. **Re-review is owed after the rebase.**

Good news on the other half: `defaults.toml` auto-merges correctly — 503 keys, and `park_attempt_retention_days` lands in `Infrastructure.Retention & sweeps` beside its siblings.

### Major — the PR body's residue attribution is still wrong

The body attributes the ~111k residue to an ongoing park storm at "~16k rows/day". Measured: the storm **stopped 2026-07-23** (1 row since; the loop is alive with 85 attempts on 07-28), and the residue ages past the 7-day window on 07-30/31, taking `TaskAttempt` to ~9,200 rows. So retention **does** clear the ceiling unaided and there is no upstream spin to chase. Please correct the body.

### Minor — a supporting claim is now false

The body states "no VACUUM exists anywhere in `src/`". True when written; false against current `main` since #3870 added `src/teatree/utils/django_db/vacuum.py` and made `retention prune --apply` vacuum. The split the body argues for **already landed on the other side** — which is precisely what turns the conflict semantic rather than mechanical.

### Note on classification

`src/teatree/config/{defaults.toml,schema.py,settings.py,setting_registries.py}` and `docs/blueprint/configuration.md` are substrate paths per `_SUBSTRATE_PATH_PREFIXES`, so this CLEAR will be substrate-classified regardless of the label recorded.

---

**No `MergeClear` has been issued for this PR, so the keystone cannot merge it.** That is the hold. It will stay held until the pre-gate fix, its tests in both directions, and a rebase that preserves #3870's VACUUM.

Issue #3634, opened 2026-07-22
Clean factory model: unified intake, WRITE-parallel/MERGE-serial, retire the batch label + skills consolidation


souliane:
## Goal

Consolidate teatree onto one coherent factory model: a single trust boundary, parallel implementation with serialized merges, a clean label/skill taxonomy, and no "batch" terminology. Implemented as a **single PR** (owner request), smallest-change-first internally.

This supersedes the ad-hoc label choreography and the interactive-implementation path. It is the durable design behind the recent reliability work (no-auto-merge policy, adversarial review before merge).

## 1. Factory model — intake -> dispatch -> WRITE (parallel) -> REVIEW -> MERGE (serial)

One factory, one trust boundary, two phases with different concurrency:

**Intake** — a single unified issue-intake scanner (folds the two existing scanners) with **one decision function**, evaluated top-down, first match wins:
1. `needs-triage` present -> **ignore** (maintainer hold).
2. An active ticket/claim/read-back already exists for this URL -> **ignore** (work exists).
3. Author trusted (repo owner/maintainers) -> **act immediately** — no admit label, no assignment, no grace window.
4. Author untrusted **and** owner-applied admit label present -> **act**.
5. Author untrusted, no label -> **ignore** (fail-closed).

**WRITE — parallel (N):** the factory owns implementation; N tickets implemented concurrently in isolated worktrees. The factory prioritizes non-conflicting tickets via a cheap area/path heuristic (not expensive conflict prediction). The factory is **always** free to push fixes to a PR that has problems — no label ever blocks review, fixes, or the gated merge of a trusted-author PR.

**MERGE — serial (1):** a single-flight merge queue; the next PR rebases/re-resolves against what just landed. This is the conflict-safety guarantee. No merge on any path without an independent adversarial-review verdict bound to the reviewed SHA (see section 5).

**The dial** lives in the `wip` concurrency surface as a **phase split: WRITE-WIP = N, MERGE-WIP = 1**, replacing todays single blunt concurrency knob. N is a runtime dial (current default 3), later driven by the resource-pressure loop.

## 2. Labels — retire "batch"

- `t3-auto` — owner-granted admission of an **untrusted authors** issue **or** PR into the factory. Replaces `t3-batch` (a forge-only convention with zero code references — a pure relabel).
- `needs-triage` — nobody takes this yet (unchanged).
- **Owner-filed issues are label-free** — trusted-author identity alone admits them; no assignment needed.
[...]

PR #3867, opened 2026-07-28
fix(config): record the require-label retirement and mark dead stored rows (#3862)

### LINKED ISSUE #3862: issue_implementer_require_label is set in the DB but consumed nowhere in src/ — dead config that reads as a live gate
`issue_implementer_require_label` is set to `true` in the control DB but is consumed **nowhere** in `src/`.

Found while diagnosing why the factory intaked nothing ([#3861](https://github.com/souliane/teatree/pull/3861)). It was checked as a candidate cause and turned out to be inert.

## Why it matters

A setting that exists, is set, and is named like a gate reads as a live control. An operator inspecting why intake refuses will find it set to `true` and reasonably conclude a label is required — which is false. It cost time in this investigation and would cost more in an outage, when the config surface is the first thing anyone reads.

The real admission rules live in `decide_intake` (`src/teatree/core/intake/factory_admission.py`) and are label-independent for a trusted author: rule 3 acts with no admit label, no assignment, no grace window.

## Fix

Either wire it or delete it. Deleting is preferable unless it has a genuine intended use — the admit-label mechanism already exists as `issue_implementer_label` and covers the untrusted-author case (rule 4).

If deleted, remove the DB row and the declaration together, and grep for stale references per the repo bar.

## Acceptance

- The setting is either consumed by code with a test proving it changes admission, or removed entirely with no stale references.
- A test or check that a declared setting with no consumer cannot ship, so this class does not recur.

souliane:
Closes [#3862](https://github.com/souliane/teatree/issues/3862).

## What

`issue_implementer_require_label` was set to `true` in the control DB but read nowhere in `src/`. [#3634](https://github.com/souliane/teatree/issues/3634) removed the field and its reader together and never recorded the retirement, so the row survived: the resolver drops an unknown key silently, and `config_setting list` rendered it under `Ungrouped` — a named gate sitting at `true`, indistinguishable from a live control.

- record the removal in `RETIRED_SETTINGS`, so a stored row hits the existing loud path (key + reason + `config_setting clear` remedy) instead of vanishing
- `config_setting list` marks every row no live setting declaration owns, so the class is closed rather than this one key
- promote `_CLEAR_REMEDY` to public `CLEAR_REMEDY`, shared by the warning and the marker so the two cannot drift
- the stale row on this box was cleared with the documented remedy

## Why

A setting that exists, is set, and is named like a gate reads as a live control. This one cost time diagnosing an intake stall ([#3861](https://github.com/souliane/teatree/pull/3861)): an operator reading the config surface finds `true` and concludes a label is required, which is false — `decide_intake` rule 3 admits a trusted author with no label, no assignment, and no grace window. In an outage the config surface is the first thing anyone reads, so a dead key that reads live is a live hazard.

"Not a known key" is **not** the same question as "dead". `loop_preset_transition_stamp` is also outside the key registries but is live state `teatree.loops.preset_transitions` rewrites every pass. The first cut of this marker told the operator to clear it — and clearing the stamp makes the next transition pass read a switch that never happened, posting a spurious Slack line. Hence the third bucket: `INTERNAL_STATE_KEYS`, each entry naming its owning module as a string (the config layer sits below `teatree.loops`, so it may not import it), plus a guard that the named module still carries the key so the carve-out cannot rot into the dead config it exists to prevent.

## Architecture pre-check — #3862

**1. BLUEPRINT § alignment** — § Configuration (DB-home settings partition) and `BLUEPRINT.md:282` (§ intake, "ONE intake scanner, ONE decision function"), which already records this setting as retired by #3634 in favour of the `decide_intake` table. The code half of that retirement never landed. No BLUEPRINT edit: the doc already states the retired state; this makes the code agree with it.

**2. FSM phase boundaries** — n/a, no `Ticket.State` / `Worktree.State` transition.

**3. Extension-point contracts** — none. No `OverlayBase` hook, scanner registration, or `*Backend` Protocol touched. `RETIRED_SETTINGS` already has two derived consumers wired (`config.resolution._coerce_setting_rows`, `core.config_migration`).

**4. Component boundaries** — `config/retired_settings.py` (the single registry a retirement is recorded in: one data row + the promoted constant); `config/stored_row_health.py` (new — the classifier and the internal-state registry); `core/management/commands/config_setting.py` (`list` renders the trailer, mirroring the existing `_flag_suffix` shape). Deliberately **not** folded into `known_settings.py`, which is documented as the minimal leaf that "closes the union without an import cycle" — classification policy plus a second registry is a different concern. Not `core/config_display.py` either: that owns value rendering and secret masking, whereas key liveness is config-registry knowledge.

**5. Dependency direction** — `stored_row_health` → `known_settings` + `retired_settings`, all config leaves. The internal-state registry holds its owner as a **string** so the config layer never imports `teatree.loops` (that would be a backwards edge). `uv run tach check`: all modules validated.

**6. Test surface**

| Behaviour | Test | Pre-fix |
|---|---|---|
| Key recorded as a removal, reason points at `decide_intake`, stored row warns loudly | `tests/config/test_retired_settings.py::TestIntakeRequireLabelRetirementIsRecorded` | RED (3 failures) |
| Four verdicts; no live key ever marked (totality); internal-state bucket; registry-rot guard | `tests/teatree_config/test_stored_row_health.py` | RED |
| `list` marks retired + unrecorded rows, names internal state without the destructive remedy, leaves live rows unmarked | `tests/teatree_core/management/test_config_setting_command.py::TestConfigSettingListMarksDeadRows` | RED (2 failures) |

The internal-state cases were separately observed RED against an emptied carve-out (4 failures), so they are not vacuous.

**7. Resilience invariants** — no external write; the `config_setting` change is presentation only. The stderr warning is already idempotent-per-resolution by design (emitted per resolution, never once per process, so a warm loop worker cannot lose it). The printed remedy is not a dead end: `clear` does not validate against `_ALLOWED_SETTINGS`, so it works on a key outside the known-key set.

**8. Identity and key normalization** — the config key IS the canonical identity; no bare-vs-qualified form. The classifier does no stripping or splitting — exact-match membership against `ALL_KNOWN_CONFIG_SETTINGS`, the single union every surface already resolves key-ness through.

**9. Behavior preservation / capability deletion** — additive at the code level. `_coerce_setting_rows` keeps all four behaviours (renamed folds, removed warns, unknown drops, parser failure raises loud); this moves ONE key from "drop silently" to the already-existing "warn loudly" bucket, and no resolved value moves either way because the key has no field to resolve onto. `warn_removed_setting` keeps the same message and remedy — only the constant's name changed. `config_setting list` keeps grouped rendering, scope labels, and prints every row including leftovers; the trailer is appended, nothing hidden or reclassified. No gate narrowed, no privacy/leak matcher touched, no must-block test inverted. The setting itself was already deleted by #3634 — this records a deletion that already happened.

**10. Removability / harness-vs-data** — the `RetiredSetting` row is one data entry deletable in isolation; `stored_row_health` is a pure leaf with one call site, so deleting it reverts `list` to today's output (blast radius: one line in the CLI). Policy lives in two tables (`RETIRED_SETTINGS`, `INTERNAL_STATE_KEYS`); the harness is a generic classifier that knows nothing about `issue_implementer_require_label`. Recording the next retirement is a one-line data edit.

## Acceptance

- [x] *"either consumed by code … or removed entirely with no stale references"* — removed. Retirement recorded, DB row cleared, `grep` across `.py`/`.md`/`.toml`/`.yaml`/`.sh`/`.json` shows only the retirement record, the tests, and the BLUEPRINT line that already documents it as retired.
- [x] *"a test or check that a declared setting with no consumer cannot ship"* — `tests/conformance/test_user_settings_readers.py` already fails any `UserSettings` field with no real reader. This PR closes the complementary hole the bug actually fell through: a **stored row** with no live declaration.

## Verification

- `t3 tool verify-gates` → **exit 0**, all gate stages green (commit + push).
- Affected tests: 97 passed post-merge.
- Full suite: 34070 passed / 24 failed. Each failure A/B'd against a pristine `origin/main` worktree — 61 of 62 node failures identical on both; the two asymmetric ones are `subprocess.TimeoutExpired` flake in the same file on different nodes. One failure **was** mine (`test_cli_reference_drift_gate` — the edited `list_rows` docstring feeds `docs/generated/cli-reference.md`) and is fixed by regenerating the reference.

- **assumed** — delete rather than wire. The issue prefers deleting "unless it has a genuine intended use"; `decide_intake` rule 4 already covers the label case via `issue_implementer_label`, so no use remains.
- **decided during the build, not planned** — `loop_preset_transition_stamp` is a live internal-state row, so key-registry absence cannot imply dead. Found by checking the marker's output against every unknown key on this box, not by the plan.
- **assumed** — acceptance criterion 2 already ships as `tests/conformance/test_user_settings_readers.py`; no second declared-setting lane was added.
- **open, deliberately out of scope** — a `t3 doctor` check reporting orphaned stored rows on an operator's box. The `list` marker covers the surface where the harm actually occurred; a doctor probe is a second surface this issue did not ask for.
- **open, untouched** — keeping runtime state in a `ConfigSetting` row at all (the mode stamp) is arguably the wrong home; the carve-out documents it rather than moves it.


souliane:
## Cold review — verdict: `hold` at `e5b69713`

CI is green (40 pass / 3 skipping / 0 fail), the branch merges cleanly into current `main` (`19858958`), no gate is weakened, `cli-reference.md` regenerates to a zero diff, and the anti-vacuity holds — reverting each component separately turns the suite red (retirement entry → 4 failed, `config_setting.py` → 6 failed, `INTERNAL_STATE_KEYS` emptied → 5 failed).

The `loop_preset_transition_stamp` carve-out is **correct and verified by execution**: it is a real row (value `"maintenance"`), `apply_preset_transition` reads it at `preset_transitions.py:59` and writes at `:65`, and clearing it makes `prior_name=""` ≠ `current_name` so `_post_switch_line` fires. The guard also does not over-carve — classifying all 38 live keys flags exactly that one row.

## Blocking: the third key exists, and it is `approval_dial`

The carve-out is a hardcoded name list rather than a property, so nothing detects another key in the same position. Scanning `src/` for every `ConfigSetting` key outside `ALL_KNOWN_CONFIG_SETTINGS` finds **three live-consumer keys** landing in the "unknown" bucket. Probed by creating the rows and running `config_setting list`:

```
approval_dial = {'merge_own_pr': 'auto'}  [global]  [unknown — no live consumer; clear with `t3 <overlay> config_setting clear approval_dial`]
default_mode = 'engaged'                  [global]  [unknown — no live consumer; clear with `t3 <overlay> config_setting clear default_mode`]
presence_upgrade_mode = 'engaged'         [global]  [unknown — no live consumer; clear with `t3 <overlay> config_setting clear presence_upgrade_mode`]
loop_preset_transition_stamp = 'maintenance' [global] [internal state — …]
```

- **`approval_dial`** is the per-action-class approval trust table. Live consumers: `policy_dial` / `effective_decision`, wired into `core/on_behalf_gate_recorded.py:249`, `loops/directive_loop/ratify.py`, `loops/outer_loop/keep.py`. It has its own operator surface — `core/management/commands/approval_dial.py` writes the row via `ConfigSetting.objects.set_value(DIAL_CONFIG_KEY, table, scope=overlay)`. **Following the printed remedy silently un-graduates every approval class.**
- **`default_mode` / `presence_upgrade_mode`** are read by `mode_resolution._setting_name()` and `config/cold_mode.py`, and written by `loops/preset_admin.rename_preset`. Their sibling `low_power_preset_name` *is* declared (`registries.py:109`), so the omission is an inconsistency rather than a convention.

None of those rows exists on this box today, so nothing is being misinformed right now. But for `approval_dial` this is a **regression rather than an incomplete improvement**: today the row prints bare; after this PR it prints a false claim with a destructive instruction attached, on a security-relevant row. That is exactly the harm the PR's own "Why" section argues the third bucket exists to prevent — and the body's claim that "the class is closed rather than this one key" is disproved by its own carve-out.

## Remedy — small

1. Add the three keys to `INTERNAL_STATE_KEYS`; they satisfy the same "owner module carries the key" rot guard.
2. Soften the negative bucket to what the classifier can actually support: membership in `ALL_KNOWN_CONFIG_SETTINGS` answers *"is this a declared setting"*, not *"does anything read it"*. Say **`not a declared setting`**.
3. Add the **inverse** conformance test: every `ConfigSetting` key constant in `src/` is either in the union or in the carve-out. The existing rot guard is one-directional — it catches the exemption going stale, never the exemption being incomplete. This is also the honest discharge of the issue's acceptance criterion 2, since `tests/conformance/test_user_settings_readers.py` (a genuine AST reader-resolution lane, not a nominal one) covers declared-field-with-no-reader, which is a different class from an un-registered live key.

## Minor

- `test_the_retired_intake_gate_is_marked` asserts only that the note is non-empty, so it passes even when the key falls into the wrong bucket. The discriminating assertion lives in the command test; this one adds nothing.
- The new test sits in `tests/teatree_config/` while its closest sibling for the same source package is in `tests/config/`. The path-mirror gate accepts both, but note `teatree.quality.affected_tests._TESTS_CONFIG_PREFIX` escalates to the full suite only for `tests/config/`.

## Note on this being the factory's first autonomous delivery

Better than a passing first attempt. It picked the right branch of the issue, cleaned up after itself on the live box, wrote tests that genuinely fail without each component, and — most creditably — discovered the `loop_preset_transition_stamp` hazard mid-build, having not planned for it, and said so plainly under "Open questions & assumptions" rather than dressing it up as foresight. That is disclosure a lot of hand-written PRs skip.

The one step missing: on finding that *a key-registry miss does not imply dead*, generalise instead of enumerate. The insight was there; it stopped at the single instance it tripped over and then wrote "the class is closed" in the body. One `rg` for other `ConfigSetting` keys outside the union surfaces `approval_dial` in under a minute. Everything blocking here was found by an inverse search of the agent's own discovery.

Issue #3871, opened 2026-07-28
No retention lane for TicketTransition (2.96M rows) or DBTaskResult (359k) — 81% of the control DB grows unbounded and is copied per worktree


souliane:
`src/teatree/core/retention.py` has lanes for `TaskAttempt` and `IncomingEvent` only. The two tables that actually dominate the control DB have no retention at all.

## Measured

Read-only against the live control DB (`~/.local/share/teatree/db.sqlite3`):

```
teatree_tickettransition            2,956,578 rows   354 MB (with indexes)
django_tasks_database_dbtaskresult    358,691 rows   161 MB
teatree_taskattempt                   340,029 rows    95.9 MB   <- the only one with a lane
```

Live data totals ~622 MB, so **the two unmanaged tables are ~81% of it**.

## Why this outranks the prune work already in flight

The control DB is copied per worktree by `paths._seed_isolated_db` (a full `sqlite3.backup()`), and 124 such copies currently consume **137 GB** — see [#3852](https://github.com/souliane/teatree/issues/3852). So the DB's size is multiplied by every worktree ever provisioned, and anything that lets it grow unbounded is multiplied too.

Current accounting for the 1.199 GB file (292,627 pages × 4096; freelist 140,757 = 48.1% already free, `auto_vacuum=0`):

| action | resulting file |
|---|---|
| [#3802](https://github.com/souliane/teatree/pull/3802)'s park prune alone | 1.199 GB — **0 bytes** (SQLite does not return freed pages without VACUUM) |
| VACUUM alone | ~622 MB |
| VACUUM + #3802 | ~560 MB |

`teatree_taskattempt` is only 95.9 MB of the 622 MB of live data, so the prune frees ~62 MB of the ~638 MB recoverable. **Without a lane on the two big tables, the file simply re-bloats and every VACUUM becomes a recurring chore rather than a fix.**

## What to build

A retention lane for each, in the same shape as the existing ones — a conservative prunable predicate, a chunked delete, and a `--dry-run` that reports what it would remove and why.

Open questions the implementer should answer with evidence rather than assume:

- **`TicketTransition`** is an audit trail of FSM moves. What is it read for after the fact, and over what window? If nothing reads transitions older than N days, that is the window. Note the board reconciler ([#3851](https://github.com/souliane/teatree/pull/3851), merged) writes a transition per reconciled ticket on an hourly cadence, so the growth rate is now partly structural.
[...]

Issue #3856, opened 2026-07-28
Dashboard has no live-work view: nothing shows what the factory is running headless right now


souliane:
Owner asked whether the dashboard shows what the factory is running headless right now. It does not.

## Current state, measured

No dash template references `claimed`, `in_flight` or a running attempt:

```
grep -rln "claimed|in_flight|running" src/teatree/dash/templates/dash/*.html
(no matches)
```

The five pages are Board, Health, Loops, Schedule, Settings. None answers "what is executing at this moment". The board shows ticket FSM state, which is a lagging view and — until [#3851](https://github.com/souliane/teatree/pull/3851) lands — a stale one.

So the single question an operator asks most often about an autonomous system has no surface.

## What to build

A **live-work view**, reachable from the nav, showing what is executing now:

1. **Running attempts** — open `TaskAttempt` rows (`ended_at IS NULL`): task, ticket, phase, execution target, lane, elapsed, and the agent session if there is one.
2. **The queue** — `pending` and `claimed` task counts, and what is next.
3. **Loop liveness** — per loop: last run, next due, and *why* it did not run if it did not. That last field is the one that matters and it now exists: [#3845](https://github.com/souliane/teatree/pull/3845) added a per-loop refusal reason (held / forced OFF / not due / disabled / preset-masked / colleague-deferred / CAS lost / raised). Render that value rather than inventing a second vocabulary.
4. **Recent terminal outcomes** — a short tail of completions, refusals and parks, so a park storm is visible as it happens rather than by counting rows days later.

Poll it with the existing htmx + idiomorph morph-swap pattern, matching the board's cadence, so scroll position stays stable.

## Deliberately NOT a charting framework

Owner floated Panel/HoloViews. Recommendation is to keep the current stack — server-rendered Django templates with vendored htmx and no build step — and add the missing *data*, not a second web framework.

Reasons, in order of weight:

- **The dashboard's problems are not rendering problems.** `/dash/health/` takes 21.3s (N+1 queries), the board renders a lifecycle nothing advances, settings is 272 forms and 260 KB. A rewrite inherits all of it.
- **Cost.** Panel brings a Bokeh server, a websocket, and a Python process holding per-session state, plus `bokeh` / `param` / `holoviews` / `numpy`. This box hit 96% disk and load 45 today.
- **It contradicts [#3855](https://github.com/souliane/teatree/issues/3855)**, where the direction is explicitly fewer mandatory dependencies and headless-first.
[...]

Issue #3825, opened 2026-07-28
Guarantee a byte-identical defaults.toml round trip: two export filters, nested TOML, and file-based import


souliane:
Owner requirement: *"I want a round trip with ZERO discrepancies: import toml, export toml → input and output are byte identical."* The working model is: edit settings in the dashboard, export, drop the file over `config/defaults.toml`, and have it work. Nested TOML is wanted as well, and import should take a file rather than a textarea.

## Why this does not work today — by design, not by bug

`config/defaults.toml` is the shipped-default **floor**; the `ConfigSetting` store holds only **deltas** from it. `export_db_to_toml` dumps the store.

`import_toml_to_db`'s own docstring states the consequence:

> A value equal to the shipped default writes NO row (the #3676 zero-seed + `restore = delete row` property), so **a dump of `defaults.toml` imports to zero rows**.

And export emits the settings table only `if global_rows:`. So `export(import(defaults.toml))` produces **no `[teatree]` table at all** — the two are not inverses, and "export then replace the file" would drop every key that was not overridden.

The seed half is already fine: `_emit_seed_tables(document)` renders `[loops.*]` / `[modes.*]` from the live models, so those round-trip.

## P1 — Two independent export filters (owner-specified)

Two checkboxes on the settings page, mirrored as CLI flags. **Both unticked by default, which reproduces today's behaviour exactly.**

1. **Export default keys only** — restrict to the `Category.DEFAULT` key set that `[teatree]` ships. Drops registries, secrets, personal identifiers, overlay scopes and unknown keys.
2. **Include values that are the same as default** — also emit keys whose effective value equals the shipped default, i.e. keys with no DB row.

They are orthogonal: filter 1 chooses *which keys are eligible*, filter 2 chooses *divergent-only vs all*. Ticking **both** yields exactly the `defaults.toml` shape, which is what makes the round trip possible.

Guard, tightened from the owner's sketch: with both ticked, assert the emitted key set **equals** the `Category.DEFAULT` set rather than merely being a subset. A subset check still permits silently losing a key, which is the failure mode that makes "replace the file" dangerous.

## P2 — The invariant, in CI

```
export(defaults-shape) after import(defaults.toml)  ==  defaults.toml     # byte-for-byte
```

Requirements to make it hold: one deterministic key order, one formatting/quoting path, reproduced heading comments, trailing newline.

[...]

Issue #3846, opened 2026-07-28
Upgrade to Python 3.14 and the latest Django


souliane:
Owner request: upgrade to Python 3.14 and the latest Django.

## Current state, measured on this box

- Project baseline is Python **>= 3.13**; CI runs `test (3.13)` and a 12-way `test-shard (3.13, N)` matrix.
- The host already has **Python 3.14.4** as `/usr/bin/python3`, and the uv-tool venv reports `3.14.4`.
- Django in the tool venv is **6.0.7**.

So the runtime is ahead of the declared baseline, which is worth resolving deliberately rather than leaving implicit.

## Scope

1. **Python 3.14** — raise the floor in `pyproject.toml` (`requires-python`), the ruff/ty target versions, `dev/Dockerfile.test`, and every CI matrix entry (`test`, `test-shard`, `test-shuffle`, `lint`, `dash-e2e`, eval lanes). The shard matrix names `3.13` in job names; those change too.
2. **Latest Django** — confirm the current release at implementation time and move to it. Note `defaults.toml` and the docs describe a 6.x baseline with 5.2 LTS compat; decide whether that compat statement still holds and update it rather than leaving it stale.
3. **Dependency sweep** — `uv lock --upgrade`, then resolve whatever breaks. `uv-audit` and the SBOM lane must stay green.

## Known hazards on this repo

- **The whole-tree suite is ~34,000 tests and peaked 20.4 GB of 31 GB on a full run.** Do not run it locally on this box. Use scoped runs and let CI's sharded lane be the authority. A version bump touches everything, so this is exactly the change most tempting — and most dangerous — to validate locally.
- **`django-fsm`, `django-typer`, `django-tasks`, `tomlkit`, `pydantic`/`pydantic-settings`, `whitenoise`, `gunicorn`** are load-bearing. `django_fsm` in particular gates every `Ticket` transition; if it is unmaintained against the target Django, that is a blocker to surface early rather than discover at the end.
- **Migrations**: `makemigrations --check` is a CI gate. A Django upgrade can change field deconstruction and produce phantom migrations. `django-linear-migrations` enforces one leaf per app.
- **The deploy is Docker** (`deploy/docker-compose.yml`: `teatree-init`, `teatree-worker`, `teatree-admin`, `teatree-slack-listener`). The image Python must move with the baseline or the containers diverge from CI.
[...]

PR #3878, opened 2026-07-29
perf(dash): bound every dashboard read by the page, not by table size

souliane:
## What

Bounds every unbounded read on the dashboard, at the source.

- **The drawer.** `/dash/tickets/287/` answered **5,005,942 bytes**; ticket 283 answered **13,954,308 bytes / 3.53s**. Three reads had no bound: `_transitions()`, `_tasks()` (every task, every attempt prefetched) and `build_ticket_lifecycle_mermaid()` (one diagram edge per transition row). Each now keeps the newest 50 transitions / 20 tasks / 10 attempts per task — the attempt cap a SQL window, since a prefetch that fetches thousands to render ten is the same unbounded read one level down — and each panel **states the total it truncated from**. The diagram draws DISTINCT edges: a state diagram is a set, so the worst ticket's 22,965 rows are 10 arrows.
- **The spend chip.** `cost_breakdown()` instantiated 340,047 `TaskAttempt` rows (680,187 `from_db` calls with the `select_related` join) and totalled them in Python — 27.7s of a 19.2s request. Every accumulator is linear in the per-row token counts, so it is now one `GROUP BY` over the small `(model, lane, phase, estimated, reported?)` key space, behind a covering index so the scan reads an index rather than a table whose rows carry the error/result payloads. `from_usages` and `from_groups` share one accumulator, so the two spellings cannot compute different numbers.
- **Repeated reads.** One loops render resolved the effective settings 4× (36 `importlib.metadata` entry-point scans, 5,508 metadata file reads, 3.2s) and re-read the preset/schedule/loop tables 2–4× each. `RequestScopedReadCacheMiddleware` + `teatree.request_cache` collapse those to one each, memoized only inside an explicitly entered scope — the CLI and the loop tick are untouched — and dropped by ANY non-SELECT statement, so a POST that mutates then re-renders still reads its own writes (a bulk `update()` fires no model signal, which is why invalidation keys on the SQL).
- **The poll.** Polled panels carry `hx-sync="this:drop"` and board cards carry a stable `id`, so a slow render cannot start a second poll over itself or re-morph the element under the cursor.
- **Transcripts have a home.** `/dash/sessions/` (nav item *Sessions*) lists recent agent sessions, each linking to its redacted tail; the transcript page's nav lands back on it.

## Why

Neither surface had changed. `TicketTransition` reached 2.96M rows and `TaskAttempt` 340k, and code that reads a whole table crosses that threshold silently — the operator sees a 21.3s health page and a card that does nothing on click. Retention postpones this; a bound at the query is what makes it structural.

## Measured, on the real 1.2 GB control DB

| Surface | before | after |
|---|---|---|
| `/dash/health/` | 22 q, **19.204s** | 22 q, **0.366s** |
| `/dash/loops/` | 22 q, 3.892s | 21 q, 0.064s |
| `/dash/board/` | 18 q, 0.389s | 18 q, 0.193s |
| `/dash/tickets/287/` | 5,005,942 B, 0.761s | 16,322 B, 0.031s |
| `/dash/tickets/283/` | 13,954,308 B, 3.526s | 45,558 B, 0.216s |
| `/dash/presets/` | 12 q, 0.096s | 12 q, 0.036s |

Django 6.0.7 has **no** automatic prefetching — verified empirically (10 rows still cost 11 queries) and by grepping `DEFAULT_NAMES` and `related_descriptors.py`. So the owner's stated preference does not apply as a feature; the explicit tools do the work: SQL aggregation where the result is a fold, `Window`/`RowNumber` where a per-parent cap is needed, `select_related` on the session index.

## RED-then-GREEN evidence

| Guard | RED output before the fix |
|---|---|
| drawer payload | `AssertionError: drawer payload 1011905 bytes` / `assert 1011905 < 250000` |
| stated truncation | `AssertionError: transition history does not state its total` |
| per-panel caps | `assert 57 == 50`, `assert 17 == 10` (slices removed) |
| lifecycle diagram | `AssertionError: assert 500 == 1` |
| cost equivalence | `AssertionError: 80.55 != 81.63368770000001` (group key missing the reported axis) |
| memo invalidation | `assert 1 == 2` (invalidation removed) |
| secret redaction | the banned term appeared in the transcript bytes (redactor removed) |
| query pins | `18 != 11` and `23 != 9` board, `15 != 11` drawer (per-row query planted) |
| poll overlap | all four polled elements listed as able to re-enter themselves |

The first cost-equivalence corpus was **degenerate** — `estimated` was perfectly correlated with `reported`, so dropping the reported axis from the group key still passed. The corpus is now the full product of the five axes.

## Architecture pre-check

1. **BLUEPRINT §** — the `/dash/` paragraph in § "Global CLI"; two new entries added beside the #3825 settings-dashboard one.
2. **FSM phase boundaries** — n/a, every read is read-only.
3. **Extension-point contracts** — none touched; `cost_breakdown` keeps its signature and `headless_cost_breakdown` (overlay SDK) is unchanged.
4. **Component boundaries** — `teatree.request_cache` is a foundation leaf (no teatree imports); the middleware that opens its scope is in `core`; the dash read models stay in `dash`.
5. **Dependency direction** — `uv run tach check` → `✅ All modules validated!`; `teatree.request_cache` declared `depends_on = []`.
6. **Test surface** — per the table above; every page's plan pinned in `tests/teatree_dash/views/test_page_query_pins.py` at two populated sizes.
7. **Resilience invariants** — no external write. The memo is idempotent by construction and invalidated by any write; a read outside a scope is unchanged.
8. **Identity and key normalization** — the costing key is the fully-qualified `(model, lane, phase, estimated, reported?)` tuple; nothing is stripped to make a comparison succeed.
9. **Behavior preservation** — the diagram drops duplicate edges (a state diagram is a set) but keeps the session-suffix label; the drawer drops rows past the caps and SAYS so. No privacy matcher narrowed, no must-block test inverted. The settings page pin came **down** 7 → 5.
10. **Removability / harness-vs-data** — deleting the middleware line reverts the memo with no call-site edits; the caps are module constants, tuning knobs rather than contracts, so the tests assert the capping BEHAVIOUR, not the numbers.

- The covering index is ~20 MB on a 340k-row table and is not written by the park-coalescing `UPDATE` (which touches no indexed column). If the retention work in [#3871](https://github.com/souliane/teatree/issues/3871) shrinks that table permanently, the index earns less but costs little.
- The caps (50/20/10) are a judgement call about what a reader needs; they are stated in the UI so a reader who needs more knows there is more.
- Nav placement anticipates [#3856](https://github.com/souliane/teatree/issues/3856): running attempts, queue depth and per-loop refusal reasons answer the same "what is the box doing" question and belong on this same *Sessions* slot. Not implemented here.

Relates to [#3856](https://github.com/souliane/teatree/issues/3856), [#3871](https://github.com/souliane/teatree/issues/3871).


souliane:
## Correction to the "no automatic prefetching" finding — it exists, in 6.1

The PR body states Django has no automatic prefetching, verified against 6.0.7. **That verification is correct for the version we run, but the feature does exist** — it landed in **Django 6.1**:

- [Django 6.1 release notes](https://docs.djangoproject.com/en/dev/releases/6.1/)
- [Auto-prefetching with model field fetch modes in Django 6.1 — DjangoCon US 2026](https://2026.djangocon.us/talks/auto-prefetching-with-model-field-fetch-modes-in-django-6-1/)
- [django-developers: Automatic prefetching in querysets](https://groups.google.com/g/django-developers/c/EplZGj-ejvg/m/nYb0jxSTBgAJ)

`QuerySet.fetch_mode()` with **`FETCH_PEERS`** acts as an on-demand `prefetch_related()`, collapsing most N+1 cases to two queries **without maintaining a list of fields to prefetch** — which is precisely the maintenance problem in these selectors, since they are built in one module and consumed in templates elsewhere. There is also a **`RAISE`** mode that turns an un-fetched access into a `FieldFetchBlocked` exception.

We are pinned at `django>=6,<6.1` (`pyproject.toml:41`) on 6.0.7, so none of it is reachable here.

**Nothing in this PR should change.** The `Window`/`RowNumber` + SQL-aggregation approach is the correct answer at 6.0.7, and the measured results stand on their own — `/dash/health/` 19.204s → 0.366s, `/dash/tickets/283/` 13,954,308 B → 45,558 B. The `assertNumQueries` pins should stay after any future upgrade too: they assert a query *budget*, which is a different and stronger claim than "nothing was left un-fetched".

Please amend the PR body so it says **"6.0.7 has no automatic prefetching; the feature arrives in 6.1, which our pin excludes"** rather than implying Django has no such feature. The repo's own doctrine is that docs and code must describe the current state accurately, and a future reader on 6.1 would otherwise be misled by this file.

Recorded on [#3846](https://github.com/souliane/teatree/issues/3846) (py3.14 + latest Django) as an additional reason to prioritise that upgrade: `FETCH_PEERS` would let these selectors drop hand-maintained prefetch lists, and `RAISE` would convert "do not write an N+1" from an instruction into deterministic enforcement — which is what the First Principles at the top of `AGENTS.md` ask for.

Issue #3526, opened 2026-07-21
Overlay contract: conditional kwarg forwarding hides a signature mismatch until the non-default path runs


adrien-oper:
## Summary

`_validation_errors` forwards the `require_sections` keyword to an overlay's `validate_pr` **only when it deviates from the default**. The intent is compatibility with overlays that implement the older two-argument signature. The effect is the opposite of safe: an overlay that does not implement the documented signature works on the common path and raises `TypeError` on the rarely-taken one.

A permanent contract violation is turned into an intermittent runtime crash, surfacing only when someone happens to use the non-default flag.

## Detail

The base class documents the full signature:

```python
# core/overlay_metadata.py
def validate_pr(self, title: str, description: str, *, require_sections: bool = True) -> ValidationResult:
```

The caller forwards conditionally:

```python
# cli/tools.py
kwargs = {} if require_sections else {"require_sections": False}
```

An overlay implementing only `validate_pr(title, description)` therefore:

| path | `require_sections` | forwarded? | result |
|---|---|---|---|
| normal validation | `True` (default) | no | works |
| sections-optional / title-only | `False` | **yes** | `TypeError: validate_pr() got an unexpected keyword argument 'require_sections'` |

The guard is precisely inverted with respect to when it is needed: the only case it forwards is the only case a non-conforming overlay cannot handle. An overlay author gets no signal at load time, no signal in normal use, and a crash the first time anyone passes the flag.

## Why this matters beyond one keyword

This is a general hazard in the overlay extension-point contract, not a one-off. Any future optional keyword added the same way inherits the same behaviour: silently tolerated until the non-default branch runs. Contract violations that only manifest on an uncommon path are the expensive kind — they survive review and testing and fail in front of a user.

## Suggested direction

Options, roughly in order of preference:

1. **Always forward the keyword.** An overlay that does not implement the documented signature then fails immediately and visibly, on the first call rather than the unlucky one. Loud and early beats quiet and later.
2. **Verify overlay conformance once, at load/registration time** — check that the overlay's `validate_pr` accepts the documented parameters (e.g. via `inspect.signature`) and fail with a clear message naming the overlay and the missing parameter. This catches the whole class rather than this instance.
3. If backwards compatibility with two-argument overlays must be retained, make it explicit and observable — detect the narrow signature deliberately and emit a deprecation warning naming the overlay, rather than inferring it from which branch happens to be running.

[...]

PR #3894, opened 2026-07-29
fix(overlay): forward validate_pr keyword unconditionally + load-time signature conformance (#3526)

souliane:
fix(overlay): forward validate_pr keyword unconditionally + load-time signature conformance (#3526)

## What
Problem
- `_validation_errors` forwarded `require_sections` ONLY when it deviated
  from the default (`kwargs = {} if require_sections else {...}`). An overlay
  whose `validate_pr(title, description)` predates the keyword then worked on
  the common path and raised `TypeError` only on the rarely-taken one — a
  permanent contract violation turned into an intermittent runtime crash.

Fix
- `_validation_errors` always forwards `require_sections=require_sections`,
  so a non-conforming overlay fails on the first call, not the unlucky one.
- New `teatree.core.overlay_conformance` checks every facet override against
  its base signature at discovery time (`_discover_overlays`). An entry-point
  overlay with a dropped positional slot or dropped keyword-only parameter
  fails loud with `ImproperlyConfigured`; a registry overlay is warned and
  skipped (mirroring the existing non-subclass registry handling). Both name
  the overlay, the hook, and the missing parameter. A renamed positional
  parameter stays legal — leading arguments forward positionally, so only a
  dropped slot or dropped keyword-only name is a real break.
- Retires the conditional-forward idiom class: any future optional keyword
  added to a hook is now caught at load, not on the non-default path.

Behavior preservation
- The old idiom's only intended effect was two-argument-overlay tolerance;
  that tolerance was itself the bug (silent until the non-default path). The
  load-time check replaces it with an explicit, observable failure. No safety
  or leak gate touched.

Open questions & assumptions
- assumed: the framework always calls a facet hook with the parameters the
  BASE declares (the documented contract), so signature conformance is the
  right invariant to enforce. (status: assumed)
- decided: chose option 1 (always forward) + option 2 (load-time verify)
  from the issue, not option 3 (deprecation-warn two-arg overlays) — loud
  and early over compatibility with an undocumented signature. (status:
  assumed — no overlay in-tree relies on the two-arg form; t3-teatree
  conforms, pinned by a test.)
- scope: `config` is a pydantic data model, not a call-forwarded hook
  surface, so it is excluded from the scan; the six behavioural facets +
  OverlayBase carry the forwarded hooks. (status: decided)

## Why


souliane:
Merged `origin/main` into this branch (no rebase). One conflict, in `src/teatree/core/overlay_loader.py`: an adjacent-import add/add — this branch added the `overlay_conformance` import, main added `from teatree.core.overlay_name_resolution import cwd_overlay_name, overlay_name_of, resolve_overlay_name` (it extracted those helpers out of `overlay_loader.py`). Kept both imports; main s extraction applied cleanly elsewhere in the file.

Scope note after the merge: the code and tests of this PR are already on `main` — `overlay_conformance.py` and `test_overlay_conformance.py` on this branch are byte-identical to main s, and `overlay_loader.py` / `cli/tools.py` / the other test files show zero delta. The only thing this PR still adds is the BLUEPRINT § Overlay-System paragraph documenting the load-time signature-conformance mechanism, which `main` does not carry (`git show origin/main:BLUEPRINT.md | grep 3526` is empty). I verified the paragraph against main s code: `overlay_signature_violations` / `conforming_or_raise` / `conforming_or_none` exist in `src/teatree/core/overlay_conformance.py` and are called from `_discover_overlays`, so the text is accurate.

So this is now effectively a docs-only PR. The title still describes the code half that landed separately — worth retitling to something like `docs(blueprint): document load-time overlay signature conformance (#3526)` before merge, since the title becomes the squash commit message. I could not apply that edit myself: `gh pr edit` fails on this repo with a projects-classic GraphQL error, and the REST fallback is refused by the local validate-mr PreToolUse hook because a title-only PATCH presents an empty description to the validator.

CI at 4e855b5: all checks pass or skip, none red.

Issue #3879, opened 2026-07-29
execute_teardown re-enqueues ~2,900x per worktree — 98.6% of DBTaskResult is duplicate dispatch, not a retention gap


souliane:
98.6% of the remaining `DBTaskResult` rows are duplicate `execute_teardown` enqueues. This is a dispatch defect, not a retention gap — no retention setting can reach it, because the rows are younger than the library's minimum window.

## Measured

After [#3877](https://github.com/souliane/teatree/pull/3877)'s prune ran:

```
DBTaskResult remaining          649,672
  execute_teardown / SUCCESSFUL 640,742   (98.6%)
  distinct argument sets            220
```

**~2,900 duplicate enqueues per worktree within 24 hours.** The same teardown, for the same worktree, re-queued and re-succeeding thousands of times.

Retention cannot help: `django-tasks-db`'s `prune_db_task_results` has a 1-day floor, and these rows are minted continuously, so they are always inside it.

## Why this matters beyond the row count

This is the **second** defect of this exact shape found in one day. The first was `TicketTransition`: `mark_reviewed_externally` self-loops on `REVIEW_POSTED` by design, which made it idempotent in *state* but not in *side effects*, so a caller re-running it per pass wrote 3,240,987 no-op audit rows — 99.99% of that table ([#3876](https://github.com/souliane/teatree/pull/3876), merged).

Same signature here: an operation that is safe to repeat is being repeated at loop cadence, and each repetition leaves a durable row. The cost lands on the control DB, which is the seed **copied into every per-worktree env dir** — so a write amplification bug in one dispatcher becomes tens of gigabytes of disk (see [#3852](https://github.com/souliane/teatree/issues/3852)).

Worth asking whether this is a class rather than two incidents: *what else re-enqueues or re-fires on every tick because the operation is idempotent?* A sweep for that pattern is likely to find more.

## What the fix has to establish

Not "prune them faster" — the write is the bug.

- **Why is teardown re-enqueued for an already-torn-down worktree?** 220 argument sets against 640,742 rows means the dispatcher is not treating a completed teardown as terminal. Find the re-entry and stop it at the source.
- If a repeat is legitimate in some case, it should be **idempotent at the enqueue seam** — recognise the outstanding/completed job rather than minting a new one.
- A regression test that runs the dispatcher over an already-torn-down worktree N times and asserts a bounded number of enqueues, observed RED against today's behaviour.

## Acceptance

[...]

Issue #3899, opened 2026-07-29
quality-gates scans the branch tip, so no merge commit from an older branch can be committed


souliane:
## The defect

`scripts/hooks/check_quality_gates.py` scans `git diff --cached`. On a **merge commit**, that diff is against the *branch tip*, not the merge base — so every change `main` accumulated since the branch forked is presented to the gate as if the branch introduced it.

Concretely, resolving a merge of `main` into a 5-week-old branch, the gate flagged:

- main's own `pyproject.toml` `per-file-ignores."src/teatree/hooks/portable/*.py"` line
- an `evals/scenarios/*.yaml` line

Neither is in the branch's delta. Confirmed by running `teatree.quality.gate_relaxation.scan_relaxation` over `git diff --cached origin/main` (the correct base): **0 blocking findings**.

Because those are not inline code suppressions, the hook takes its `return 1` path rather than the WARNING path (`main()`, the `is_code_suppression` branch at `:155`).

## Why it matters more than one blocked commit

`skills/sweeping-prs/SKILL.md` documents the maintenance workflow as **merge `main` into the branch, never rebase** — merging is what preserves recovered commits whose author is gone, and the same rule appears in `/t3:ship`.

This gate makes that workflow **un-committable** for any branch older than the most recent `per-file-ignores` addition. And `quality-gates` has **no escape hatch** by design (#525) — `t3 <overlay> gate --help` confirms there is no `quality-gates` kill-switch. Its sibling, the anti-relaxation gate, ships `ALLOW_GATE_RELAX` and passes with a truthful reason; this one has nothing.

So the options at the point of failure are: abandon the merge, or `--no-verify` / prek `SKIP` — i.e. bypass every other gate too. That is a bad choice to force on anyone maintaining an old branch, and the likely outcome is the branch is simply left stale.

**This plausibly contributes to branch stranding.** A branch that cannot be refreshed cannot be made mergeable, so it sits until something else supersedes it. Of 66 orphaned branches triaged on 2026-07-29, 26 had "landed by another route" — the work was redone rather than the branch revived.

## Evidence

Hit live while resolving #3898. The resolution itself was complete and green — 68/68 in `tests/teatree_loop/test_resource_pressure_mechanical.py`, `ruff check` + `ruff format --check` clean, `tach check` green — and the commit still could not be made. The verified delta is preserved at `/home/teatree/agent-work-salvage/orphan-branches/3898-resolved-vs-main.patch`.

[...]

Issue #3891, opened 2026-07-29
17 of 53 agent scratch worktrees hold uncommitted or unpushed work that nothing observes


souliane:
## Measured 2026-07-29

`.claude/worktrees/` holds **53 agent scratch checkouts, 12 GB**. Of those, **17 carry real work that was never committed or never pushed**:

- **13** have modifications to *tracked source files* — 2 to 24 files each (not build artifacts; verified by separating `^??` untracked from `^[MADRC]` tracked)
- **4** have local commits on a branch that **does not exist on the remote**:
  `fix-3671-repair-halt-dedup`, `3869-fix-session-start-skill-autoload`,
  `feat/docker-containerized-cli-headless-settings`, `fix/3521-diff-coverage-msg`

The work survives only because a previous session hand-ran a salvage sweep into `/home/teatree/agent-work-salvage/` (167 files). That is a person noticing, not a mechanism.

## Two concrete gaps found while checking

**1. Two worktrees had no snapshot at all.** `fix-headless-claim-lease` and `wt-3521-recover` were outside the salvage set. `fix-headless-claim-lease` held **79 KB of STAGED changes across 21 files**.

**2. Staged work is invisible to the obvious salvage command.** `git diff` in that worktree returns **0 bytes** — the changes were in the index, not the working tree. Only `git diff HEAD` sees them. Any salvage or dirtiness check written as `git diff` silently reports "clean" on a worktree holding staged work.

The existing snapshots were spot-checked and *did* capture staged work correctly, so this is a latent trap rather than past data loss — but it is the kind that reads as success while failing, which is the recurring shape the last hand-off called out.

## Why this is a defect and not housekeeping

An agent that dies mid-task leaves tracked-file edits in a scratch worktree with no commit, no branch on the remote, and no record anywhere in the DB. Nothing observes it. `clean-all` correctly refuses to reap a dirty checkout (#706/#835), so the disk cost accumulates instead — and the guard that protects the work is the same guard that hides it, because a permanently-kept worktree looks identical to a legitimately busy one.

## Suggested shape

- when a dispatched agent's worktree is torn down, commit-or-snapshot any tracked modification automatically; never drop it silently
- surface "scratch worktrees holding unpushed work" as a doctor finding with an age, so 17 of them cannot sit unnoticed
- use `git diff HEAD` (or `git status --porcelain`) wherever dirtiness is decided — never bare `git diff`
[...]

Issue #3840, opened 2026-07-28
Ticket FSM never advances on merge: 205 tickets stuck in review_posted, exactly 1 merged — this is what makes the board look stale


souliane:
## The observation

The owner reports the dashboard shows stale information. It does — but the board is not the defective component. It is faithfully rendering a lifecycle that stopped being driven.

Six PRs merged at 13:15-13:16 today (#3816, #3827, #3824, #3806, #3836, #3815). Ten minutes later, the tickets for those PRs still render in the **NOT STARTED** column of `/dash/board/`, as `!3815`, `!3816`, `!3806`.

Live state distribution across all 328 tickets:

| state | count |
|---|---|
| `review_posted` | **205** |
| `ignored` | 60 |
| `not_started` | 55 |
| `delivered` | 5 |
| `planned` | 2 |
| **`merged`** | **1** |

One. Across the whole history, including a session that merged 44 PRs.

## What this means

The board groups tickets by `Ticket.State`. If the FSM never advances past `review_posted`, then 205 cards pile into the reviewing lane and stay there indefinitely, while merged work sits in Backlog. Every symptom the owner sees follows from that:

- merged PRs displayed as NOT STARTED,
- dwell counters reading `in column 135h 28m` and `146h 7m` — roughly six days, because nothing has moved those cards,
- a board that does not change no matter what actually ships.

The `tickets` loop is running and reports work — a manual tick returned `ran loop 'tickets' — 58 signal(s), 58 action(s)` — yet no ticket advanced and the total stayed at 328. Whatever those 58 actions were, advancing a merged ticket was not among them.

This is the "two pipelines, one walked" pattern recorded previously with the identical figures (205 / 1). It has not moved since, which suggests nothing currently closes the loop between "the PR merged on the forge" and "the ticket reaches MERGED".

## Not to be confused with

Newly filed issues (#3819, #3821-#3826, #3838, #3839) also do not appear on the board, but that is **expected** — tickets are created for PRs the owner authors and for issues that are assigned or labelled for the implementer. Those issues carry `bug` and no assignee, so their absence is correct behaviour, not staleness. Verified before filing so the two are not conflated.

## Also visible on the same board, worth separating out

1. **A raw Python traceback rendered as card body text**, beginning `Traceback (most recent call last): File "/opt/teatree/uv/tools/teatree/lib/python3.13/site-packages/claude_agent_sdk/_internal/tr…`, truncated mid-path. Operator diagnostics leaking into a work-overview card.
[...]

PR #3900, opened 2026-07-29
feat(cleanup): capture a checkout's unshipped work before anything may reap it

souliane:
feat(cleanup): capture a checkout's unshipped work before anything may reap it

## What
A reaping pass KEEPS a checkout holding work, and a kept checkout is
indistinguishable from a busy one — so 17 of 53 scratch checkouts (12 GB)
sat unobserved until a person ran a hand salvage sweep.

- `core/cleanup/unshipped_work.py`: read-only probe of the staged/unstaged/
  untracked delta plus the commits on no remote, written to the proven
  salvage-bundle shape (`.uncommitted.patch` / `.commits.patch` / `.files` /
  `.meta`) and to a durable `UnshippedWorkRecord` keyed on the checkout path.
- Wired ahead of the disposition in the raw-orphan reaper and ahead of the
  guards in `cleanup_worktree`, so it also covers the `force=True` hard-delete.
  Purely additive: no verdict changes, no #706/#835 guard is loosened, and a
  `--dry-run` still writes nothing.
- The patch is `git diff HEAD`: bare `git diff` compares the working tree to
  the INDEX, so a checkout whose whole delta is staged reports zero bytes.
  Two conformance walks pin that — an AST lint over every git-diff argv in
  `src/`, and a behavioural walk driving all ten dirtiness deciders against a
  real staged-only checkout.

Audit result: every existing decider already reads `git status --porcelain`
or a revision-anchored diff; the walks were observed RED against a planted
index-blind decider before landing.

## Why


souliane:
## Why

Work stops reaching `main` at five distinct points, and nothing observes any of them. Measured on this box 2026-07-29:

| state work dies in | measured |
|---|---|
| unstaged / staged / unpushed | **17 of 53** agent scratch worktrees, 12 GB |
| pushed, no PR | **68 remote branches** ahead of `main` |
| PR open, never merged | driver exists but skips silently |
| merged, FSM never advances | **496 of 498** `MergeClear` rows have `ticket_id` NULL against 208 real merges |
| terminal ticket re-swept | **770,773** duplicate task rows, growing ~42,000/h |

It survived only because someone hand-ran a salvage sweep. This PR makes each transition leave a durable, drained record instead.

**This branch demonstrated the bug while being created**: pushing it produced no PR, because `ensure-pr` defers on a first push and nothing drains the deferral. This description was written by hand for that reason.

## The two root causes

**The "no-orphan invariant" hands off to a caller that cannot exist.** `ensure-pr` runs at `stages: [push]` (pre-push). On a first push the remote ref does not exist yet, so it defers, exits 0, prints `Passed`, and creates nothing — its docstring says "the post-push `ensure-pr` opens the PR", but **git has no client-side post-push hook** and none exists here. No drain existed. All 68 orphans took that path.

**A `post_transition` receiver keyed on the target state alone fires on a state-preserving self-loop.** `mark_reviewed_externally` lists `REVIEW_POSTED` as its own source deliberately, so a re-review at a moved head SHA can re-stamp. Receivers that only ask "is the target terminal?" therefore mint work forever: 99.997% of the 770k duplicate rows come from 203 reviewer tickets holding **zero** worktrees between them. Same shape as #3876, whose fix sits 300 lines up in the same file.

## What lands

- **Capture before reap** — tracked, staged and unpushed work is snapshotted and recorded in the DB before any teardown, including `force=True` reapers.
- **The deferral becomes an obligation** — a `PendingPullRequest` row, drained by the existing `dispatch` scanner, with a doctor FAIL when one ages and `pr discharge-pending` as its escape.
- **The keystone closes the loop** — it now marks the `PullRequest` merged and adopts the owning ticket, so a merge advances the board. `ticket backfill-clears` recovers what it can historically.
- **The session-end check is armed** — it was gated on whether a lifecycle skill happened to load, which has nothing to do with whether work was stranded.
- **Skips stop being silent** — an aged `pr_sweep` skip surfaces with PR, reason and age.
- **#3879 fixed** — terminal tickets leave the replay candidate set, and teardown enqueues on *entering* terminal rather than on *being* terminal.

Two conformance walks make the classes unrepeatable rather than merely fixed: one fails any `post_transition` receiver blind to a self-loop (exempt set derived from the live FSMs, so a new self-loop turns it red rather than aging into a stale exemption), and one fails any skill doc naming a symbol the tree does not have — four orphaned branches had implemented `get_active_session`, which exists only as an illustration.

## Verification

`dev/ci-parity-fast.sh` EXIT=0 — **34,672 passed**, 69 skipped, 5 xfailed. Migration chain `0039 → 0040 → 0041`, `makemigrations --check` clean.

Every fix was observed RED before it was written. An independent cold review returned HOLD on three blocking defects — a capture that could crash a teardown, a swallow that dropped the obligation on a merely-busy DB, and a `repo_path` persisted as `"."` — all fixed and re-verified here.

## Two things on the record

**A `merge=generated` file silently dropped work during integration.** `.gitattributes` marks `docs/generated/management-commands.md` with a driver that discards both sides and re-runs the generator; the merge reported clean while losing a row. Caught and repaired, but it is a real hazard for any future multi-branch integration.

**Three deferred-import pegs were raised** in `deferred_import_pegs.toml`, each a genuine cycle/app-registry deferral with an inline rationale — the case the gate's own message sanctions. Restructuring those edges into declared tach sub-node edges is real work that was not attempted.

Closes #3879.
Relates to #3810, #3840, #3856, #3891, #3899.

souliane:
## Open finding on this PR: per-diff coverage gate (BLUEPRINT §17.6.3)

Not a CI failure — this is a local `PreToolUse` hook, and it is why this PR's description is in the comment above rather than in the body (the hook intercepts `gh api -X PATCH` on the body; `gh pr edit --body-file` separately fails repo-wide with a projects-classic GraphQL error).

The gate reports 13 new production symbols with no **name-level import** in any changed test:

```
ClearBackfillCommands, Command, DischargeResult, PendingPrCommands,
PendingPullRequestManager, PullRequestQuerySet, RelativeRepoPathError,
SweepSkipStreakManager, backfill_clear_tickets, defer_unpushed_pr,
is_missing_table_error, scan_file, status_porcelain_z_strict
```

It reads import statements only, so a `module.symbol(...)` attribute-access call does not count as a reference. Most of these are exercised — each contributing branch ran its own tests green — but through attribute access.

**The correct resolution, per symbol:**

- **Exercised** → add `from <module> import <symbol>` to the changed test that exercises it, and use the imported name in the body. No bare unused import (ruff `F401`), and no `# noqa` — that is tech debt this repo forbids without approval.
- **Not exercised** → that is a genuine coverage hole, not a gate artifact. **Write the test.**

An import added purely to silence the gate for an untested symbol converts a true finding into a false pass, which is precisely the failure class this PR exists to remove. Two attempts at this fix were lost to transient API errors (500, then 529); it is small and well-specified, but it should be done properly rather than quickly.

`Command` is likely a django-typer command class present in several modules — resolve which one the gate means from the diff before touching anything.

Issue #3583, opened 2026-07-21
Broken/orphan worktrees accumulate — periodic reaper should drop teatree-worktrees dirs that fail git rev-parse + collapse the /home/teatree/wt alternate root


souliane:
From the transcript analysis (2026-07-21). Worktree reaping lags creation.

## Evidence
- Worker log shows **16** `WARN Could not check git hooks: …/teatree-worktrees/<name> is not a git checkout (git rev-parse failed)` — for statusline-refresh, shard12-leak, bench-guard, oauth-switch, p3-toolgrant, leak-range, etc. These dirs are still registered/scanned but are broken checkouts.
- Transcripts: "stale worktree" in 61, `bench-guard` in 26. Resolvers repeatedly had to decide whether a sibling worktree at an old SHA was live before working.
- One agent worked from a **non-standard root** `an alternate worktree root` (at an old SHA), outside the canonical `the canonical worktrees dir` — splitting the namespace the reaper/doctor scan.

## Cost
Setup-time WARN spam on every session; agents burn time reasoning about stale siblings; risk of two worktrees for the same branch racing (see #3561).

## Fix
1. **Periodic broken-worktree reaper**: drop any `teatree-worktrees/*` dir that fails `git rev-parse` (a broken checkout can't hold live work). Fold into the existing reaper or `t3 doctor`.
2. **Single worktree root**: eliminate/redirect the `an alternate worktree root` alternate root so all worktrees live under the canonical `the canonical worktrees dir`; the reaper/doctor scan one place.
3. Surface a `t3 doctor` FAIL when a registered worktree dir is not a git checkout.

Partially related: #3489 (`clean-all --dry-run` only previews pass 1 — operators can't safely preview a cleanup) and #3551 (`--drive-subagents` handover touches other sessions' worktrees). This issue is the **broken-checkout accumulation** itself, not covered by those.

souliane:
Resolved by #3679 (merged as 498ae24b). Reopen if any part remains outstanding.

PR #3916, opened 2026-07-30
fix(workspace): reap broken worktree checkouts + doctor FAIL on registered non-checkouts (#3583)

souliane:
fix(workspace): reap broken worktree checkouts + doctor FAIL on registered non-checkouts (#3583)

## What
- reap `teatree-worktrees/*` dirs whose `.git` fails `git rev-parse` in
  `reap_orphan_isolated_worktree_roots` (folded into the existing clean-all
  reaper) — a broken checkout holds no git-recoverable work (its branch ref
  lived in the now-unreachable common dir), so it is dropped, not kept
- `t3 doctor` hard FAIL on a registered `Worktree` dir that exists but is not a
  git checkout (`_check_registered_worktree_checkouts`) — the reconciler's
  missing-dir finding never fires on present-but-broken, so it lingered silent
- shared `is_broken_checkout` predicate: single source of truth for reaper + doctor
- safety: a HEALTHY checkout and the unmappable-live keep are untouched

Scope vs #3583:
- done #1 periodic broken-worktree reaper (folded into clean-all; doctor surfaces each run)
- done #3 doctor FAIL on a present-but-not-a-checkout registered dir
- NOT done #2 collapse the alternate worktree root into one canonical dir

Open questions & assumptions:
- open: #2 (single worktree root) not implemented — needs an owner call on which
  root is canonical (the DB-isolation root vs config `worktree_root()`, which are
  deliberately separate) and whether to migrate live worktrees (high blast radius).
  Recommend a dedicated follow-up ticket.
- assumed: "periodic" is satisfied by the doctor FAIL (doctor/self-heal cadence)
  plus the existing clean-all reaper, per the ticket's "fold into the existing
  reaper or t3 doctor".
- assumed: adopted a prior session's uncommitted in-flight work in wt-3583 (no
  live sibling; ~8h stale) and committed it to make it durable.

docs: n/a — internal reaper/doctor hardening, no user-facing command or flag added

Relates to #3583

## Why

### DIFF
--- a/src/teatree/cli/doctor/checks_bootstrap.py
+++ b/src/teatree/cli/doctor/checks_bootstrap.py
@@ -1,6 +1,6 @@
 """Fresh-box bootstrap-hardening doctor checks (umbrella #3404).
 
-Four gates that turn silent late failures on a freshly-provisioned or migrated
+Five gates that turn silent late failures on a freshly-provisioned or migrated
 box into loud, up-front ones:
 
 - :func:`_check_gh_token_permissions` (#3405/#3477) — a missing REQUIRED permission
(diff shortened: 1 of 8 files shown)


souliane:
Superseded by #3935, merged at `221ea12b`.

This PR's reap path keyed deadness on `git rev-parse` failing, which reports a checkout created in a different execution context as dead. Held for that reason — it would have deleted live work.

#3935 ships the same capability with an evidence-based predicate instead: `NOT_A_CHECKOUT` requires no `.git` present *and* git agreeing there is no repository, and `report_unresolvable_worktree_dirs` is enforced non-destructive by the absence of a `shutil` import. #3583 is closed by it.

Delete-branch skipped — nothing here is unique to this branch.

PR #3929, opened 2026-07-30
fix(diff-coverage): name the import-only workaround in unreferenced-symbol failures (#3521)

### LINKED ISSUE #3521: diff_coverage: unreferenced_changed_symbols is blind to `mod.sym()` attribute access, and cannot be safely widened
## The false positive

`unreferenced_changed_symbols` in `src/teatree/utils/diff_coverage.py` decides "does a changed test reference this changed production symbol" from imports alone:

```python
elif isinstance(node, ast.Import):
    imported.update((alias.asname or alias.name).split(".")[0] for alias in node.names)
...
referenced = {s for s in all_symbols if s in imported_all and s not in shadowed_all}
```

`ast.walk` visits `Import`, `ImportFrom`, `FunctionDef`, `AsyncFunctionDef` and `ClassDef` only — never `ast.Attribute`. So a test written as

```python
import prod_module
...
prod_module.changed_symbol(...)
```

binds only the name `prod_module`. The symbol is reported unreferenced no matter how many call sites exist.

This blocked a real, green PR whose tests genuinely covered every flagged symbol — hand-counted qualified call sites of 16, 10, 6 and 3, backed by 14 individual behaviour reversions each verified red. A fifth symbol was covered indirectly through its caller; mutation testing (gutting the function body) took that suite from 24/24 to 23/24, so that coverage was real too.

## Why the obvious fix is wrong — please do not merge one

The natural fix is to also collect `ast.Attribute` accesses whose root is an imported name. This was implemented and then abandoned after adversarial review. Measured baseline-vs-fixed on runnable probes, the widened rule regressed 5 of 6 vacuity patterns from BLOCKED to LEAKING:

| Pattern | before | after |
| --- | --- | --- |
| stdlib collision — prod `def run()`, test only calls `subprocess.run([...])` | blocked | leaks |
| stub-out — `mod.widget = widget`, assertions hit the local copy | blocked | leaks |
| unreachable — `if False: mod.widget()`, real assertions hit the local copy | blocked | leaks |
| attribute on an imported class — prod `def cwd()`, test only calls `Path.cwd()` | blocked | leaks |
| pytest collision — prod `def raises()`, test only uses `pytest.raises(...)` | blocked | leaks |
| control — no import at all | blocked | blocked |

Narrowing to "only a call counts, only on the defining module" removes the collision cases but not the rest, because `patch.object(...)` followed by `mod.widget()`, and a `@pytest.mark.skip`ped test calling `mod.widget()`, are both syntactically calls. Syntax cannot see that one is stubbed and the other never runs. Signature-only diffs also launder: adding an optional parameter, adding a return annotation, or widening a type hint each left production gutted with tests green and the widened gate passing.

Body-coverage would resolve this semantically — it was verified to block every laundering probe while passing a genuine exercise — but it needs coverage data, and per `enforcement_tools.py:83` the structural check is by design the standalone check when there is none:

> `no coverage data at {coverage_file} — the per-diff line-coverage check measured nothing (only the symbol check ran)`

In that regime body-coverage would either flag everything or credit everything.

So the conclusion looks structural rather than like a patchable bug: any structural rule permissive enough to credit `mod.sym(...)` is permissive enough to be laundered by one no-op line, and adding the module-object route widens the surface without closing the pre-existing `from x import y` equivalents. Net worse than today.

## Suggested direction

Leaving the false positive in place is defensible — it is visible and fails loudly, whereas a leak is silent. Two things would make it much cheaper to live with:

1. **Name the workaround in the failure message.** Today it says "cover/reference it", which is misleading when the symbol is already covered. Saying that the check reads imports only, and that `from mod import sym` makes an existing reference visible, would have saved a lot of time here.
2. **Reconsider the shadowing rule's scope.** `shadowed_all` is collected with `ast.walk` across *all* changed test files, so any nested helper anywhere sharing a symbol's name kills that symbol's reference globally. Separately, `imported_all` is a flat module-blind set of bare names, so a symbol imported from an unrelated module already satisfies the gate for a same-named production symbol. Both look independently worth a look.

Happy to open a PR for the message change if that direction seems right.

souliane:
fix(diff-coverage): name the import-only workaround in unreferenced-symbol failures (#3521)

## Why

The per-diff coverage gate refuses a push when a changed production symbol is not referenced by a changed test. It reads a test's IMPORT statements only — a `module.symbol(...)` attribute call does not count — so a symbol that is thoroughly exercised can still fail the gate, and the message did not say how to satisfy it. Every author hit the same wall and had to rediscover the same workaround.

These commits were written when #3521 was worked, landed after that branch's PR merged, and were never pushed. Recovered rather than rewritten.

## What

The failure message now names the import-only rule and the exact remedy, so the gate teaches instead of just refusing. Covered by additions to `tests/teatree_utils/test_diff_coverage.py` and `tests/test_block_uncovered_diff_hook.py`.

Closes #3521


souliane:
Closing as already shipped, not stranded — my premise for opening this was wrong.

I opened it believing these were post-merge commits that never got pushed. The timestamps say otherwise: `9ec2ecd2` was authored 2026-07-23 04:09:47 and `33c64042` at 17:44:25, while #3677 merged at 19:38:49. Both PREDATE that merge — they ARE the work #3677 shipped, from a branch tip that was simply never pruned. `origin/main` carries the change byte-for-byte (merge commit `5b2cdb54`).

Merging would not be a harmless no-op: because main's copy sits at a different file offset, git sees no conflict and the merge adds ~30 lines of pure duplication — a second `_UNREFERENCED_SYMBOL_IMPORT_HINT` shadowing the first, and a duplicate `TestFindingNamesImportWorkaround` class that pytest would silently half-collect. CI's `lint` check is already FAILURE on the resulting ruff F811.

What led me astray, recorded so it does not recur: I used `git diff origin/main...branch`, which shows what the branch did since the merge-base and therefore still displays the full original work after a squash merge. It is not evidence that anything is missing from main.

The remaining actionable item from #3521 is the deeper fix — teaching the gate to credit `module.symbol(...)` attribute access rather than only name-level imports — which the commit message defers pending owner sign-off. That belongs on its own ticket, not on an open PR that ships nothing.

Issue #3502, opened 2026-07-20
statusline: optional opt-in to render in an explicitly engaged session when autoload is off


adrien-oper:
### Context

In a downstream vendored fork I had been carrying a local change to
`hooks/scripts/statusline.sh` that widened the render gate from

```sh
if [ -n "$session_id" ] && ! autoload_enabled; then
```

to also render when the session carries an engage marker
(`<session>.teatree-active` / `<session>.t3-engaged`).

I've now dropped that local change rather than re-applying it on each sync, because
it fights a deliberate design decision rather than fixing a bug: BLUEPRINT.md is
explicit that the loop statusline is a display-*visibility* surface and gates on
`autoload` **alone**, precisely so that a colleague who clones the repo and loads a
teatree skill (marker present, `autoload` off) does not get the statusline — the
#256 guarantee. My local change made exactly that case render, so as a default it
would be wrong.

Filing it here instead in case the underlying want is worth supporting explicitly.

### The want

When I engage a session by hand (`/teatree`) with `autoload` off, I'd like the
statusline in that session. Today the only way to get it is `autoload` on globally,
which is a different (and broader) thing than "show it in the session I explicitly
engaged".

### Possible shape

An opt-in setting, defaulting to `false`, so the #256 guarantee is unchanged for
anyone who doesn't set it:

```sh
# render when autoload is on, OR when this session was explicitly engaged AND
# the owner opted into that (default false → unchanged for a colleague).
session_engaged() { [ -n "$1" ] && { [ -f "$state_dir/$1.teatree-active" ] || [ -f "$state_dir/$1.t3-engaged" ]; }; }
```

`_statusline_chain_db()` already looks like the right template for reading such a
key — it does a cold `sqlite3` read of a DB-home setting with the WAL fallback, so
a new key wouldn't need new plumbing.

No strong opinion on whether this is worth the extra setting — happy for this to be
closed as working-as-intended if the `autoload`-alone gate is the line you want to
hold. Mostly filing so the reasoning is recorded rather than living as an
undocumented downstream patch.

### Not a bug

To be clear, nothing is broken: with `autoload` off the gate prints the neutral
hint line (#3233), so the bar is never mysteriously blank.

_Unrelated, while syncing: the `LC_ALL=C` work in #3311 fully supersedes the fix I
had carried for #3286, and the byte-mode UTF-8 glyph grouping is a genuine
improvement over what I had — that patch is now retired downstream. Thanks._

souliane:
[...]

Issue #3233, opened 2026-07-14
statusline.sh silently blanks (exit 0, no output) when autoload is off, only under CC non-TTY invocation


souliane:
## Bug
`hooks/scripts/statusline.sh` (~lines 127-129) exits 0 with ZERO stdout when `autoload_enabled` is false. Claude Code discards statusline output unless it is non-empty, so the entire status bar renders blank.

## Why it is a nasty trap
The gate only fires when `$session_id` is set, and `session_id` is only parsed when stdin is NOT a TTY. So:
- Run by hand in a terminal (TTY stdin): session_id empty, gate skipped, the 4 lines print perfectly.
- Invoked by Claude Code (JSON payload on stdin): session_id set, gate fires, silent exit 0, blank bar.

This makes the failure invisible to every run-the-script-directly debugging attempt. On a fresh box with no `autoload` DB row (the #256 opt-in default-OFF), the teatree statusline is silently blank under CC even though the script and terminal are perfect.

## Fix options
- `t3 setup` should set `autoload` (or prompt) so the shipped statusline works out of the box for the owner; and/or
- the script should not silently blank: when suppressed by the opt-in gate, emit a one-line hint instead of zero bytes so the bar is never mysteriously empty; and/or
- `t3 doctor` should detect statusLine-wired plus autoload-off and warn.

## Repro
printf a JSON payload with a session_id into the script: 0 bytes when autoload off (what CC sees); with T3_AUTOLOAD=1 it prints the 4 lines.

PR #3940, opened 2026-07-31
feat(statusline): default-off opt-in to render in an explicitly engaged session (#3502)

souliane:
feat(statusline): default-off opt-in to render in an explicitly engaged session (#3502)

## What
The loop statusline is a display-*visibility* surface and gates on `autoload`
alone (#256). An owner who does not want the broad `autoload`-on visibility but
does want the statusline in a session they engaged by hand (`/teatree`) had no
knob short of enabling `autoload` globally — a different, broader thing than
"show it in the session I explicitly engaged".

Add the default-OFF `statusline_in_engaged_session` opt-in. Only when set does an
EXPLICITLY-engaged session (the `.teatree-active` / `.t3-engaged` marker) render
the loop statusline with `autoload` off — the per-session counterpart of the
global `autoload` flag. Default-OFF keeps the #256 colleague guarantee intact:
marker present + autoload off + opt-in unset -> still blank.

- settings.py: register `statusline_in_engaged_session` (strict bool, default
  False) in OVERLAY_OVERRIDABLE_SETTINGS + the _WorkspaceCoreSettings field;
  DB-home only (no env layer) since only the bash statusline consumes it.
- statusline.sh: `_statusline_in_engaged_session_db_value` cold sqlite3 read
  (mirrors `_autoload_db_value`'s mode=ro -> immutable WAL fallback) +
  `statusline_engaged_render` gate; widen the suppression condition to
  `&& ! statusline_engaged_render`.
- BLUEPRINT.md + docs/blueprint/loop-topology.md: document the one widening.
- Tests: config round-trip + 5 render-gate cases. Anti-vacuity verified —
  reverting the gate widening turns the two render tests RED.

## Architecture pre-check — #3502

1. BLUEPRINT § alignment — loop-topology.md / BLUEPRINT "Session-start auto-load
   is opt-in (#256)": the statusline visibility gate. One default-OFF widening;
   BLUEPRINT + loop-topology.md updated in this same change.
2. FSM phase boundaries — n/a, no Ticket/Worktree state transition.
3. Extension-point contracts — none. No OverlayBase / scanner / hook Protocol
   change; the bash statusline is a leaf reader gaining one setting key.
4. Component boundaries — setting registration in src/teatree/config/settings.py
   (config registry); render gate in hooks/scripts/statusline.sh (statusline
   hook). No business logic in the hook beyond a cold read mirroring the autoload
   reader.
5. Dependency direction — no new imports (dict entry + dataclass field only);
   tach unaffected.
6. Test surface — TestStatuslineGating (subprocess vs the real bash) +
   test_statusline_engaged_session_setting (config round-trip) + the golden
   field test. Anti-vacuity proven.
7. Resilience invariants — the only external surface is a cold READ (no write).
   Fails silent/closed (empty -> no render) on missing sqlite3/DB/read error,
   mirroring _autoload_db_value; fallback transport = the #3233 neutral hint
   (bar never blank). verify-by-re-read / idempotency / heartbeat n/a (no write).
8. Identity/key normalization — n/a; single global key, exact-match read
   (scope='' AND key='statusline_in_engaged_session'), no bare/qualified duality.
9. Behavior preservation — purely additive widening behind a default-OFF flag.
   The #256 must-suppress behavior is preserved and pinned by
   test_no_optin_with_marker_shows_hint + test_optin_false_row_with_marker_shows_hint.
   No must-block test inverted; no privacy/leak matcher touched.
10. Removability / harness-vs-data — removable: delete the setting + the
    `&& ! statusline_engaged_render` clause + the reader to revert to the
    autoload-alone gate (blast radius: one gate line + one function). Harness-vs-
    data: the varying part (render in an engaged session?) is DATA (a DB config
    row the owner sets); the harness stays the generic gate reading it.

- Setting name `statusline_in_engaged_session` — the issue proposed an opt-in
  without naming it. status: assumed.
- DB-home only, no `T3_*` env override — only the bash statusline consumes it and
  the issue points at the `_statusline_chain_db` (DB-only) template. status: assumed.
- Honour EITHER engagement marker (`.teatree-active` OR `.t3-engaged`), matching
  the issue's suggested `session_engaged()`. status: assumed.
- Default OFF to preserve the #256 guarantee. status: decided-by-user (issue
  author requires default false).
- Whether the feature is worth an extra setting at all — the issue author had
  "no strong opinion … happy for this to be closed as working-as-intended".
  Implemented per the issue's "possible shape"; reviewer may prefer to hold the
  autoload-alone line and close instead. status: open.

## Why


souliane:
Closing as superseded — verified against `main`, not assumed.

The capability shipped under a different name: `statusline_engaged_render` (#3679, #3710), citing this same issue.

- `src/teatree/config/settings.py` — `statusline_engaged_render: bool = False`
- `src/teatree/config/defaults.toml` — `statusline_engaged_render = false`
- `hooks/scripts/statusline.sh` — engaged-session detection and the widened suppression condition
- Config and render-gate tests on `main` are a superset of this branch's, adding non-bool-ignored and broken-DB-fails-closed

Semantics are identical: default OFF, DB-home, no env layer, either engagement marker, #256 guarantee preserved. Landing this would add a second, duplicate setting alongside the one already shipped. #3502 is closed.

Issue #3872, opened 2026-07-29
clean-all still proposes deleting a live checkout's env dir after #3870 — a silent scan miss with no gap recorded


souliane:
After #3870 landed and deployed (container confirmed at `b602f486`), `t3 <overlay> workspace clean-all --dry-run` still proposes removing the isolated env dir of a checkout that exists and is registered as a git worktree.

**The reclaim was NOT run.** This is a pre-flight finding, not an incident.

## The contradiction

The dry-run emits:

```
WOULD Remove orphan isolated env dir: 1b2e4f54981d
```

But that env dir belongs to a live checkout. Evidence:

1. The SessionStart banner for this session named it explicitly — *"This entry point resolves the control DB at `…/teatree-worktrees/1b2e4f54981d/db.sqlite3` (worktree code with no explicit `XDG_DATA_HOME` is auto-isolated onto its own DB)"* — emitted while cwd was `…/.claude/worktrees/hook-python-django`.
2. That checkout exists right now and `git worktree list` in its clone reports it (126 worktrees registered, one of them this).
3. The env dir holds a 1.2 GB `db.sqlite3`, a `backups/` dir, a `quote-scanner.jsonl` last written 21:20, and a marker file whose name embeds this session's own id.
4. It carries **no `owner-checkout.path` stamp** — so the backfill added by #3870 never discovered its owning checkout either.

So the scan neither found the checkout nor recorded a gap: `complete=True`, and the dir is classified as an orphan.

## Why this matters more than the one directory

The env dir holds only a re-seedable DB copy, so deleting *this* one is probably recoverable. The problem is what it implies: **the keep-set has a silent hole, and a silent hole is exactly what #3870 set out to remove.** If one live checkout can be missed with no gap recorded, the 42-directory reclaim set cannot be trusted as a whole without knowing why.

The cold review of #3870 ran an independent cross-check — `find /home/teatree -xdev -name .git` returned 219 checkouts and the production scan returned 219, reported as exact set equality with `gaps=0`. That result and this observation cannot both be right, so one of them has a flaw worth finding:

- if the scan really found 219 including this checkout, then the slug→dir mapping is what is wrong, and `1b2e4f54981d` is not the slug the scan computes for that path;
- if the scan did not find it, then `find` and the production walk disagree, and the walk has a traversal gap the cross-check did not surface.

Both are worth knowing. The second is the more serious.

## Leading hypothesis, unconfirmed

[...]

PR #3942, opened 2026-07-31
fix(workspace): make every skip in the checkout scan record a gap (#3872)

souliane:
fix(workspace): make every skip in the checkout scan record a gap (#3872)

## What
`clean-all` proposed deleting a live checkout's isolated env dir while the
checkout registry reported `complete=True`. #3870 made the reaper fail closed on
a gap, but `scan_checkout_paths` had two skips that recorded no gap, so a miss
was indistinguishable from an empty answer — and completeness is the whole
evidence a deletion rests on.

Both were reproducible on the host, unchanged, before this fix:

- symlinked dirs were skipped outright — 87 under `$HOME`, and a
  bounded-vs-symlink-following diff of the same walk proved exactly one live
  checkout missing from the keep-set with no gap recorded
- the depth cap (10) truncated 208 subtrees silently; the deepest real tree is 15

Fix:

- walk THROUGH symlinked dirs, deduplicating recursion on the resolved path so a
  loop terminates and a tree reachable by several spellings is walked once
- probe `.git` before the dedup and before the listing, so a checkout is recorded
  under every spelling it is reached by and one whose contents cannot be listed
  is still recorded; the probe is guarded because `Path.exists` re-raises EACCES
  rather than answering False, which would have taken the whole scan down
- record a gap when the depth cap truncates, and raise it to 24 so it is runaway
  protection rather than a silent coverage policy
- gap an unresolvable directory

The reaper's fail-closed contract is unchanged: a gap still keeps every dir.
Measured after the fix on the same host — 71 checkout spellings (was 69,
including the previously-missed symlinked clone), zero gaps, 6.7s.

Open questions & assumptions

- assumed: following symlinks is preferable to gapping them. Gapping 87 symlinked
  dirs would leave the registry permanently incomplete, so the reaper would never
  reclaim anything — a safe but useless tool. Following them removes the miss
  instead of reporting it forever.
- assumed: `_NEVER_A_CHECKOUT` stays the one gap-free exclusion. Those dirs are
  package/venv/tool caches that cannot hold a checkout the resolver mints an env
  dir for.
- assumed: `_MAX_SCAN_DEPTH = 24` is headroom over the deepest measured tree (15),
  not a tuned value. Exceeding it is now loud, so the number is not load-bearing.
- open: the dir named in the ticket (`1b2e4f54981d`) cannot be re-diagnosed — its
  owning checkout has since been removed and its slug preimage is unrecoverable
  from a one-way hash, so which of the ticket's two branches applied to it
  specifically is unknowable. The defect class it pointed at is fixed and
  demonstrated on a currently-reproducible instance.
- open: the `except OSError` around `entry.is_dir()` stays uncovered (it was
  before this change too). It guards a permissions race mid-walk, which no
  deterministic test can stage; it is moved verbatim, not added.

## Why


souliane:
Closing as superseded — verified against `main`, not assumed.

All three commits are already on `main` from this same issue:

- "make every skip record a gap" → #3874. `main`'s `checkout_registry.py` is a strict superset of this branch's version, adding a `scanned_roots` field.
- depth cap → `main` already carries `_MAX_SCAN_DEPTH = 64`.
- venue-aware owner judgement → #3885. `main` replaces this branch's same-device (`st_dev`) classifier with `owner_stamps.venue_can_observe()`, which is strictly more conservative — scanned-root containment AND a listable parent — and additionally keeps unstamped dirs, which this branch explicitly deferred. `main` also ships the owner-stamp backfill and its CLI, which this branch does not.

Landing it would rename `open_for` back to `prepare_for` (breaking current callers and docs) and install a competing, less conservative liveness classifier. #3872 is closed.

Branch deleted.

PR #3960, opened 2026-07-31
fix(config,hooks): make three configured settings take effect (#3873, #3869, #3883)

### LINKED ISSUE #3869: autoload=true does not load skills at SessionStart — the only selection path is gated on UserPromptSubmit
Owner report: *"skill is not auto-loaded when session start — it should, considering I have the auto setting on."*

The setting is not being ignored. The injection point is simply not at session start.

## Measured

`autoload` is genuinely on:

```
key=autoload  value=true  source=file/env  scope=overlay:t3-teatree
```

And the engagement gate honours it — `_teatree_engaged` (`hooks/scripts/hook_router.py:509`) is an OR:

```python
return _autoload_enabled() or _teatree_active(session_id) or _t3_engaged(session_id)
```

so `autoload = true` alone engages the session. That part works.

## The actual gap

The only skill-selection path is guarded by a prompt and returns early without one (`hook_router.py:643-648`):

```python
"""Suggest cwd/overlay-context skills — never a free-text scan of the prompt."""
session_id = data.get("session_id", "")
prompt = data.get("prompt", "")
if not session_id or not prompt:
    return
```

That handler is on **`UserPromptSubmit`**. There is no SessionStart equivalent. The SessionStart path (`handle_session_start_bootstrap`, ~`hook_router.py:3786`) is the *singleton loop orchestration bootstrap* — it arms loops and the statusline, and never injects skills.

So the observable behaviour is:

- session starts → skills **not** loaded;
- owner sends the first message → skills **suggested**;
- the agent then loads them via the Skill tool.

Which is precisely "not auto-loaded at session start". The `autoload` flag's own docstring says it "both ENGAGES the session and ARMS its loops" — note it does not claim to load skills, so the code is self-consistent; it is the owner-facing contract that does not match.

## Why the distinction matters

Suggest-on-first-prompt has three failure modes the owner is hitting:

1. **The first turn runs unskilled.** Whatever the first message asks for is answered before the relevant skill is in context — and the first turn is often the one that sets the approach for the whole session.
2. **A headless/no-prompt session gets nothing.** A dispatched worker that never receives a `UserPromptSubmit` never triggers the suggester at all.
3. **It is a suggestion, not a load.** Nothing guarantees the suggested skill is actually loaded; that depends on the agent acting on the hint.

## Related, do not conflate

`_loop_auto_load_active` (`hook_router.py:516-534`) requires **both** `_teatree_active` AND `_autoload_enabled`, where the engagement seam one line above uses OR. That asymmetry is deliberate and documented (it gates loop/statusline arming so a colleague cloning the repo is never nagged), and it is **not** the cause here. Mentioned only so the next reader does not "fix" it by mistake.

## What is wanted

With `autoload = true`, a session should have its context skills loaded at SessionStart, before the first turn is answered — not suggested afterwards.

## Acceptance

- A SessionStart injection point that loads context skills when `autoload` is on, using the same resolver as the existing suggester rather than a second selection path (a second answer to "which skills apply" is the [#3854](https://github.com/souliane/teatree/issues/3854) duplication class).
- A test proving skills are present in the first turn's context with `autoload = true`, observed RED against today's behaviour.
- Default-OFF behaviour preserved: a session with `autoload = false` and no teatree skill loaded must still get nothing, and the never-lockout property of the PreToolUse skill gate must be unchanged.
- The headless dispatch path is covered too, or explicitly stated as out of scope with a reason.

### LINKED ISSUE #3873: ConfigSetting override reads fail under load and silently fall back to shipped defaults — safety gates degrade with no operator-visible signal
Under concurrency, `ConfigSetting` override reads fail and the resolver falls back to compiled-in dataclass defaults — **including for safety gates** — and the only signal is a log line.

## Observed

```
2026-07-29 04:27:01 ERROR teatree.config ConfigSetting global-scope override read FAILED unexpectedly
  — resolving with NO DB override tier for this read (safety gates fall back to dataclass defaults).
    This is a real read fault, not a bootstrap no-op; fix the DB/read error.
2026-07-29 04:27:01 ERROR teatree.config ConfigSetting overlay 't3-teatree'-scope override read FAILED unexpectedly
  — [same]
```

Intermittent and ongoing: **4 occurrences in a 25-minute window**, while a direct read at the same moment succeeds:

```
$ t3 <overlay> config_setting get autoload
  autoload = True  [source: db, global]
```

So the row is present and readable; the failure is transient, not a missing value.

## Conditions

Load average ~11, nine concurrent agents plus the worker fleet, all against the same control DB. The DB is **1.2 GB** with `auto_vacuum=0` and roughly half the file dead pages, which makes every read touch a larger file than it needs to (see [#3871](https://github.com/souliane/teatree/issues/3871) for the retention side and [#3852](https://github.com/souliane/teatree/issues/3852) for the size).

The first fault at 04:27 predates any of tonight's reclaim work, so it is not an artefact of directories being removed.

## Why this is worse than a transient read error

The message says it plainly: *"safety gates fall back to dataclass defaults."*

That means under load, a gate can silently evaluate against a **shipped default instead of the operator's configured value**. Every knob that exists to restrain autonomous behaviour — merge approval, admission ceilings, disabled loops, private-repo allowlists — is resolved through this path. A gate that reads `false` because the DB read failed, rather than because the operator set `false`, is indistinguishable from a correctly-resolved gate at the call site.

This is the same failure shape as several defects found this week: a partial failure presenting as a definite answer. The privacy matcher truncating a real address, the admission governor reading a partial quota outage as total, the `clean-all` keep-set coming back silently short — and now a config read returning "no override" when it means "I could not tell".

## What the fix has to establish

Not merely "retry the read". The distinction that must survive is **"there is no override" vs "I could not determine whether there is an override"**, and those must not resolve the same way.

Concretely, worth considering:

- a failed override read should be **fail-closed for safety-relevant keys** rather than silently permissive — a gate whose value could not be resolved should refuse, not proceed on a default;
- alternatively the read should retry with backoff and only then raise, so the caller decides rather than the resolver deciding for it;
- and the fault should surface somewhere an operator sees, not only in a worker log. There are doctor alarms for table size, loop freeze, question age and dead-ticket spend; there is none for "the config tier degraded".

## Reproduction

Not reproduced deterministically. It appears under sustained concurrent load; the four occurrences in 25 minutes coincided with nine agents running. A reproduction harness that hammers `ConfigSetting` reads from N processes against a large DB would establish it, and would also serve as the regression test.

## Acceptance

- A failed override read is distinguishable from an absent override at every call site that gates behaviour.
- Safety-relevant keys fail closed rather than falling back to a shipped default.
- A test proving a gate does NOT silently permit when its override read fails, observed RED against today's behaviour.
- The condition is visible to an operator without reading worker logs.

### LINKED ISSUE #3883: The headless posture is configured but unenforced — an interactive session can implement teatree directly
The `headless` posture is a configured value that **nothing enforces**. An interactive Claude Code session can implement teatree directly — edit `src/`, commit, push, and spawn `t3:coder` agents — while the configuration says all such work runs headless through the factory. The setting is true and inert at the same time.

Owner: *"teatree is supposed to autoload in claude code, and it's configured to run headless so in practice you should not have started coding by yourself."*

## What happened

On 2026-07-29 the main interactive session implemented teatree directly for an entire working day: it hand-wrote fixes, committed, pushed, opened and merged PRs, and dispatched ten `t3:coder`/`t3:tester`/`t3:e2e` agents. Nothing refused any of it.

The consequences were measurable, and they are the argument for the fix:

- ten concurrent agents drove load to **44** and RAM to **92%**, which tripped the `budget:low_ram` guard and **stalled `directive_loop` with 24 owner directives unconsumed**;
- the disk reached **0 bytes free**, and a worktree control DB **corrupted** (`~/.local/share/teatree-worktrees/c236659ccc03/db.sqlite3.corrupt-20260729`);
- roughly **520k subagent tokens** produced two merged PRs and one reproduction, because agents contended for one disk and were restarted mid-task;
- a coordinator scope-cut, written from a stale mental model, told an agent its finished work was out of scope — 107 KB of working implementation was nearly lost.

None of this was a capability failure. Each step was individually reasonable. The failure is that **no state made the wrong thing impossible**.

## Two defects, one symptom

**1. The instruction never arrives (`#3869`).** `autoload` is genuinely on and `_teatree_engaged` (`hook_router.py:509`) is an OR, so the session IS engaged. But the only skill-selection path is guarded by a prompt and returns early without one (`hook_router.py:643-648`), and it is wired to `UserPromptSubmit`. `handle_session_start_bootstrap` (~`hook_router.py:3786`) arms loops and the statusline and never injects skills. So the first turn of every session runs unskilled, and a session that never receives a `UserPromptSubmit` gets nothing at all.

**2. Even when it arrives, it is only an instruction.** A loaded skill saying "route t3 work to the factory" is prose an agent may ignore, forget, or reason around — and this one did, repeatedly, across a day, while apologising each time. `AGENTS.md` now opens with the owner's First Principles, which state the rule this issue exists to satisfy: *instructions are the fallback for whatever cannot be ENFORCED deterministically.* This can be enforced. It is not.

## What the fix must establish

A **deterministic refusal**, not a stronger instruction.

When the session is teatree-engaged AND the configured posture is headless, a `PreToolUse` gate refuses:

- `Edit` / `Write` / `NotebookEdit` targeting the teatree repo's `src/`, `tests/`, `skills/`, `hooks/`, `docs/`;
- `Agent` dispatches of implementation subagent types (`t3:coder`, `t3:debugger`, `t3:tester`, `t3:e2e`, `t3:shipper`);
- `Bash` invocations that commit or push to the teatree repo.

The refusal message must name the remedy — file or find the issue, let intake pick it up — rather than merely blocking. A gate that says "no" without saying "do this instead" produces an agent that retries creatively, which is worse.

### What must stay allowed

The main session's actual role, which is genuinely valuable and must not be gated:

- **reading, searching, diagnosing** — the root-causing that produced #3876, #3879, #3882 was main-session work and should stay;
- **reviewing and merging** — `ticket clear` / `ticket merge` are the coordinator's job;
- **answering deferred questions**, filing and commenting on issues;
- **host operations the factory cannot do** — disk reclaim, DB prune, killing orphaned processes.

The line is: **the main session monitors and dogfoods; the factory implements.**

The owner's framing is narrower than "coordinator": *"then it's dogfooding only, you here just monitoring."* So the main session's remaining job is to **use the factory and report what it finds** — run the CLI, watch the loops, read the dashboard, notice what is broken, file it. That is dogfooding, and it is the activity that produced most of today's findings. It is not a licence to fix what it finds; noticing and fixing are different actions, and only the first belongs here.

That distinction matters for the gate's design: dogfooding **requires** running `t3` commands, including ones that mutate host state (prune, reclaim, kill a stuck process). Those must stay allowed. What is refused is authoring the repo's own source.

### Never-lockout, and the state-reconciliation carve-out

This repo has a standing never-lockout contract, and it applies here. Two exemptions are needed:

1. **An explicit, audited override**, following the existing `e2e-bypass` / `dod-override` / `plan-bypass` pattern — single-use, recorded, with a reason. A genuine emergency (the factory itself is down, a fix is needed to restore it) must not be blocked by the thing it is trying to repair.
2. **Work already in flight.** The owner's framing: work should not be handed back to the factory when *"it's not possible to do it headless because it already started here and the state reconciliation is cumbersome."* An in-flight branch with uncommitted state in a main-session worktree is finished where it started; only NEW work is refused. The gate keys on "is there already a live checkout for this ticket in this session" — not on the ticket's mere existence.

## Why a gate is the right tool here despite the doctrine

`AGENTS.md` notes that for an **AI actor** a gate fires after the mistake, so the instruction is the cheaper control. That is true for behaviour with no durable side effect. It does not hold here: the mistake's cost is not a repair cycle, it is a **stalled directive loop, a filled disk, a corrupted DB, and near-loss of a colleague's work**. When the side effect is durable and expensive, the net has to be underneath.

The instruction half should still be fixed (#3869) — the gate is the backstop, not the design.

## Acceptance

- With the headless posture configured and the session engaged, an `Edit` to `src/teatree/**` is **refused**, and the refusal names the factory route. Test observed RED against today's behaviour.
- An `Agent` dispatch of `t3:coder` is refused under the same condition, same evidence.
- Read-only inspection, review/merge commands, question answering and issue filing are **not** refused — pinned, so a future tightening cannot silently swallow the coordinator's role.
- The audited override unblocks exactly one action and is recorded with its reason.
- The in-flight carve-out is testable: a ticket with an existing live checkout in this session is not refused; the same ticket without one is.
- With the posture NOT headless, nothing is refused — default-off preserved, so a colleague cloning the repo is never gated.
- `#3869` is fixed or explicitly deferred with a reason; if deferred, this gate must not depend on the skill having loaded.

## Related

[#3869](https://github.com/souliane/teatree/issues/3869) (autoload never fires at SessionStart), [#3855](https://github.com/souliane/teatree/issues/3855) (headless-first direction), [#3848](https://github.com/souliane/teatree/issues/3848) (make the contract structural rather than prose), and the `no-hand-driven-t3-work` memory — which is itself only an instruction, and therefore exactly the thing this issue exists to replace.

souliane:
fix(config,hooks): make three configured settings take effect (#3873, #3869, #3883)

## What
Three settings that were stored and then ignored, each diverging from the
configured value with nothing reporting the divergence.

#3873: a failed ConfigSetting override read returned an empty mapping, which is
also what a healthy read of an empty table returns, so an unreadable tier and an
absent one resolved identically and the gates fell back to shipped defaults. Two
of those defaults are the most permissive value the setting has, so stored
restraint was widened by a read that merely failed. The read now retries a
bounded number of times, then reports the scope degraded rather than empty; the
autonomy and approval keys take their most restrictive value while the tier is
unreadable; provenance reports the tier as unresolved instead of crediting one
never consulted; a TOML export refuses rather than persisting an unverified
absence; and the degradation is recorded beside the control DB where a new
doctor check surfaces it. Environment overrides keep winning.

#3869: autoload engaged a session but loaded no skills, because the only
selection path is wired to UserPromptSubmit and returns early without a prompt.
SessionStart now resolves the same skills through the same resolver and writes
the same demand set the PreToolUse gate reads, before the first turn, and runs
before the statusline seed that would otherwise suppress it. Default-off and
never-lockout behaviour is unchanged.

#3883: a new PreToolUse gate refuses interactive authoring of this repo while
the resolved runtime is headless: edits under the authored directories,
dispatches of implementation sub-agents, and commit or push. Inspection, review,
merge, answering, issue filing and every t3 command stay allowed. The gate keys
on which lane is acting rather than on the path, because the factory's own
workers run through the SDK with this same hook set. It fails open: an
unreadable lane or runtime allows. Work already in flight in a linked worktree
is exempt, and an audited single-use token unblocks one call and records why.

Closes #3873
Closes #3869
Closes #3883

## Why


souliane:
## RED evidence, per test

Each regression test was observed failing against unchanged behaviour before its fix, with a passing foil beside it so that neither a uniformly-permissive nor a uniformly-restrictive implementation could satisfy the suite.

**#3873 — 8 RED, 8 controls green.** The sharpest case: a stored `autonomy = babysit` resolved to `FULL` once the override read failed, because the shipped default is the most permissive value the setting has. Also RED: the scope was not marked degraded, the merge/answer gates did not fail closed, `mode` and the on-behalf posting mode did not fail closed, provenance credited the shipped file for a tier never consulted, and the read was not retried at all. Green throughout (proving non-vacuity): a clean read of an empty table stays undegraded, a bootstrap state stays silent, the shipped default still wins on a healthy read, and an environment override still beats the fail-closed value.

**#3869 — 3 RED, 4 controls green.** RED: no skills named in the SessionStart context, no demand set written for the PreToolUse gate, and the statusline seed suppressing the selection. Green: an unengaged session gets nothing, an empty selection writes no demand (so the gate can never become a lockout), and a failing selection costs the hint rather than the session.

**#3883 — 7 RED, 12 controls green.** RED: the edit was not refused, the implementation sub-agent dispatch was not refused, the refusal did not name a remedy, and the audited override had no effect and left no record. Green from the start, and the assertions that matter most: a factory-dispatched SDK session is not refused, an unreadable lane allows, an unreadable runtime allows, and the coordinator surface (inspection, review, merge, issue filing, host-mutating `t3` commands) is untouched.

## What happens when the #3883 gate cannot answer

It allows. This is the one gate here that fails open, and the asymmetry is deliberate because it inverts the house rule.

- lane signal absent or unrecognised → allow;
- runtime unreadable (teatree not importable from the hook interpreter, DB unreachable, unrecognised value) → allow;
- any internal error in the gate → allow;
- repo or path unresolvable → allow.

A wrong refusal halts the entire factory and every SDK consumer at once, and would present as every headless task refusing with no obvious cause. A wrong allow costs one hand-written edit that a human notices. An "unknown" verdict is not evidence of an interactive session, so it must never be treated as one.

## A note on the contention reproduction

The retry test reproduces the mechanism — one contended read raises, the next succeeds — not the load. A genuine multi-process reproduction against a large control DB is not attempted here, and the fix does not depend on one: the defect is in how the failure is typed, not in how often it happens, and that is reachable deterministically. Adding a retry without fixing the type would have hidden the bug rather than fixed it.

## Caught during the work

The type checker found that the runtime read had a wrong call signature. It would have raised on every call, been swallowed by the fail-open, and left the gate permanently inert — while every test that mocked the reader stayed green. That is the same vacuity shape the tests here are built to avoid, so the real reader now has its own cases against a real control DB.

Issue #3366, opened 2026-07-17
statusline: loop:inbox renders ·DRIVERLESS despite actively ticking


souliane:
## Symptom
The rendered statusline top bar shows `loop:inbox <Nm>·DRIVERLESS` (yellow alert palette) for the `inbox` mini-loop -- the only one among ~12 rendered loop rows lacking a driver-liveness suffix.

Reproduced identically across two independent isolated ticks (NO_COLOR and colorized), so it is not a one-off render race.

## Expected vs actual
`t3 loop list` run directly against the live worker shows inbox as enabled and actively cycling on its normal ~1 minute cadence (recent last-tick, near-term next-tick). But the statusline's DRIVERLESS marker (src/teatree/loop/driver_detection.py, statusline_loops.py around lines 150 and 392) is documented to mean a blank driver on an owned loop row that looks healthy but never ticks. Those two signals directly contradict each other for this one loop.

## Hypothesis (unconfirmed -- for whoever picks this up)
inbox's LoopLease row is likely stale or orphaned from before its recent enable via PR #3348, which fixed Loop.enabled plus a separate stale per-loop override that had been masking it (see memory note inbox-loop-override-mask-vs-enabled). The enable path may never have re-registered or refreshed a driver lease for it, even though the underlying Loop DB row is ticking fine via whatever mechanism actually runs it now.

## Impact
At minimum this is a misleading-observability bug -- an operator glancing at the statusline would reasonably conclude the loop is broken or dead when it is not. It may also indicate a real gap in the driver-registration path for loops enabled via the sanctioned preset/override mechanism, as opposed to ones enabled since initial deploy -- worth checking whether other loops enabled via the same path show the same drift.

## Where to look
- src/teatree/loop/driver_detection.py
- src/teatree/loop/statusline_loops.py (around lines 150 and 392)
- The LoopLease model and whatever writes or refreshes it on enable, cross-referenced against the override-clear path introduced by PR #3348

Filed as part of a dogfooding pass (t3:dogfooding-teatree skill, Hunt for bugs mode), user-confirmed for filing and fixing.

PR #3975, opened 2026-07-31
fix(statusline): suppress loop:<name> ·DRIVERLESS while the loop is ticking (#3366)

souliane:
fix(statusline): suppress loop:<name> ·DRIVERLESS while the loop is ticking (#3366)

## What
The per-loop `·DRIVERLESS` chip read only the owner LoopLease's stored
`driver`, which an anonymous worker/cron tick never rewrites (it runs with an
empty session_id and so never re-claims the owner lease). For a loop enabled
via the preset/override path, that stored driver fossilises blank while the
loop ticks fine on its normal cadence — so `loop:inbox` painted ·DRIVERLESS in
the alert palette even though `t3 loop list` showed it actively cycling. The
two observability surfaces contradicted each other.

·DRIVERLESS means "claimed but never ticks". Gate the per-loop chip on the
authoritative cadence ledger: suppress it when `Loop.last_run_at` is within two
of the loop's own `delay_seconds` cadences — the same STALL_FACTOR=2 threshold
`teatree.loops.live` (the `t3 loop list` source) uses, so the two surfaces can
no longer drift. Fails SAFE to keeping the alert on any unconfirmable read
(missing / never-run row, no fixed cadence, DB error).

Only the `loop:<name>` chunk is gated; the master-anchor DRIVERLESS path and
the loud claim-time warning are unchanged.

Open questions & assumptions:
- (assumed) The right liveness threshold is loops.live's STALL_FACTOR=2 × the
  loop's own delay_seconds, matching `t3 loop list`. Chosen over the statusline
  render-age floor (max(2*cadence, 300s)) because the bug is defined by
  disagreement with `t3 loop list`, not with the render-age freshness gate.
- (assumed) The fix belongs on the read side (statusline) rather than the write
  side (refreshing the owner lease driver on anonymous ticks): the read-side
  gate is the minimal, safe correction and never touches the claim/lease write
  path. A write-side self-heal for anonymous-tick loops may be worth a
  follow-up but is out of scope here.
- (assumed) STALL_FACTOR cannot be imported into loop_scoping (it would form a
  loop_scoping -> loops.live -> statusline_loops -> loop_scoping cycle that tach
  forbids), so the "2 cadences" multiplier is a documented local constant
  mirroring it, and Loop is reached via apps.get_model to avoid a new tach edge.
- (open, pre-existing, unrelated) tests/teatree_loop/test_outbound_audit.py::
  GitHubNoteVerifierTests::test_returns_none_when_github_module_unimportable
  fails on clean HEAD in this environment — not touched by this change.

## Why


souliane:
Closing as superseded — the #3366 fix is already on main.

`loop_is_actively_ticking` landed in `loop/loop_scoping.py` and the DRIVERLESS suppression now lives in `loop/statusline_loop_chunks.py` after that module was split out of `statusline_loops.py`. The test file here is byte-identical in test set to the one already on main.

This branch's last sync with main predates that refactor, so merging it would revert a large amount of unrelated work rather than add anything. Nothing here is lost.

Issue #3978, opened 2026-07-31
Issue intake deadlocks at a budget of 2 and reports itself healthy while admitting nothing


souliane:
## What

Issue intake can deadlock completely, and reports itself healthy while it does.

The intake scanner is gated on an in-flight budget: it claims a new issue only while
`in_flight_count < issue_implementer_max_concurrent`. The shipped concurrency is **2**, and a
claim that stops progressing is only released after a **24h** stall grace. So **two** stalled
claims are enough to stop the factory admitting any work at all, for a day at a time.

Observed live: both slots held, `in_flight=2/2`, with seven admitted issues in the queue and
**zero** tickets, pending tasks or claimed tasks anywhere. The loop was enabled and ticking on
cadence the whole time.

## Why nothing surfaces it

Every signal available to an operator reads normal:

- the loop is `enabled` and its last-run stamp advances every tick;
- it does not error — a full budget makes the scanner factory return `None`, so the tick
  legitimately does nothing and reports success;
- the queue looks like a queue: labelled issues accumulate with no indication they are
  unreachable.

There is no surface anywhere that says "intake is at budget and claiming nothing". The only
way to see it is to read the marker ledger directly.

## How the two slots got stuck

Two different failure shapes, neither of which the graces catch quickly:

1. **Work succeeded, ticket never caught up.** The claim's ticket reached `shipped` and its PR
   merged — but the ticket stayed at `shipped`, which is deliberately NOT a marker-release
   state (shipped means "PR open, not yet landed"). With the ticket frozen one step short of a
   releasing state, the slot is held until the 24h stall grace expires. The release condition
   and the thing that actually happened never meet.
2. **Dispatch died at once.** The claim created a ticket whose only task failed immediately;
   the ticket never left `not_started`. Nothing retried it, and it too held its slot for the
   full grace.

Both are the *normal* outcomes of an attempt that fails — which is what makes the deadlock
reachable in ordinary operation rather than only under some rare fault.

## Suggested direction

The graces and the release states are individually defensible; the problem is the combination
with a budget of 2 and no visibility.

- A ticket whose PR has landed should release its slot on that fact, rather than waiting out a
  stall grace for a step the ticket will never take on its own.
- A claim whose ticket has no live work and no open PR is not "slow", and should not be held to
  the same grace as one that is genuinely mid-flight.
- Intake being at budget must be **visible**: a tick that claims nothing because the budget is
  full should say so, and a budget that stays full while admitted issues wait should fail
[...]

Issue #2385, opened 2026-06-14
Layer teatree.core in tach + split the flat core root to surface the ~347 masked intra-core deferred-import cycles


souliane:
Surfaced by the 2026-06-14 architectural review (rank 14). tach.toml defines teatree.core as ONE undivided module with forbid_circular_dependencies=true, so intra-core cycles are invisible by construction. There are ~347 '# noqa: PLC0415' function-local deferred imports inside src/teatree/core papering over a real bidirectional cycle: models/ticket.py function-locals 'from teatree.core.tasks import execute_provision' while tasks.py top-levels 'from teatree.core.models import ...'; models also defer-import up into gates, with a top-level model->gate edge at local_stack_queue.py.

The flat core root has ~85 loose files that cluster cleanly (review/, worktree/, messaging/, identity/, backends_glue/), and models/__init__.py is a 176-line hand-curated barrel that drifts on every model change.

Direction: split tach's teatree.core into layered sub-modules (models as a leaf; gates/tasks/runners above, never the reverse) with an import-linter contract forbidding models->tasks/gates; dispatch transitions via an injected callable/signal instead of function-local imports; promote the obvious clusters to sub-packages (git mv, one cluster per PR); generate or drop the barrel. Large blast radius — needs design sign-off and should land cluster-by-cluster, not in one PR. Pairs with grouping the 8 transient *_marker idempotency rows into models/markers/ (rank 9, lands independently).

souliane:
## Done — `teatree.core` is now sub-layered and the cycle is structurally forbidden

Shipped as a 3-PR sequence, each cold-reviewed (maker ≠ checker) + keystone-merged:

- **#2408** (PR-1) — carved `core/modelkit/` leaf (`fibonacci`/`db_retry`/`phases` moved); installed the intra-core deferred-import ratchet `tests/quality/test_intra_core_deferred_import_ratchet.py` (FROZEN=217). Empirical finding: only `modelkit` was declarable in PR-1; `models`/`managers` carried ~50 deferred up-edges into parent-core that had to be severed first.
- **#2411** (PR-2a) — severed 21 deferred up-edges: tasks/worktree_tasks → `core/signals.py` post_transition receivers; gates/overlay_loader/cost → `core/modelkit/gate_registry.py`; `ReviewState` → `core/modelkit/review_state.py`. Added `tests/quality/test_models_no_parent_core_up_edges.py`. FROZEN 217→206.
[...]

PR #3981, opened 2026-07-31
fix(intake): release a budget slot on evidence and surface a jammed one (#3978)

souliane:
fix(intake): release a budget slot on evidence and surface a jammed one (#3978)

Intake claims only while in_flight_count < issue_implementer_max_concurrent, so two
claims that stop progressing close intake entirely at a budget of two, a day at a time.
Nothing surfaced it: a full budget makes the scanner factory return None, so the tick
does nothing and reports success — enabled loop, advancing last-run stamp, no error,
labelled issues piling up unreachable.

- Release on the fact that ends the attempt, not on a grace the ticket cannot reach. A
  MERGED PR completes its marker immediately (SHIPPED is deliberately not a
  marker_release_state, so a ticket frozen there after its PR landed satisfied no
  release condition at all). A ticket with nothing queued and no open PR is held to a
  2h dead-claim grace, capped so it can never outlast the 24h mid-flight one.
- One reading of the budget — core.intake.budget.read_intake_budget — shared by both new
  surfaces, so the tick and the operator cannot disagree about whether intake is jammed.
- The tick logs IntakeBudget.report() naming every held slot instead of declining
  silently; `t3 doctor check` HARD-FAILs (_check_intake_budget_deadlock) when a full
  budget is held entirely by claims with no active task and no open PR.
- `t3 loop reclaim-markers` exposes --dead-grace-hours, bounded at zero like its siblings.

Both new release rules observed RED before the fix.

Open questions & assumptions

- assumed: "doctor fails while admitted issues are unclaimed" is implemented as the
  DB-observable signature "full budget, zero progressing holders". `t3 doctor check` is
  offline and cannot enumerate waiting forge issues without a network read per run; the
  chosen signature fires in exactly the reported scenario and never on a busy factory.
- assumed: 2h for the dead-claim grace. A live attempt queues its next task within
  minutes, and the existing "recent work keeps the slot" case is pinned at 1h. It stays
  a parameter (find_stale(dead_grace=...), --dead-grace-hours), not a config row — no
  per-overlay policy varies it today.
- assumed: a 30-minute settle window before a claim can be judged non-progressing. A
  dispatch lands its ticket and first task seconds after the marker row is written, so
  judging inside that window would read every healthy claim as stuck.
- assumed: an early release cannot cause a duplicate dispatch — the ticket still owns
  its issue_url in the intake scanner's _ACTIVE_TICKET_STATES, so rule 2 still refuses.
- open: whether issue_implementer_max_concurrent should default above 3. The issue
  raises it only as a "consider"; left unchanged, since with slots released on merge and
  dead claims released in hours the reported deadlock no longer turns on budget size.
- assumed: the two new deferred-import pegs are load-bearing edges rather than lazy-import
  convenience. Verified by reading every import site: `budget.py`'s four are ORM reads
  (`ImplementedIssueMarker` / `Ticket` / `Task` / `PullRequest`) in a module reached from
  the scanner factory, which loads before the app registry is populated; the two added in
  `implemented_issue_marker.py` are `PullRequest` reads on a peer model in the same
  package, deferred for the same reason the existing `Ticket` deferral there is.
- assumed: raising the pegs is the sanctioned resolution over promoting the edges to
  declared tach sub-node edges (#2385). The peg file's own header permits a raise when the
  edge is genuinely load-bearing, with the rationale carried in the commit message.

## Why


souliane:
CI is red on three checks, all from one root cause: the intra-core deferred-import ratchet.

`test (3.13)` is only the aggregate gate ("require every test shard to have passed"), so it
goes green on its own once the two shards do. The real failures are:

- `tests/quality/test_intra_core_deferred_import_ratchet.py::TestIntraCoreDeferredImportRatchet::test_no_file_exceeds_its_peg`
- `tests/quality/test_deferred_import_pegs.py::TestPegTablesLoad::test_committed_pegs_have_no_live_drift`

Reported drift: `(('src/teatree/core/intake/budget.py', 4, 0), ('src/teatree/core/models/implemented_issue_marker.py', 3, 1))` — i.e. observed 4 against a peg of 0, and observed 3 against a peg of 1.

Neither is a defect in the change; both edges look load-bearing. `tests/quality/deferred_import_pegs.toml` under `[intra_core]` needs:

- `"src/teatree/core/intake/budget.py" = 4` — a new entry. All four are ORM model reads
  (`ImplementedIssueMarker`, `Ticket`, `Task`, `PullRequest`) that can only resolve at call time,
  since this module is reached from the scanner factory, which loads before the app registry is
  populated. The `TYPE_CHECKING` pair at the top of the file is not counted.
- `"src/teatree/core/models/implemented_issue_marker.py" = 3` — was `1`. The two additions are the
  release-on-merge rule's `PullRequest` reads. It is a peer model in the same package, so they stay
  call-time for the same reason the existing `Ticket` deferral there already documents.

The peg file asks for the rationale to live in the commit message, so both entries want a comment
plus a commit note rather than a bare number bump.

Verified locally against this head: with only those two peg values changed, both test files pass
(20 passed). No source change is needed.

Issue #3842, opened 2026-07-28
Protect shipped loops/presets/schedules by detecting silent inertness, not by forbidding deletion


souliane:
## The question

Owner asked whether loops, presets and schedules that ship by default should be made impossible to delete.

## What the data says

Measured on the live box:

```
shipped in defaults.toml : 27 loops
present in the DB        : 27
missing from the DB      : 0
disabled in the DB       : 10
```

Disabled: `backlog_sweep`, `ci_eval_heal`, `dm_sweep`, `dogfood`, `eval_local`, `issue_disposition`, `memory_skim`, `outer_loop`, `snapshot_warmer`, `triage_assessor`.

**Nothing has ever been deleted.** Ten of twenty-seven sit present and inert.

Deletion is also already self-healing: the seed is `get_or_create` by name, so `t3 setup` recreates a removed shipped row. `defaults.toml`'s own header states this. A deleted shipped loop is therefore the *recoverable* failure.

Every failure that actually cost time recently was present-but-inert, never missing:

- `t3 <overlay> ticket sync-completions` exists and is wired to no loop, so the dashboard has a janitor that never runs (#3841)
- `review` is enabled, `colleague_facing = 1`, and last ran 6 hours ago against a 300s cadence
- `directive_loop` is held dark by an override
- `availability_override.json` sat stale for a week rather than absent (#3826)
- a `tickets` tick reports `58 signal(s), 58 action(s)` while no ticket advances (#3840)

An undeletable flag would have prevented none of these.

### 1. Drift report against the shipped seed set

One check answering three questions in one place:

- which shipped definitions are **missing** from the DB,
- which are **disabled**,
- which are **enabled but have not ticked** in N× their cadence.

That covers deletion, disablement and starvation together. `t3 worker status` already reports the third for loops — this generalises it, sources the expected set from the shipped seed rather than from the DB, and extends it to presets and schedules.

Sourcing from the seed is the important part: a check that reads the DB for both "expected" and "actual" cannot detect a missing row, which is the same self-referential defect as a golden compared against its own renderer (#3836).

### 2. Soft-protect deletion with a typed confirm, rather than forbidding it

Deleting a shipped definition should require a typed phrase naming what will stop happening — the pattern already exists (`stop-the-fleet` for the fleet kill-switch, and presets already refuse deletion while a schedule slot or override still names them).

Preferred over absolute prohibition because:

- it keeps a legitimate removal possible without a code change,
[...]

Issue #3987, opened 2026-08-01
GitHub create_pr fails permanently when the no-orphan push-gate already opened a PR for the branch


souliane:
## What happens

`ShipExecutor.run()` calls `GitHubCodeHost.create_pr()`, which always runs `gh pr create`. When the no-orphan push-gate hook (`ensure-pr`) has already opened a PR for the branch — which happens on the branch's first push, before the ticket's own ship step runs — `gh pr create` exits 1 with:

```
a pull request for branch "<branch>" into branch "main" already exists:
https://github.com/<owner>/<repo>/pull/<n>
```

`create_pr()` has no handling for this and raises `CommandFailedError`. `execute_ship` reports `ok: False` and the ticket never advances past `SHIPPED` — permanently, since retrying hits the identical error every time.

## Why it happens

Any branch that gets pushed more than once before `t3 <overlay> pr create <ticket_id>` runs will trip this: the no-orphan invariant's whole point is to open a PR proactively, and `ShipExecutor` has no path to discover and adopt a PR that a different actor already created.

## Discovered while

Shipping ticket #3842 — its branch had been pushed once (to run local verification), the no-orphan hook opened PR #3986 for it, and the subsequent `t3 teatree pr create 521 --sync` failed with exactly this error. Worked around by hand-editing the PR via `gh pr edit` for that one ticket; the underlying gap in `create_pr()` remains for every future ticket that hits the same ordering.

## Fix (already written, orphaned)

A fix exists — `GitHubCodeHost.create_pr` (split into `teatree.backends.github.pr_create.create_pr` to stay under the module-health LOC cap) catches the exact "already exists" `CommandFailedError`, then calls the existing fail-closed `teatree.core.forge_pr_probe.find_open_pr_for_branch` probe to independently CONFIRM an open PR exists before adopting its URL — a non-"already exists" failure, or an unconfirmable probe, still raises. Full test coverage (9 tests, 100% line coverage on the new module) already written and green.

It was committed on the `3842-...` ticket branch, but that branch's PR (#3986) squash-merged before the fix's own commit could ship — the branch is now orphaned relative to `main`. Needs to land via its own ticket/PR off current `main`.

## Scope note

Scoped to the GitHub client only. The GitLab client raises a different error shape (a REST 409 via `post_json`, not `CommandFailedError`), so parity for GitLab is a separate follow-up.

souliane:
[...]

PR #3989, opened 2026-08-01
feat(loops): report shipped-seed inertness; typed confirm on shipped delete (#3842)

souliane:
feat(loops): report shipped-seed inertness; typed confirm on shipped delete (#3842)

## What
Measured on the live box: 27 shipped loops, 27 present, 0 missing, 10 disabled.
Nothing has ever been deleted — every failure that cost time was present-but-inert.
An undeletable flag would have prevented none of them.

Detection (the half that carries the value):

- `teatree.loops.seed_inertness.shipped_inertness()` names every shipped
  loop/preset/schedule that is missing, disabled, or not ticking. The EXPECTED set is
  read from the shipped seed tables, never the DB — a check that reads the DB for both
  sides cannot see a deleted row (the #3836 self-referential defect).
- Severity splits on the shipped `default_enabled`, which is what sourcing from the seed
  buys: a loop that ships ON and is off REGRESSED (fault); one that ships off and is off
  is a note. Without the split the live box's 10 dark loops are 10 lines nobody reads.
- Surfaced by `t3 loops audit [--json]` (non-zero on any fault) and a crash-proof
  `t3 doctor check` check that echoes faults only.
- Reuses `loop_staleness.stale_loops` for the 3x-cadence reading, including its
  suppressed/unexplained split — no second copy of the cadence arithmetic.

Soft protection (rather than forbidding deletion):

- `shipped_guard.require_shipped_delete_confirm` demands a typed `stop-<name>`. The
  refusal quotes the shipped description, because the operator deleting a shipped loop is
  usually unclear on what it does. Same doctrine as the fleet stop switch.
- Enforced at `delete_preset` (existing) plus new `delete_loop` / `delete_schedule` seams
  — neither existed, so the Django admin button was the only way to remove either.
- The admin routes a SHIPPED row to those seams instead of deleting in place; a
  non-shipped row keeps the ordinary admin delete, and Django's own permission check is
  still consulted first.
- The confirm never overrides an integrity refusal: a preset a slot still names, and the
  active calendar, are refused whatever phrase is typed.

Also fixes: `manage.py loop_preset delete` deleted the `Mode` row directly, bypassing the
referrer refusal the dashboard enforces. It now routes through the shared seam.

Open questions & assumptions

- `loop_preset delete` bypassing the referrer refusal was pre-existing. Rerouting it
  through `delete_preset()` is a behaviour change that STRENGTHENS the guard (assumed —
  the divergence between the dash and CLI paths reads as an oversight, not a design).
- Dropped the planned preset `unreachable` finding: 4 of 7 shipped presets are named by
  no slot/override/setting on a fresh install and are reached by hand, so reporting it
  made the audit noisy out of the box (decided by the fresh-seed acceptance test).
- The audit's expected set covers loops, presets and schedules; `t3 loops audit` is a
  slight misnomer for a report spanning three families, kept because `teatree/loops/`
  owns all three seed modules and it is where an operator looks (assumed).
- `is_shipped`/`shipped_description` live in `teatree.config.seed_defaults`, not beside
  the delete policy, because the Django admin (domain layer) consumes them and may not
  import `teatree.loops` (decided by tach).
- The management command raises `SystemExit(1)`, not `typer.Exit`, per the house rule —
  pinned by a test, since `call_command` swallows `typer.Exit` and CI would read green.
- Schedule `empty` is a fault only for the ACTIVE calendar; an empty inactive one is a
  note (assumed — an inactive calendar governs nothing).

## Why


souliane:
Closing as a duplicate of #3988 — byte-identical content, verified by tree SHA (both heads resolve to the same tree `82fc86c8`).

The two differ only in where they live. #3988 is on a branch named for the issue it fixes and carries a matching title. This one sits on the `3842-...` branch and inherited that ticket's title ("report shipped-seed inertness; typed confirm on shipped delete"), while its actual diff is the GitHub `create_pr` adoption fix — the #3842 work itself already landed on main via #3986, so the title here describes something this PR does not contain.

Keeping #3988 so the branch, title and content agree. Deliberately NOT deleting this branch: the content is identical and reachable from #3988, and once that merges the landed-content probe should discharge any owed-PR obligation for this branch on its own rather than needing a manual decision.

Worth noting for #3987 that a work-on-the-wrong-branch outcome like this is a plausible neighbour of the PR-creation retry behaviour being fixed there.

Issue #3617, opened 2026-07-22
t3 worker ensure spawns a duplicate worker (pid-file blindness) → task lease thrash


souliane:
`t3 worker status`/`t3 worker ensure` decide 'running?' from the pid file, which reports NOT running when a worker actually holds the singleton flock (stale/missing pid file). So `t3 worker ensure` spawns a SECOND worker next to the live one; both drain the task queue and thrash leases (`stuck_loop: lease lost for task N: re-claimed by another worker`), failing every coding/planning task.

Fix: gate `ensure` (and the `status` running-check) on the kernel flock (`flock_is_held`), not the pid file, so a live-but-pid-blind worker is detected and `ensure` is a true no-op. Relates to the #3602 flock-fallback work; extend it to the ensure/spawn path. Add a test: a live flock-holder + stale pid file → `ensure` spawns nothing.

PR #4023, opened 2026-08-03
fix(worker,doctor): report a singleton held from outside the deployment (#3976)

### LINKED ISSUE #3976: Worker singleton held from outside the container starves the deployed worker, silently and indefinitely
## What

A `t3 worker` process started **outside the container** holds the worker singleton, and the
containerized `teatree-worker` service can then never start. It boots, refuses with

```
worker already running (PID <n>) — see .../teatree/worker.pid
```

and exits — restarting on that cycle indefinitely.

## Why it survives a look at the box

Everything reports healthy, which is what makes this persist:

- `t3 loops list` shows every loop ticking on cadence with recent `last` times — because the
  out-of-container worker really is driving them;
- `t3 worker status` reports `worker: RUNNING`, since the flock genuinely is held;
- the service is `Up`, because it had just restarted when looked at.

The only place the truth appears is the service's **stderr**, and only for the ~2 seconds
between one boot and the next exit. The restart counter climbing is the sole durable signal,
and nothing surfaces it. Observed at a restart count in the dozens while the deployment
looked entirely healthy.

The pid file and lock live on a path shared between the host and the container, so the two
workers contend for one singleton across a boundary neither side reports.

## Impact

Two things, neither of which shows up as a failure:

- the sanctioned worker never runs, so the deployment's actual behaviour is whatever the
  out-of-container process does — a different code revision, a different environment, and
  outside the deploy pipeline entirely;
- each refused boot re-runs the full provisioning pass first (repo housekeeping across every
  registered worktree, skill provisioning, network fetches) and then throws it away, several
  times a minute, indefinitely.

## Suggested direction

The singleton behaves correctly — the gap is that losing the race is silent and unbounded.

- The refusal should say **where** the holder is, not just its pid: a pid that cannot be
  resolved in the refusing process's own context is a *different* condition from a second
  copy of the same service, and the two need opposite responses. This is the same
  view-mismatch question as souliane/teatree#3952 and the worktree-reaping report.
- A service that refuses to start for the same reason N times running should escalate rather
  than restart forever — a restart loop that can never succeed is a hard failure wearing a
  healthy face.
- `t3 doctor check` should fail when the singleton holder is not the service that is supposed
  to hold it. Today no check compares those two facts, so the split is invisible.

## Acceptance

- `t3 doctor check` fails, naming the holder, when the worker singleton is held from outside
  the deployment.
- The refusal message distinguishes "another copy of this service holds it" from "the holder
  is not resolvable here".
- Repeated identical refusals surface as a failure instead of restarting silently.
- Proven RED: with the check removed, a singleton held from outside is reported healthy and
  the test fails.

souliane:
Closes [#3976](https://github.com/souliane/teatree/issues/3976).

## What

A `t3 worker` started outside the container holds the worker singleton over a lock file the host and the deployment share, so the containerized service can never start — it boots, refuses, and restarts on that cycle indefinitely while every surface reads healthy: the loops DO tick (the outside holder drives them), `t3 worker status` DOES report RUNNING (the flock genuinely is held), and the service IS Up because it had just restarted. The only signal was ~2s of stderr per boot and a restart counter nobody watches.

- **singleton** — the acquirer records its `ExecutionContext` (pid namespace, hostname, `TEATREE_ROLE`) as a second line beside the pid, so a refusal distinguishes "another copy of this service holds it" from "the holder is NOT resolvable here". Line 1 stays the bare pid; `read_pid` parses first-line, so a legacy one-line lock file and every existing consumer are unchanged.
- **`singleton_refusals`** (new) — a durable streak ledger keyed on the refusal REASON. The holder pid is deliberately excluded, so a foreign holder restarting does not reset the streak; a restart in a fresh container reads what its predecessor wrote, and a successful acquire clears it.
- **worker command** — records each refusal and, at `ESCALATION_THRESHOLD` (3) identical refusals in a row, exits `STARVED_EXIT` (5) with a loud headline instead of churning silently.
- **doctor** — `_check_worker_singleton_holder` HARD-FAILs on a standing streak, and — where a deployment defines a sanctioned holder — whenever the singleton is held by anything other than its own `worker`-role service, naming the holder. Crash-proof: a failed probe degrades to WARN.
- **`t3 worker status`** — names where the holder is, and points at `t3 doctor check` when it is not this deployment's worker.

## Acceptance

| Criterion | Where |
|---|---|
| `t3 doctor check` fails, naming the holder, when the singleton is held from outside the deployment | `_check_worker_singleton_holder`, `tests/teatree_cli/doctor/test_worker_singleton_holder_check.py::TestHolderAttribution` |
| The refusal distinguishes "another copy of this service" from "not resolvable here" | `singleton._verdict_clause`, `tests/teatree_utils/test_singleton.py` |
| Repeated identical refusals surface as a failure instead of restarting silently | `Command._report_refusal` + the doctor streak FAIL, `tests/teatree_core/management/test_worker_command.py` |
| Proven RED | see below |

## Verification

- Ticket's own tests: `110 passed` (`tests/teatree_cli/doctor/test_worker_singleton_holder_check.py`, `tests/teatree_utils/test_singleton.py`, `tests/teatree_utils/test_singleton_refusals.py`, `tests/teatree_cli/test_cli_worker.py`, `tests/teatree_core/management/test_worker_command.py`).
- **Proven RED (re-verified this cycle):** neutering `_holder_findings` to return `[]` turns 6 of the 10 doctor tests red — a singleton held from outside is reported healthy. Restored, all green.
- `uv run tach check` — all modules validated (no backwards edge into `teatree.docker`).
- Whole suite on this box: `33478 passed, 12 failed`. All 12 failures are pre-existing ambient-environment leaks unrelated to this diff, each isolated: 7 in `tests/cli_doctor/test_doctor_check_command.py` fail identically with the new check stubbed out (root cause: a shadowing `t3` on `PATH` plus a DB-blocked config read), and the other 5 are ambient-credential leaks — e.g. `test_code_host_from_overlay_returns_none_when_no_token` passes under `env -u TEATREE_GH_TOKEN`.

## Architecture pre-check — souliane/teatree#3976

**1. BLUEPRINT § alignment.** §5.6 Loop Topology (the singleton `t3 worker` owns the tick cadence). Purely additive — no BLUEPRINT statement is contradicted; `docs/blueprint/loop-topology.md` gains the paragraph describing the new reporting.

**2. FSM phase boundaries.** n/a — no `Ticket.State` / `Worktree.State` transition touched.

**3. Extension-point contracts.** No `OverlayBase` hook, scanner registration, or `*Backend` Protocol change. One doctor-check registration is added (an internal registry). Every consumer of the changed `teatree.utils.singleton` surface was re-read: `cli/worker.py`, `cli/overlay.py`, `cli/slack/listen.py`, `cli/doctor/{checks_runtime,checks_slack_roundtrip,self_heal}.py`, `core/runners/service_launch.py`, `core/management/commands/worker.py`, `loop/{driver_detection,queue_drain,worker_lifecycle}.py`, `loops/live.py`, `core/loop_lease_liveness.py`, `eval/regression_corpus_fixtures.py`.

**4. Component boundaries.** The holder RECORD belongs in `utils/singleton.py` — only the acquirer, under the lock, can write it truthfully. The streak ledger is its own module (`utils/singleton_refusals.py`): a restart-loop concern, not a lock concern, keeping the primitive side-effect-free for the four other singletons that have no restart loop. The worker command owns record/clear/escalate; `cli/doctor/checks_runtime.py` owns the health surface.

**5. Dependency direction.** `teatree.utils` is `foundation`; the new module imports only `teatree.paths` + stdlib — no new edge. Deliberately does NOT call `teatree.docker.workflow.is_running_in_container` (that edge is backwards under tach); the container-vs-host fact is recorded data. `uv run tach check` ✅.

**6. Test surface.** Holder record written under the lock / legacy one-line file still parses / same-context vs foreign message / streak increments and resets / worker exit codes + ledger clear / doctor FAIL on foreign holder, FAIL on standing streak, PASS for the sanctioned holder — each pinned in the five test files above.

**7. Resilience invariants (#1192).** verify-by-re-read: the record is written under the flock and re-read live by every probe. fallback-transport: a missing/garbled record degrades to an explicit UNKNOWN reported as such, never silently to "fine". idempotency: `record_refusal` is keyed on the reason fingerprint (increments, never appends); `clear_refusals` on a missing file is a no-op. heartbeat / sub-agent contract: n/a (no long-running work, no dispatch).

**8. Identity and key normalization.** The canonical key for "does this pid mean anything here" is the PID-namespace link — the kernel's own answer, immune to hostname collision. Hostname and role are descriptive only. One producer (`current_context`), one comparator (`holder_verdict`), no string stripping. The *sanctioned-holder* question uses a deliberately different key, `TEATREE_ROLE`: sibling containers have different namespaces but are all deployment services, so role — not namespace — is what doctor compares. Both keys compared whole.

**9. Behavior preservation / capability deletion.** Additive; nothing replaced or narrowed. `read_pid` gains first-line parsing and the legacy single-line format still parses (own test). `flock_is_held`, `singleton`'s reuse-in-place / never-unlink contract (#3617), and `AlreadyRunningError`'s positional signature and leading message text are unchanged. No privacy/leak/security matcher touched; no must-block test inverted.

**10. Removability / harness-vs-data.** Removable: unregister 2 lines in `app.py`, delete the ledger module whole, and the record's second line becomes ignorable (`read_pid` already tolerates it). Harness-vs-data: the facts live in data (two files under `DATA_DIR`), the rules in harness (a verdict enum and one threshold). The threshold is a single global with no per-item variation, so it stays a documented constant rather than a `ConfigSetting` with no varying dimension.

- **assumed** — the escalation threshold is 3 consecutive identical refusals. A drain-then-deploy hand-over can legitimately lose the race once or twice while the outgoing worker releases the flock; three identical ones cannot.
- **assumed** — a holder that recorded no execution context is a FAIL inside a containerized deployment, not a WARN. Failing open there re-creates the exact silent-healthy state this fixes; the cost is a transient FAIL during the one rollout that upgrades a host-held worker.
- **assumed** — `restart: unless-stopped` restarts the worker regardless of exit code, so "escalate rather than restart forever" is delivered as a durable, doctor-visible failure plus a distinct exit code, not an in-process refusal to exit.
- **open** — the issue also notes each refused boot re-runs the full provisioning pass before refusing. Not addressed here: it needs a preflight in `deploy/entrypoint.sh` ahead of provisioning, which is outside this scope. The escalation makes the waste visible rather than ending it.
- **decided-by-code** — the container-vs-host discriminator is recorded data, not a call into `teatree.docker.workflow.is_running_in_container`, because `teatree.docker` depends on `teatree.utils` and that edge is backwards under tach.


souliane:
## Test evidence — ticket [#3976](https://github.com/souliane/teatree/issues/3976)

### CI (authoritative)

All required checks on this head (`71b2dda6`) are green: **40 SUCCESS, 3 SKIPPED, 0 pending, 0 failed** — including both `test-shard` and all 4 `test-shuffle` (order-dependence) jobs, `preflight`, `blueprint-cross-pr`, and `doc-update-gate`.

### Local

- **Ticket's own tests** (110 tests across the 5 touched/added test files): all green.
- **`bash dev/test-affected.sh`** (diff-scoped lane, escalated to whole-tree per its own policy): `33554 passed, 12 failed, 63 skipped, 5 xfailed`. The 12 failures are the same set observed pre-diff and are pre-existing ambient-environment leaks unrelated to this change — reproduced with the new doctor check stubbed out (7 of the 12, in `tests/cli_doctor/test_doctor_check_command.py`, fail identically) and reproduced on `main` (the other 5 are `TEATREE_GH_TOKEN`/network-credential leaks, e.g. `test_code_host_from_overlay_returns_none_when_no_token` passes under `env -u TEATREE_GH_TOKEN`).
- **`uv run tach check`**: all modules validated — no backwards edge into `teatree.docker`.
- **Manual functional smoke test** against this box's own live deployment (read-only, no singleton contention): `t3 worker status` reports `RUNNING` with no holder-mismatch line (this container's worker is the sanctioned holder — correct), and `t3 doctor check` runs the new `_check_worker_singleton_holder` check clean (no FAIL from it) in the live environment.
- **Proven RED** (re-verified this cycle): neutering `_holder_findings` → `[]` turns 6 of the 10 doctor tests in `tests/teatree_cli/doctor/test_worker_singleton_holder_check.py` red — a singleton held from outside the deployment is reported healthy. Restored, all green.

### Acceptance criteria (from the issue)

| Criterion | Status |
|---|---|
| `t3 doctor check` fails, naming the holder, when the singleton is held from outside the deployment | ✅ `TestHolderAttribution` |
| The refusal distinguishes "another copy of this service" from "not resolvable here" | ✅ `singleton._verdict_clause` + tests |
| Repeated identical refusals surface as a failure instead of restarting silently | ✅ `Command._report_refusal` + `TestRefusalStreak` |
| Proven RED | ✅ see above |

Issue #4016, opened 2026-08-02
A completed coding run is discarded when a worktree probe is unverifiable (landing gate crashes instead of failing open)


souliane:
## Symptom

A **completed** headless `coding`/`debugging` run is recorded as `crash` (exit 1) and its result envelope is discarded, whenever any of the ticket's registered worktrees is present on disk but its commit-count probe cannot be verified.

Observed on one ticket three times in a row (three separate dispatches, each burning a full agent run). Every attempt died at the same frame — *after* the agent finished:

```
File ".../agents/attempt_recorder.py", line 166, in record_result_envelope
    landing_error = landing_verification_error(task, phase=phase)
File ".../agents/landing_verification.py", line 49, in landing_verification_error
    if not any(worktree_has_commits_ahead(wt) for wt in checkable):
File ".../core/models/ticket_worktree_checks.py", line 98, in worktree_has_commits_ahead
    raise WorktreeProbeUnverifiableError(msg) from exc
WorktreeProbeUnverifiableError: could not verify commits ahead for <worktree> (main..<branch>)
```

## Root cause

`WorktreeProbeUnverifiableError` is raised in exactly one place and **caught in zero places** across `src/`:

```
$ rg -n "WorktreeProbeUnverifiableError" src/
core/models/ticket_worktree_checks.py:25:  class WorktreeProbeUnverifiableError(RuntimeError):
core/models/ticket_worktree_checks.py:98:      raise WorktreeProbeUnverifiableError(msg) from exc
```

The typed error was introduced so a caller could tell "could not verify" from "verified: nothing to ship" and **hold the tick** rather than dispose. That intent is right, and its own docstring says so. But no consumer was ever taught to handle it, so on the two `agents/` call sites it does not hold anything — it crashes the recorder.

The two `agents/` call sites are structurally unable to "hold the tick": the agent has already finished and its envelope is in hand. Crashing there does the exact opposite of holding — the run is recorded as a crash and the work is thrown away.

`landing_verification.py`'s own module docstring already states the correct policy and the code contradicts it:

> Fail-open on the unverifiable: ... "couldn't determine" is never "did not land".

## Why the probe is unverifiable here (not transient)

[...]

Issue #3921, opened 2026-07-30
Audit: one merge decision spread across 15 tables, 9 with no reconciler, and CI taxes deletion


souliane:
## Why

The owner's read is that this project has become a patchwork of unplanned workarounds, producing more bugs than features. Today's outage supported it: a merge deadlock caused by three independent mechanisms each blocking the same merge, none aware of the others. This audit set out to confirm or refute that with evidence rather than assert it.

It is confirmed, and the mechanism is more specific — and more fixable — than "sloppiness".

## Headline

The question **"has this PR been reviewed, and may it merge?"** is answered by **15 durable tables plus one JSON blob, 6 independent reviewing-task producers, 6 competing dedup mechanisms, 2 parallel decision ladders, and 13 merge-time assertions**. No single component can see the whole answer.

- **9 of 11 stateful mechanisms have no reconciler**, or one slower than the damage.
- **Exactly one TTL exists on the entire path** (`MRReviewLock`, 2h), and its reconciler `reconcile_stale()` has one production caller — inside the merge gate the stale lock is blocking.
- **11 overlapping pairs**, 9 of which can produce a live production disagreement.

## The deletion tax — why it only ever grows

This is the finding that explains the rest. **Four quality gates fail on improvement**: `test_no_flat_core_regrowth.py:165` asserts `==` rather than `<=`, and there is a `test_no_file_is_under_its_peg`. Removing a mechanism turns CI red until a pin is hand-edited downward.

Three files are at **exactly zero slack** today: `overlay.py` at exactly 500 LOC, flat core leaves at exactly 103, `hook_router.py` at exactly 4289. Addition is cheap; deletion is a build failure. The comment block at `test_no_flat_core_regrowth.py:105` records that the pin was raised to 103 by **#3910 — the outage ticket itself**.

## The no-retirement hypothesis: confirmed, 0 of 17

Not one of 17 mechanisms removed or narrowed an existing one when it was created. One model file has ever been deleted repo-wide (off this path); **zero gate files ever** — `core/gates/` went 0 to 47 in 54 days, monotonic at every sample.

The qualification makes it worse: guards are sometimes retired — **in prose and behaviour only, while the table, column, enum member and code path remain**. Five such cases are live. A dead mechanism that still takes locks is more dangerous than one nobody noticed.

## Two live bugs falling out of the map

[...]

PR #4030, opened 2026-08-03
refactor(core,review): retire the dead review state and unify the claim predicate (#3921)

souliane:
## What

Recovered work for #3921 that was completed and then discarded: retire the dead review state,
route the claim predicates through one shared rule, give the dispatch claims a deadline and a
bounded retry, and make four improvement-taxing ratchets one-sided.

54 files, +1533/-1143 measured from the merge base.

## Why this PR exists separately

The coding run finished and committed these 8 commits, but the worktree probe that verifies a
landing could not answer, so the run was treated as having produced nothing and the ticket rolled
back. The commits stayed on a local branch with no remote ref and no PR — reachable only from one
machine's disk and invisible to every sweep that drains unfinished work.

That is a live reproduction of #4016 (a completed coding run discarded when the worktree probe is
unverifiable). Filing this PR both preserves the work and gives that issue a concrete case.

## State

The branch is behind `main` and needs it merged before this is mergeable. Reviewing that merge
carefully matters here: a previous branch in this family had its own fixes silently reverted by a
merge that produced no conflict, because the reverted lines were ones the merge had no reason to
flag. Verify after merging that the changes described above are still present, rather than
trusting a clean merge result.


souliane:
Closing as already-shipped, not merging — every commit's content is already in `main`, and merging this branch would be actively harmful.

Verified against `origin/main` symbol by symbol:

| Branch change | State in main |
|---|---|
| retire dead review state | `review_loop.py` GONE, `LEG_REVIEWER` GONE |
| shared claim predicate | `acquirable_q` in `expiring_claim.py`, `auto_review_dispatch.py`, `codex_review_marker.py` |
| unified `_ACTIVE_STATES` | present across the claim models |
| migrations 0042 / 0043 / 0044 | all present; main's leaf is 0053 |
| `test_review_loop.py` deletion | already gone from main |
| BLUEPRINT expiry-acquirable note | already recorded |

`git cherry -v origin/main HEAD` reports all 8 commits as `+` (not upstream). That is a **patch-id false positive** — main acquired the same content through different commits, so the patch-ids differ while the effect is identical. Worth noting for anyone else auditing this family: `git cherry` is not sufficient evidence here, symbol-level checks are.

**Why merging would be wrong, not merely redundant.** This branch is 68 commits behind main. Its 10-file conflict set includes `max_migration.txt` and two models that main has since moved on. Resolving those conflicts in the branch's favour would revert main's newer work — which is exactly how a sibling branch in this family had its fixes silently reverted by a conflict-free merge. The stale branch is the hazard, not the merge direction.

The recovered work this PR existed to preserve is safe: it is in main already.

Issue #4035, opened 2026-08-03
Doctor-aggregate test times out under shard contention (20s local vs >60s ceiling), reddening unrelated PRs


souliane:
`tests/teatree_core/test_schema_guard.py::BehindSelfDbReportingTest::test_doctor_check_command_aggregates_pending_migrations`
exceeds the 60s `pytest-timeout` ceiling under shard parallelism, on PRs whose diff cannot affect
it. Observed twice on the same PR, including after an explicit re-run, so it is not a one-off.

Measured: **20.15s locally** on an idle tree, passing. Under the 12-way shard matrix it exceeds
60s — roughly 3x, consistent with contention rather than with the test doing more work.

The test drives the full `t3 doctor check` aggregate, which is the slowest command in the CLI and
has grown as checks were added. A 20s test against a 60s ceiling has only 3x of headroom, and the
sharded lane routinely consumes that.

Why this matters beyond the red check: the failure names a test the PR author never touched, so
the correct response ("not mine") is indistinguishable from the tempting one ("re-run until
green"). Whichever an agent picks, nothing improves, and the next PR pays the same tax. An
autonomous loop that learns to re-run until green has learned to ignore its own gates — see the
verification half of #4033.

Options, roughly in order of preference:

1. Make the aggregate cheaper for this test — it needs the aggregation behaviour, not every
   check's real work.
2. Split the assertion so the slow aggregate runs once rather than per case.
3. Raise the ceiling for this node only, with the measurement recorded, so the headroom is a
   stated decision rather than an accident.

Raising the global timeout is the option to avoid: it hides every other test drifting toward the
same edge.

souliane:
Wider than one test. A third distinct case hit the same 60s ceiling today:

- `tests/teatree_core/test_schema_guard.py::BehindSelfDbReportingTest::test_doctor_check_command_aggregates_pending_migrations` (the original; PR #4037 scopes it)
- `tests/teatree_cli/doctor/test_self_heal.py::DoctorJsonSurfaceTest::test_check_without_json_does_not_route_to_json`
- `tests/teatree_loop/test_tick.py::TestBuildDefaultJobsWiring::test_tick_multi_overlay_prefixes_summary` (in the **shuffle** lane, on an unrelated PR)

Three unrelated tests, three unrelated PRs, one ceiling — so this is not a slow test, it is a ceiling that no longer matches what the sharded and shuffled lanes actually cost. Scoping each one as it fires treats the symptom and guarantees a fourth.

[...]

PR #4037, opened 2026-08-03
fix(tests): give the doctor-aggregate case the same scoped timeout as its siblings

souliane:
## What

`BehindSelfDbReportingTest` gets the scoped `@pytest.mark.timeout(240)` its two sibling classes in
the same file already carry.

## Why

All three classes drive real migrations — this one unapplies an initial migration in `setUp` and
reapplies it on cleanup. The other two were given a scoped ceiling for exactly that reason
(#1189), with a comment explaining that the work exceeds the global 60s under parallel contention.
This class was the one left out, which put it one contention spike away from red.

Measured 20s single-core; the 12-way shard matrix stretches that past 60s. It failed three
consecutive runs of an unrelated PR, naming a test that PR never touched.

That failure shape is the reason this is worth fixing rather than re-running: "this isn't my
change" and "re-run until it passes" produce the same action, so neither the author nor an
automated loop learns anything, and the next PR pays the same tax. Tracked as #4035.

## Why scoped rather than global

Raising the global ceiling would clear this and hide every other test drifting toward the same
edge. The 60s hang-detector is doing real work everywhere else; this is one class that genuinely
needs more, matching the precedent already set beside it.

### DIFF
--- a/tests/teatree_core/test_schema_guard.py
+++ b/tests/teatree_core/test_schema_guard.py
@@ -194,6 +194,12 @@ class TestSchemaGuardOnPrivateAlias:
         assert MergeClear.objects.using(alias).count() == 0  # healed: the table is now usable
 
 
+# `setUp` unapplies an initial migration and the cleanup reapplies it, so this class
+# pays the same real-migration cost as its self-healing sibling below and needs the same
+# headroom: measured 20s single-core, which the 12-way shard matrix stretches past the
+# global 60s ceiling. Scoped like the sibling (#1189) so the 60s hang-detector keeps its
+# meaning everywhere else — a global raise would hide the next test drifting to the edge.
+@pytest.mark.timeout(240)
 class BehindSelfDbReportingTest(TransactionTestCase):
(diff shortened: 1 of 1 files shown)


souliane:
Superseded by #4059, merged at 11:40.

This branch and #4059 independently reached the identical fix — `@pytest.mark.timeout(240)` on `BehindSelfDbReportingTest`, scoped rather than a global raise, for the same #1189 reason. That agreement is worth recording: two separate analyses concluded the cost here is irreducible (the `setUp` migration unapply/reapply on the shared `default` connection cannot be stubbed without removing the precondition under test).

`main` already carries that marker, which is why this branch now conflicts on exactly that line — the change is a no-op against current main.

#4059 also carried the second, unrelated cause (`test_check_without_json_does_not_route_to_json` running the whole real doctor aggregate for a pure routing assertion) plus the `.venv-hook` scan fix from #4031, which is what let it go green standalone and break the deadlock the three PRs were in.

Issue #4052, opened 2026-08-03
Forge portability: gate and shrink the raw-GitHub leak past CodeHostBackend


souliane:
## Problem

Teatree has a `CodeHostBackend` protocol and both `backends/github` and `backends/gitlab` implementations, but core code routinely bypasses the seam: a raw grep finds on the order of 114 `gh`-shaped call sites in `src/teatree/` outside `backends/github`, and at least 8 modules reference GitHub directly outside the backend layer (`url_classify.py`, `url_title_fetcher.py`, `publish_destination.py`, `publish_surface.py`, `_publish_detection.py`, `_repo_visibility.py`, `hooks/_python_rest_detection.py`, `cli/codex.py`). Nothing exercises the non-GitHub path in normal operation, so it rots silently — the exact failure mode already observed once in #3794 (the review CLI misreporting a `glab` login problem because the GitLab path was rarely run). This was captured verbatim as owner directive #43: "teatree must work outside github too." No PR has yet touched this; it is unaddressed on `main`.

The raw counts above are unverified greps, not a triaged finding — some hits may be comments, docstrings, or legitimately backend-internal. The work here is to turn that into a real, shrinking number with a gate behind it, not to trust the number as-is.

## Design

Three parts, in dependency order:

1. **Triage the leak, don't just count it.** Walk the ~114 call sites and the 8+ modules and classify each as either legitimately forge-specific (belongs inside a `backends/*` implementation) or leaked core logic (must be routed through `CodeHostBackend` instead). Only the former is allowed to name a forge.
2. **Fix at least one real leak** from the identified list — route it through `CodeHostBackend` rather than a direct `gh`/GitHub call — proving the pattern works before it's declared a rule.
3. **Add a shrink-only ratchet** on raw forge-shaped call sites in core paths, in the same family as the existing flat-core and deferred-import ratchets: it records a baseline count, fails if a new commit raises it, and is proven RED by deliberately adding a violating call site in a test fixture.
4. **Exercise a real non-GitHub path in tests/CI**, so a break in the GitLab backend is caught by a run, not discovered by an operator months later the way #3794 was. This does not have to cover everything — one representative end-to-end flow (e.g. publish detection or the review CLI) run against the GitLab backend is enough to close the "nothing proves this" gap.

[...]

Issue #4041, opened 2026-08-03
Checks that cannot verify are silently treated as checks that passed


souliane:
## Problem

Several checks in teatree can end up in a state where they genuinely cannot tell whether something is true, false, or not applicable — the dependency they'd probe isn't reachable, the repo path can't be resolved, the architecture moved on and the thing they were checking no longer exists, or the text heuristic they use just didn't match this case. In every one of these members, that "I don't know" state gets silently folded into the good branch: no violation, no work to do, exempt, still alive. The check doesn't lie exactly, it just never learned how to say "unverifiable," so degraded environments, blind spots, and stale assumptions all read the same as everything being fine — sometimes permanently, sometimes long enough for real work to get lost.

## Approach

Give each of these checks a real third answer instead of a boolean or a silent default. Concretely: reuse the tri-state pattern teatree already has for the cold-sqlite tier (`COLD_READ_OK` / `COLD_READ_UNREADABLE`) as the shape for a small "check result" convention — confirmed-good, confirmed-bad, or unverifiable-with-a-reason — and apply it at each of the seams below. The important part isn't the enum itself, it's the rule at every call site: an unverifiable result must be surfaced (logged, refused, reported, escalated), never quietly treated as a pass. That's one change in shape, applied seven times at seven different seams that all had the same gap.

## Acceptance

- **Hook interpreter / ORM reachability (was #3831).** Add a runtime probe (`_check_hook_interpreter_has_django()` in `checks_cold_hooks.py`, wired into `doctor/app.py` next to its cold-sqlite sibling) that actually shells out through `run-hook.sh` and checks whether `bootstrap_teatree_django()` succeeds, plus a conformance lane asserting every call site that branches on that function's return either records the degraded case or is on a reviewable allowlist.
  Test (RED today): on a host where the selected hook interpreter can read sqlite but can't `import django`, `t3 doctor check` stays green and none of the ~18 call sites record anything. GREEN: doctor reports the unreachable-ORM state, and the conformance lane fails if a new call site branches on the boolean without handling degradation.

- **Retired-subsystem eval guard (was #3850).** Extend the guard in `tests/conformance/test_retired_subsystem_evals.py` to also scan scenario body text (prompt/scenario/expect), not just the scenario name, for retired-subsystem words — or require each scenario to declare an explicit, validated `subsystem` tag instead of relying on name matching.
[...]

PR #4068, opened 2026-08-03
docs(skills): sweeping-tickets merges into an existing ticket, never a new epic (#4065)

### LINKED ISSUE #4065: /t3:sweeping-tickets mints new umbrella epics; it must reuse the best-fitting existing ticket and merge the others' content into it
Owner: *"I don't want umbrella tickets, instead you merge reuse existing one and merge the others'
content to it"*

## What the skill does now

`/t3:sweeping-tickets` is built around creating new tracking epics. Its own description says it
consolidates "by merging related tickets into a small set of tracking epics", and the disposition
table routes a consolidation to "the epic (existing or **newly proposed**) it now belongs to".

Applied to the 2026-08-03 sweep, that minted **twelve new epics (#4041–#4052)** and closed 48
issues into them as `not planned`. The tracker went 63 → 15, but 12 of the surviving 15 are rows
that did not exist before the sweep and that nobody had asked for.

## Why that is the wrong shape

An umbrella ticket is a new container with no history, no discussion and no prior context. Folding
N real tickets into one invents a row whose only content is a checklist — the original threads,
their evidence and their comments all become closed and secondary. The consolidation reads as
progress (63 → 15) while the substance moved into rows that are strictly newer and thinner than
what they replaced.

Reusing an existing ticket keeps the thread that already has the history, and merging the others'
content into it means the surviving row gets *richer*, not emptier.

## The change

1. **Never create a new epic/umbrella row during a sweep.** Remove "newly proposed" from the
   consolidation disposition; the target must be an existing open ticket.
2. **Pick the best-fitting existing ticket as the merge target** — the one whose scope most nearly
   contains the others (prefer the oldest / most-discussed when several fit, since that is where
   the history lives).
3. **Merge the others' content INTO it** — not a bare checklist line. The folded ticket's
   substance (its statement of the problem, its evidence, its acceptance criteria) is appended to
   the target so nothing thins out, then the standalone closes with a pointer.
4. If genuinely nothing fits, the ticket **stays standalone**. That is a legitimate outcome and
   counts toward the operator's cap; it is not a reason to invent a container.
5. Update the skill description, the doctrine section, the disposition table, and the CLI recipe
   block that greps for `label:epic` candidates.

## Note on the existing epics

This does not ask for #4041–#4052 to be unwound — that would be a second churn over the same
rows. It stops the next sweep from doing it again. Whether to collapse the twelve back into
pre-existing tickets is a separate call for the owner.

souliane:
Fixes #4065.

Owner: *"I don't want umbrella tickets, instead you merge reuse existing one and merge the others'
content to it"*

## What the skill did

`/t3:sweeping-tickets` was built the other way round. Its description promised consolidation "into
a small set of tracking epics"; the disposition table routed a fold to "the epic (existing or
**newly proposed**)"; step 5 was titled **"Propose the epic set"**.

Run against the tracker on 2026-08-03 it minted **twelve new epics (#4041–#4052)** and closed 48
issues into them. The count went 63 → 15, and twelve of the fifteen survivors were rows nobody had
asked for.

## Why an umbrella row is the wrong shape

A fresh epic is a container with no history, no discussion and no prior context. Folding N real
tickets into one invents a row strictly thinner than what it replaced, while the originals — with
their evidence and their threads — all go closed and secondary. The open count drops and the
substance degrades, which reads as progress and is not.

Reusing a ticket that already exists keeps the thread that holds the history, and merging the
others' content into it leaves the survivor **richer** than it was.

## The change

- **The merge target must already exist.** "Newly proposed" is removed from the disposition table;
  creating an epic/umbrella row during a sweep is now explicitly forbidden.
- **Target selection is defined**: the open ticket that already covers most of the candidate's
  scope, preferring the oldest / most-discussed on a tie, because that is where the history lives.
- **Merged content is substance, not a checklist line** — problem statement, evidence, acceptance
  criteria move across.
- **"Nothing fits" is a legitimate outcome.** The issue stays standalone and counts toward the
  operator's cap. If the cap cannot be met without minting rows, the skill now says so and lets the
  operator choose, rather than closing the gap itself by inventing containers.
- Step 5 rewritten from "Propose the epic set" to "Choose merge targets from the tickets that
  already exist"; the CLI recipe no longer greps `--label epic` for candidates but reads the open
  set; README skills catalogue regenerated by its pre-commit hook.

## Scope

Existing epics #4041–#4052 are deliberately **left alone** — owner: *"leave the epics now that they
are there, for next time do it better"*. Unwinding them would be a second churn over the same rows
for no gain. This changes only what the next sweep does.

Docs/skill-only change: no `src/` touched, so no migration or coverage implication. The two
remaining mentions of "new epic" in the file are the prohibitions themselves.


souliane:
Correction to the PR description, for the record.

The body states "README skills catalogue regenerated by its pre-commit hook". That is true only for `README.md` — the `update-readme-skills` pre-commit hook does keep it generator-consistent. It was **not** true for `docs/generated/skills-catalogue.json` and `.md`, which the hook does not cover and which were left carrying the old epic-minting summary. That is what turned `docs-drift` red, and it needed the separate regeneration commit af26935b.

Leaving the original text in place rather than rewriting it, so the record shows what was claimed and what was actually the case.

Issue #4053, opened 2026-08-03
config_setting export should carry each key's type and admissible values, not just its value


souliane:
`config_setting export` emits each key's value and its help text, but nothing about what the key
will ACCEPT:

    agent_runtime = "headless" # whether dispatched work runs headless or in an interactive terminal
    autoload = true # load the suggested skills automatically at session start
    wip = "full" # how much work may be in progress at once

Reading that, an operator cannot tell whether `agent_runtime` takes any string or exactly two
values, nor what the alternatives to `"full"` are. The export is the surface used to review
defaults away from the dashboard, which is precisely where that question gets asked.

## The derivation already exists

`teatree.config.schema.setting_choices(key)` returns a key's admissible values when the schema
constrains it to a listable set, and `()` when the type is open (str, int, list, mapping). It is
the same single derivation the dashboard consumes to decide between a `<select>` and a free-text
input — 115 of 267 settings are constrained. The export simply does not read it.

So this is not new inference: it is routing an existing answer to a second surface, which also
keeps the two surfaces from disagreeing later.

## Shape

Carry the type, and the choices when there are any, as a TOML comment beside the value — the
export must stay valid TOML and stay round-trippable through `import`. For example:

    agent_runtime = "headless"  # str, one of: interactive | headless — whether dispatched work runs headless…
    autoload = true             # bool — load the suggested skills automatically at session start
    wip = "full"                # str, one of: slow | medium | full | boost — how much work may be in progress…
    headless_max_turns = 400    # int — turn ceiling for one headless run

Open types name the type and no list; constrained types name the alternatives. Exact formatting is
the implementer's call provided the file still imports cleanly.

## Worth pinning

A test that a constrained key's export line names every value `setting_choices` returns, and that
an open-typed key names no list — otherwise the two surfaces drift, which is the failure this
change exists to avoid.

souliane:
Fixed on `main` — closing as done.

Landed in `0e8bd216e` (PR #4081), whose body carries a `## C — #4053` section but omitted the number from its `Closes` trailer, so this never auto-closed.

Evidence on `main`:

- `src/teatree/config/setting_groups.py:232-250` — `setting_comment()` composes `setting_annotation(key)` with `setting_help(key)`.
- `src/teatree/config/setting_groups.py:253-265` — `_commented()` attaches that comment to every exported line.
- `src/teatree/config/setting_annotation.py:76-88` — emits the type and admissible values, e.g. `str, one of: slow | medium | full | boost`.

[...]

PR #4071, opened 2026-08-03
feat(config): export says what each key ACCEPTS, not only its value (#4053)

souliane:
feat(config): export says what each key ACCEPTS, not only its value (#4053)

## What
`config_setting export` and `defaults.toml` carried a value and a help sentence,
so a reader could see `wip = "full"` but not whether it takes any string or one
of four words — the question the export exists to answer away from the dashboard.

- `schema.setting_type_name(key)`: the sibling derivation of `setting_choices`,
  off the same field annotation (bool/int/float/str/list/table). Exact-type
  lookup, so a bool never resolves through int.
- `config/setting_annotation.py`: the ONE rendering of that pair as one line —
  `int`, or `str, one of: slow | medium | full | boost`. `bool` names its type
  and no list (its two values ARE the type). `""` and `None` render visibly.
- `setting_groups._commented` prefixes it onto the help sentence. Already the
  single renderer both TOML surfaces share, so export and `defaults.toml` cannot
  disagree; the schema import stays deferred, keeping this module pydantic-free.
- `defaults.toml` regenerated through the same `render_toml` emitter — comments
  only, no value moved, `export -> import -> export` still a fixed point.

Tests observed RED first (267 failing assertions on the unannotated comment):
every constrained key's rendered line names every value `setting_choices`
returns; an open-typed key names its type and no list; the annotated dump still
imports back into the store.

Open questions & assumptions
- bool names its type and no list rather than `true | false` (assumed) — the
  issue's own worked example shows `autoload = true # bool — …`, and 104 keys
  would otherwise carry a list the type already states. Pinned by its own test.
- `dict`-valued settings are named `table`, TOML's word for what they render as
  (assumed); `list` and the other names follow the issue's examples.
- `None` (only `agent_harness_provider`) renders as `none` and `""` as a visible
  pair of quotes (assumed) — both are real states the schema admits, and bare
  they would vanish into the separators beside them.
- The annotation lands on `defaults.toml` too, not just the export dump
  (assumed) — `grouped_settings_table` is deliberately the one renderer for
  both, and splitting it would reintroduce the drift this change removes.
- Open: whether the dashboard should show the type token beside its select. Out
  of scope here; it already derives the choices from the same function.

## Why
Reviewing defaults away from the dashboard (`config_setting export`, `defaults.toml`)
is where "may this take any string, or one of a fixed set?" gets asked, and the
export answered only the current value, not the constraint. The derivation already
existed for the dashboard's selects (`schema.setting_choices`); this routes the same
answer to the TOML comment so the two surfaces can't disagree.

### 1. BLUEPRINT § alignment
Config/settings surfaces (BLUEPRINT § on `grouped_settings_table` / `setting_help.py` /
`setting_choices`). A setting's admissible values are derived once and routed to every
surface that explains a setting — the dashboard select today, the TOML comment now. No
second source of truth introduced. BLUEPRINT.md updated in this PR to describe it.

### 2. FSM phase boundaries
n/a — no `Ticket.State` / `Worktree.State` transition.

None. No `OverlayBase` hook, scanner registration, or `*Backend` Protocol touched.

### 4. Component boundaries
- `src/teatree/config/setting_annotation.py` (new) — the one answer to "what this key
  accepts": `setting_type_name(key)` off the field annotation, paired with the schema's
  existing `setting_choices(key)`, written out as one line. Sits on the surface side of
  `schema` for the same reason `setting_choices` gives about its own values — how a
  stored form READS is a surface decision (`dict` renders as `table`) — and because
  `schema.py` sits at its module-health LOC ceiling (`schema.py` ends up untouched).
- `setting_groups._commented` — already the one renderer both TOML surfaces share.

### 5. Dependency direction
`setting_annotation` → `schema` (config → config, no new cross-module edge). `tach check`
green. `setting_groups` reaches `setting_annotation` through a deferred import so it keeps
its documented pydantic-free import cost.

### 6. Test surface
- `tests/teatree_config/test_setting_annotation.py` — per-shape derivation (bool / int /
  float / str / list / table / enum / Literal / optional-enum), totality over the schema.
- `tests/teatree_config/test_setting_groups.py::TestTheCommentNamesWhatTheKeyAccepts` —
  every constrained key's rendered comment names every value `setting_choices` returns;
  an open-typed key names its type and no list.
- `tests/teatree_core/test_config_migration.py::TestTheDumpSaysWhatEachKeyAccepts` — the
  same property on the real `config_setting export` dump, plus export→import round trip.

### 7. Resilience invariants
No external write. The durable artifact is `src/teatree/config/defaults.toml`, regenerated
through the same `render_toml` emitter the owner-approved snapshot uses, so the
`export(import(defaults.toml))` fixed point holds — verified by the existing shipped-file
conformance suite re-parsing the file.

The join is the flat setting key, unchanged. `setting_type_name` / `setting_choices` /
`setting_help` are all keyed on it and all return the neutral empty value for a key the
schema does not declare.

`_commented` today emits `setting_help(key)` or no comment at all. Both preserved: the
annotation is prepended, and a key with neither annotation nor help still gets no comment.
No safety/privacy/leak gate touched; no must-block test inverted.

### 10. Removability / harness-vs-data
Removable: delete `setting_annotation.py`, revert three lines in `_commented`, regenerate
`defaults.toml`. Harness, not data — the admissible set is derived from the schema
annotation, so a hand-kept table would reintroduce the drift this change removes.


souliane:
Superseded by #4081, which carries this commit cherry-picked (`66074d784`, original authorship preserved) plus the fix for its red test.

Verified at #4081's tip: all 11 files present, `setting_annotation.py` and `test_setting_annotation.py` included, nothing dropped by the cherry-pick.

Consolidated per the owner's instruction that the settings/loop batch ship as a single PR rather than one per item.

Issue #4185, opened 2026-08-04
Eight loops are structurally starved: the timer chain is built from Loop.enabled, which the preset overrides — and every diagnostic reports them healthy


souliane:
> **Reworked 2026-08-04** after the preset model was settled. The body below cited `engaged` as the active preset; the settled set is five — `present` (was `engaged`), `away` (was `unattended` + `offline`, merged), `low-token` (was `low-power`), `maintenance` (redefined), `off`. `heads-down` is cut. **The rename has not landed yet**: `origin/main` and the live DB still carry the seven old rows, so an implementer reading the code today will still see `engaged`. Read every `engaged` below as "the active preset, whatever it is currently named". The finding is unaffected by the rename — it is about `Loop.enabled` being the lowest-precedence input while the timer chain is built from it alone.

## The defect

A loop that the active preset forces ON while its `Loop.enabled` column is `False` **never gets a timer chain**, so it never ticks — and every diagnostic in the system reports it as healthy.

Three sites read the raw `Loop.enabled` column instead of the unified admission verdict:

1. **`timer_chain_loop_names()`** — `src/teatree/loops/timer_reconciler.py:92`:
   ```python
   enabled = set(Loop.objects.enabled().values_list("name", flat=True))   # .filter(enabled=True)
   return registered & enabled
   ```
   Worse than passive: `ensure_loop_timers` actively **prunes** any timer outside `chain_names` (`timer_reconciler.py:157-162`), so a timer that did exist is removed.

2. **`_measured_loops`** — `src/teatree/loops/loop_staleness.py:250` — same `Loop.objects.enabled()`, so the staleness alarm cannot see the starved loop either.

3. **`_admission_block` / `loop_block_reasons`** report the loop as **admitted** (empty block string), which is what the dashboard renders.

But the actual admission chain (`loop_table.py:185-255`, `loop/loop_state_db.py:39-65`) is:

```python
if held: return False
if forced is not None: return forced
return preset_state if preset_state is not None else configured_enabled
```

`Loop.enabled` is the **lowest-precedence** input, consulted only when the preset has no opinion. The active preset `engaged` carries an explicit opinion on **all 27 loops**, so `configured_enabled` is never reached — yet the timer chain is built from it alone.

The result: the loop is admitted, reported healthy, invisible to the staleness alarm, and has no driver.

## Evidence

Every loop with `enabled=1` has exactly **1 READY + N SUCCESSFUL** timer rows and ticked within the hour.

[...]

PR #4193, opened 2026-08-04
feat(core): vendor sync — two Stop gates, skill progressive disclosure, dream loop, deploy topology and path resolution

souliane:
Vendor sync from the downstream fork. 296 files, +11632/-2518. The previous description covered
roughly 8% of the diff; this one enumerates all of it, because a reviewer cannot approve what is not
listed.

## What is in here

**Two Stop gates and their detectors.**
`hooks/scripts/{answer_first,unbacked_claim}_gate.py`, `src/teatree/hooks/{answer_first,unbacked_claim}_scanner.py`,
registered in `hooks/scripts/hook_router.py`, with kill-switch subcommands in
`src/teatree/cli/teatree_gate.py` and their settings in `config/{cold_hook_settings,schema,setting_help}.py`.

The answer-first gate fires when a user asks something answerable and the turn reports a delegation
instead. Acting is not answering: dispatching work says nothing the asker wanted to know, so they ask
again. It is a three-way conjunction — an answer-seeking question, a delegation report, and no answer
— so precision comes from the shape rather than a word list. Unlike its sibling Stop gates it does
not skip an attended turn, because a waiting human is exactly when the failure hurts.

The unbacked-claim gate refuses a diagnosis or an escalation with no artefact behind it.

**Router shrink.** `_git_state_for_repo` / `_render_git_state_section` move from `hook_router.py` into
`stop_snapshot_slot.py`. This is not cosmetic: the router is over the module-health LOC cap, over-cap
files may only shrink, and the gate registrations add lines. The extraction is what pays for them.

**Skill progressive-disclosure split** (~65 files). Long skills keep a spine in `SKILL.md` and move
mechanics into `references/*.md`. This is the largest cluster and the one that most needs review —
skill injection concatenates `SKILL.md` only, so doctrine moved into `references/` stops reaching
sub-agents. Guard tests that assert on skill bodies move with the content.

**Dream loop** — distillation cursor, decay budget tiers, re-index, plus migration
`0052_dreamrunmarker_distill_cursor`.

**Deploy topology and host identity** — `deploy/docker-compose.host-identity.yml`, `deploy.sh`,
`deploy/t3`, `watchdog.sh`, `hooks/scripts/session_lane.py`.

**Path and venue resolution** — `paths.py`, `core/worktree/{venue,stack_probe}.py`, `worktree_env.py`,
`utils/ports.py`.

**Management commands and runners** (~21 files), **agents / lane_b / action verification** (~22),
**skill pinning** (`provisioning/skill_pin.py`, `cli/doctor/checks_skill_pins.py`,
`cli/setup/skill_pin_audit.py`), the **`single_branch_repos`** setting and its guard hook, and
assorted `cli/review`, `cli/slack`, `cli/doctor`, `db/write_domain`, `quality/antipatterns.yaml` and
eval-fixture changes.

**Vendor-sync migrations** — `0052`, `0054`, `0055`, `0057_merge_upstream_vendor_sync`, plus a merge
node reconciling the leaf against main.

**Removed** — the `private_tests` setting (no remaining code readers) and the `e2e run --headed` flag.
| area | files |
|---|---|
| `src/` | 103 |
| `tests/` | 81 |
| `skills/` | 63 |
| `evals/` | 19 |
| `hooks/` | 12 |
| `docs/` | 10 |
| `deploy/` | 5 |
| `agents/` | 2 |

## Round two — the cold-review HOLD, discharged

`origin/main` is merged in (never rebased). Every gate result measured before that merge is void and
has been re-run.

### Blocker A — the `origin/main` merge

`git merge-tree` reported exactly two conflicting files, both from
[#4196](https://github.com/souliane/teatree/pull/4196).

`src/teatree/loops/loop_staleness.py` takes `origin/main` **whole**, and only after confirming main
already carries every semantic this branch added:

| this branch added | in main? | where |
|---|---|---|
| force-OFF implies `suppressed` | yes | `_is_suppressed` = `not planes.admits(...)`; `loop_state_admits` returns `forced` whenever it is not `None` |
| force-ON beats an OFF preset | yes | same predicate — `forced=True` short-circuits before `preset_state` |
| the arm-enumerating cause line | no, deliberately | main derives suppression from the verdict rather than listing arms, so enumerating them would re-create the drift #4196 removed |

Keeping this branch's body would have re-narrowed `_measured_loops` to the column-enabled rows and
re-broken the frozen-fleet alarm #4196 had just restored.

`tests/teatree_loops/test_loop_staleness.py` takes main as the base, then ports forward the two cases
main has no equivalent of (force-ON outranking an OFF preset, and a hold outranking force-ON),
rewritten in main's real-row fixture style rather than this branch's patched-tuple seam.

**The frozen-fleet measurement.** A new test builds the all-off-mask case from real `Mode` /
`ModeOverride` rows through the real resolver — no patched seam — and asserts `frozen_fleet is True`
with `considered == 3`. Two planted controls:

- measuring **membership alone** → `considered == 0`, `frozen_fleet` False. That is the seven-hour
  blind spot returning, and the new test is RED on it.
- measuring **column-enabled alone** (this branch's version) → the #4185 case
  (`test_a_preset_forced_on_column_disabled_loop_is_measured_and_stale`) goes RED.

Restored: 40 passed in the file, 154 across the loop suites. A read-only `loop_health()` against the
live box reports `considered=21`, `frozen_fleet=false`, mode named — no blind spot.

### Blocker B — the answer-first gate false-fired 10/10

Two independent defects, both of which passed all 35 of the gate's own tests.

**The parser.** `last_assistant_turn` ended the turn at the first `user` entry, but a tool RESULT is
recorded as a `user` entry — so the turn was cut at the first tool call and everything written before
dispatching fell outside it. `_last_user_text` already walked past those entries; its mirror did not.
The predicate now has one home in `question_gates` and both halves read it.

**The detector.** Leg 3 was a keyword allowlist, so it failed CLOSED on any causal answer phrased
outside the list. It now also passes when the turn carries declarative residue once the delegation
report and its boilerplate are stripped, so it blocks only a turn that is nothing but the dispatch
line. Enumerating the dispatch side is sound where enumerating the answer side was not: dispatch
vocabulary is a small closed domain the agent itself writes; answers are open natural language.

Measured over the 10 recorded cases, plus 4 true positives as the control:

| case | before | after |
|---|---|---|
| answer before the dispatch tool call — explicit "because" | BLOCK | allow |
| answer before the dispatch — polarity opener ("No — the shard is red") | BLOCK | allow |
| answer before the dispatch — "I don't know yet" | BLOCK | allow |
| answer before the dispatch — "It was blocked by a failing shard" | BLOCK | allow |
| "It was blocked by a failing shard." | BLOCK | allow |
| "Conflicts with main." | BLOCK | allow |
| "It is still waiting on CI." | BLOCK | allow |
| "The branch is behind main by eleven commits." | BLOCK | allow |
| "The merge keystone needs a human approver." | BLOCK | allow |
| "A reviewer left it on hold last night." | BLOCK | allow |
| **control** — "Dispatched a lane to merge it." | BLOCK | **BLOCK** |
| **control** — dispatch + "Will report back." | BLOCK | **BLOCK** |
| **control** — handoff + "Standing by." | BLOCK | **BLOCK** |
| **control** — "On it. Fanned out three lanes." | BLOCK | **BLOCK** |

10/10 → 0/10 false fires, with all four true positives still firing. The first four were measured at
the GATE level against real transcripts (the parser defect); the other six at the scanner level (the
detector defect).

**The gate ships ENABLED**, default `True` in both `cold_hook_settings.py:75` and
`answer_first_gate.py:55`, which agree.

Blast radius across all six `last_assistant_turn` consumers, re-run at the post-merge head:

| consumer | verdict |
|---|---|
| `hook_router` #807 structured-question gate | 107 passed |
| `standing_goal_stop_gate` | 20 passed |
| `unbacked_claim_gate` | 15 passed — inherits the parser fix, default left at `True`, no false-fire shape found |
| `completion_claim_gate` | 11 passed |
| `answer_first_gate` | 28 passed |
| `speak_on_stop` (the 600-char TTS slot) | 15 passed |

### Blocker C — doctrine that reached zero sub-agents

`skill_injection` concatenates `SKILL.md` bodies and has no concept of a reference file. Restored to
their spines: the on-behalf egress owner with `t3 slack react` / `notify react`, the stale-answer
doctrine, the review skill's `gh issue view` fetch plus the refuse-blind-reviews and
reviewer-does-the-verification verdicts, the whole-file `--ours` prohibition, and the squash-merge
cross-check. References keep the mechanics.

Three deterministic checks, each observed RED against its planted loss:

1. `test_mandated_command_pinned_by_a_scenario` now asserts against the `agent_sections` **slice**
   rather than the whole file, resolved through the harness's own `load_agent_definition`. Measured:
   with `t3 slack react` present in the file but outside the graded section, a whole-file grep finds
   it and the slice check goes RED — the blind spot the old assertion had.
2. `test_reference_is_never_a_doctrines_only_home` — a command MANDATED in a reference must appear in
   the sibling `SKILL.md`.
3. `test_non_negotiable_stays_in_the_spine` — a `Non-Negotiable` HEADING in a reference must head a
   spine section. **This check found the sixth loss the review had not named** (the squash-merge
   cross-check).

Checks 2 and 3 carry a frozen baseline of what is already orphaned on `main` (out of scope here per
the in-scope rule) and fail if a baseline row is fixed without being removed, so the list can only
drain. Neither is a suppression: a new orphan fails immediately.

### Blocker D — the two red eval scenarios

Route taken **per anchor**, from the transcripts of the CURRENT failing job (run 30952705003), not
plan 64's two-heads-stale ones.

| scenario | anchor | route |
|---|---|---|
| `failed_command_is_not_a_finding` | 1 (`fatal:`) | untouched — already matching |
| `failed_command_is_not_a_finding` | 2 | **route 2** — widened to word FORMS of already-anchored words |
| `staged_version_is_what_gets_committed` | 1 | untouched — already matching |
| `staged_version_is_what_gets_committed` | 2 | **route 1** — doctrine strengthened, matcher untouched |

The recorded answer for the first withheld its conclusion exactly as asked — "I have not established
whether the two trees match or differ", "Status: Inconclusive" — and graded red purely on inflection
("established" vs `establishes`, "inconclusive" vs `conclude`). Verified deterministically against
the recorded text: old anchor no match, new anchor matches on `not established`.

The second gets no widening, because there is no verify/check/read/confirm verb **anywhere** in its
transcript. It reasons correctly about the mechanism and then asserts what the pushed commit contains
without reading it and without naming the read that would settle it — substantively incomplete on the
dimension the scenario grades. So the doctrine gains what it was missing: read the state the claim is
ABOUT, never a local proxy.

Controls: both `_fail` fixtures still RED, both `_pass` fixtures still GREEN,
`green_lane_requires_a_pass_count` untouched and both its fixtures unchanged, `t3 eval coverage`
reports 0 gaps.

### Disposition — every round-two finding

| finding | disposition |
|---|---|
| A — `origin/main` merge / `loop_staleness` conflict | FIXED `fb252b9e7` |
| B — answer-first gate false-fires | FIXED `d152f97bc` |
| C — doctrine reaches zero sub-agents | FIXED `0fd72ed5a` |
| D — two red eval scenarios | FIXED `7afd5deb3` |
| M1 — dream cursor atomicity had no guard | FIXED `c27f1aaab` |
| M2 — `_gh_conflict_state` untested on the primary forge | FIXED `c27f1aaab` |
| M3 — `prek_hook` no restore on a RAISED install | FIXED `c27f1aaab` |
| M4 — back-compat re-export shim in `dream.engine` | FIXED `58d8eb764` |
| m1 — inert `merge_extra` typing claim | NOT A DEFECT in code — narrative overclaim, corrected here. `merge_extra` has 10 call sites and only two `cast("TicketExtra", ...)`; no code change |
| m2 — orphaned `fetch_pr_is_draft` | FIXED `58d8eb764` — deleted with its four assertions; the protocol requires `fetch_pr_draft_state`, and no caller existed in `src/`, `tests/` or `scripts/` |
| m3 — decay budget projection modelled arithmetically | FIXED `58d8eb764` — kept arithmetic, added the table test that makes its "EXACT against `render_index_lines`" claim falsifiable |
| m4 — no test that `drop_foreign_config_hooks` SPARES a root-bound hook | FIXED `c27f1aaab` — the destructive function's only negative control |
| m5 — stale `ac-reviewing-codebase` comments | NOT A DEFECT — read all four named sites; each describes the skill accurately. `cli/assess.py` searches only agent-skill dirs because the in-tree copy carries `SKILL.md` and no `scripts/cli.py`. The frozen migration was not touched |
| m6 — false byte-identity claim for `ac-reviewing-codebase/SKILL.md` | CORRECTED here: it is **not** byte-identical — it drops exactly one line, `subagent_safe: true`, and that drop is REQUIRED by `tests/conformance/test_no_subagent_safe_in_skills.py` |

### Migration

`0058_merge_vendor_sync_and_availability_removal` is kept — its number is load-bearing, since the
merge node must sit above both `0057` leaves. [#4199](https://github.com/souliane/teatree/pull/4199)
has renumbered to `0059`, so that collision is resolved. If
[#4198](https://github.com/souliane/teatree/pull/4198) lands first, this renumbers and keeps both
`0057` parents.

### Local verification

Full suite at `58d8eb764`: **3 failed, 39650 passed**, 68 skipped, 5 xfailed. All three failures were
one root cause in `tests/teatree_quality/test_skill_cli_ratchet.py` — restoring the `gh issue view` /
`glab issue view` fetch to the review spine moved those raw calls back out of the reference, so the
per-item ledger no longer matched the live set. Fixed in `e6fc91cd9` by regenerating the ledger
through the tool's own `--update-baseline`; the diff is exactly those two keys, both byte-identical
to keys `origin/main` already carries. That file's 80 tests are green, and `pyproject.toml` is the
only thing that names the ledger, so no other test reads it.

### No relaxations

`git diff origin/main...HEAD` adds no `noqa` beyond the repo's standard deferred-import (`PLC0415`)
and crash-proof-hook (`BLE001`) markers, no raised ceiling, no `--no-verify`, no unconditional skip.
The repo's own relaxation detector passes. The over-cap ratchet holds: `hook_router.py` is 4030
non-blank-non-comment lines against 4068 on `main`.


souliane:
## Cold review — verdict: hold at `d259312d3` (being fixed in place)

Recording this on the PR so the findings are durable and the fixes are auditable. 31 findings: 2
critical, 15 major, 14 minor. `main` is green, so everything below is introduced here.

### Critical

**1. The answer-first gate cannot fire on any turn that used a tool — the flagship feature is inert.**
`hooks/scripts/answer_first_gate.py:81-98`. `_last_user_text` walks the transcript newest-first and
stops at the **first** entry whose role is `user`. In real transcripts a `tool_result` is recorded as
`type:"user"` / `message.role:"user"` with `tool_result` content blocks, so the text-block filter
yields nothing, the function returns `""`, and `find_unanswered_question` returns `None` on its
empty-input guard.

Measured against real transcripts: a turn with no tool call blocks as the test asserts; a turn with a
tool call is allowed. A "delegation report" turn is by definition one that made a `Task`/`Agent`
call, so the gate never fires on the case it exists for — it only fires when the agent *claims* a
dispatch it never made. `tests/test_answer_first_gate_stop_hook.py` builds transcripts with no tool
calls at all, which is why it passes: the test exercises a path production never reaches.

Fix: keep walking past user entries whose blocks are all `tool_result`, and pin it with a transcript
that contains a tool call.

**2. CI is red on every lane** — 12 test shards, 4 shuffles, lint, module-health, docs-drift,
test-path-mirror, and 2 behavioural eval scenarios.

### Major

- `deploy/docker-compose.host-identity.yml` is entirely unwired. `deploy/t3:22` assigns
  `HOST_IDENTITY_FILE` and never uses it; `compose()` still passes only `$COMPOSE_FILE`; `deploy.sh`
  never adds the `-f`. The file's own header states both do. The off-box path-identity fix is dead.
- `skills/ship/SKILL.md` lost the verbatim near-zero-comments preamble to `references/`.
  `src/teatree/agents/skill_injection.py` concatenates only `SKILL.md`, never `references/*.md`, so
  every ship-phase sub-agent silently loses the rule. Same mechanism behind two failing prose tests:
  the split moved graded doctrine out of the body that actually reaches a worker.
- `cli/doctor/self_heal_slack_config_token.py:61-65` handles ROTATED/UNRECOVERABLE only, so
  `STORE_UNWRITABLE` prints nothing and returns True — a silent green while the token pair ages out.
- `cli/slack/setup.py:266` — `assert_writable()` raises `SlackConfigTokenStoreUnwritableError` caught
  nowhere; `t3 slack setup` / `provision` / `socket_doctor` traceback instead of exiting cleanly.
- `loops/dream/distill.py:189` — the cursor is written in bare autocommit *before* `write_clusters`,
  and advances even over batches that raised. An auth outage walks the cursor through the whole
  corpus, losing memory windows permanently.
- `core/management/commands/dream.py:255` — one broken batch out of 24 now aborts cross-link, merge,
  re-index and decay *after* rows were already written, leaving them unreconciled. `main` counted the
  failure and carried on.
- `loops/dream/decay.py:562` — the budget projection renders the header as `render_index_lines([])`
  (~180 bytes) while the written index replaces it with the priority preamble. With a preamble
  present, decay archives zero files and the gate fails nightly with no diagnostic naming the cause.
- `loops/loop_staleness.py:281` — `_is_suppressed` inverts its own docstring: a loop forced ON
  against a preset that masks it off resolves as admitted yet reports suppressed, so a wedged
  forced-on loop reads clean.
- `core/overlay.py:609` — `get_repo_clone_url` has zero callers; the live path is
  `provisioning.repo_clone_url`. An overlay implementing the new hook gets `""`. Removing it also
  restores the identity-surface cap.
- `backends/forge_merge_rpc.py:152` — `_glab_conflict_state` is defined and never called, so GitLab
  merge state stays UNKNOWN and every open GitLab MR emits a false "conflict unreadable" signal. The
  `gh` twin is wired correctly.
- `provisioning/declared.py:131` skips specs with fewer than three segments, so the repo's only
  third-party pin is never measured — and the pin check emits nothing when every *measured* pin is
  current, so silence reads as "all pins current". An un-enumerable pin must report UNVERIFIED.
- `provisioning/skill_pin.py:120` string-compares a head sha to the pinned ref, so a legal symbolic
  pin is permanently "behind" and the remediation advises freezing a deliberately-floating ref.
- `core/prek_hook.py:161` unlinks hooks from the clone's **shared** hooks dir before `prek install`,
  with no restore on failure — a failed install leaves every sibling worktree of that clone with no
  pre-commit/pre-push/commit-msg.
- `agents/harness.py:383` pops the parked resume thread before evaluating a credential lookup that
  can raise; the new `CredentialError` catch returns without restoring it, and exhaustion on this
  lane terminates rather than parking.
- `utils/ports.py:71` — the in-container branch answers "how do I reach the host", but the value is
  cached as a DB host for a different container set.

### Minor, in brief

Both new Stop gates misfire at a measured 0.13% of 3,847 real turns — systematically, not randomly:
"blocked because" trips the unbacked-claim gate despite a comment claiming that case was excluded;
`.log`/`.jsonl`/`.conf` are not recognised as artefact citations; and an honestly-hedged, URL-cited
escalation is penalised for admitting some evidence is still pending. The answer-first predicate
treats polite imperatives ("can you also…") as questions. No deadlock risk — a block is capped at one
extra turn.

Also: the generated skill catalogues were hand-edited rather than regenerated; a newly-added
portability test is red against its own tree; and the README still documents a flag this PR removes.

### Verified clean — not worth re-litigating

The migration graph is sound (single leaf, correct `max_migration.txt`, no backfill hazard). No
`typer.Exit` in any touched management command. The stash rework, merge-driver command, start-runner
destroy-last reordering with its fail-closed container check, and the `private_tests` removal all
check out.

Issue #4049, opened 2026-08-03
Epic: declared-versus-installed version skew


souliane:
## Why these are one thing

The runtime the code declares and the runtime it actually gets have drifted apart, in the language floor and in the pinned e2e browser image.

## Members

- [x] #3846
- [x] #3972

Each member keeps its own thread until it is folded here; folding closes the standalone with a
pointer back to this checklist, never because the work is done. Reopen any of them if this epic
ever needs to split back out.

souliane:
## Dissolved 2026-08-24 — its two members share a label, not a design

This host was created by the 2026-08-03 sweep, which folded #3846 and #3972 into it and closed
both as `not planned` while copying only their issue numbers here. Neither requirement text
ever moved, nothing implemented either one, and this host carried the `epic` label — which
intake refuses outright — so the two requests were unreachable for three weeks.

On re-reading both members, there is no shared design to reframe around. One is a language and
framework floor bump touching `pyproject.toml`, the CI matrix, `dev/Dockerfile.test` and the
lockfile. The other is a Docker base-image tag drifting against a lockfile-resolved pip package,
touching `dev/Dockerfile.e2e` alone. Disjoint toolchains, disjoint files, no common fix. The
only thing they had in common was a thematic phrase — the runtime the code declares versus the
runtime it gets — which is the kind of surface pairing that produced the problem in the first
place.

Forcing a unified problem statement onto them would have produced a host nobody could implement
in one change, so the fold is refused rather than rewritten.

## Where the requirements live now

Both members are reopened, carry the admit label so intake can claim them, and each has been
re-checked against current `main` with its acceptance criteria restated:

- #3846 — move to Python 3.14 and the current Django, sequenced Python then Django then the
  dependency sweep. Still entirely unimplemented; the one merged pull request that referenced
  it turned out to be unrelated.
- #3972 — realign the Playwright base image with the lockfile-resolved package. Still
  unimplemented, and the drift has widened since filing, so the fix should target whatever the
  lockfile resolves at the time rather than the version named in the original report.

Closing this row only because both requirements are now provably carried elsewhere and are
individually claimable, which was not true while it stood.

Issue #779, opened 2026-05-16
fix(lifecycle): visit-phase/pr-create silently target the wrong isolated DB when not run from the worktree dir


souliane:
## Recurring structural defect (bit #764, #628, #778 — same root cause each time)

Each worktree has an isolated DB. `t3 <ov> lifecycle visit-phase` and `t3 <ov> pr create` resolve the DB from the **process CWD / invocation context**, not from the ticket/branch. When an agent runs them from a different directory than the ticket's worktree (the main clone, another worktree, or after a context where CWD drifted), the write/read silently lands in a **different isolated DB**:

- Phase visits report success but "don't persist" (they persisted — in the wrong DB).
- The shipping gate computes a stale/partial phase set for the ticket's actual worktree DB and blocks (`shipping requires: <phase>`), even though the phase was "recorded."
- Diagnosis is expensive and repeatedly misattributed (mistaken for a #755-class lost-update; #778's was actually a `retro` recorded against the wrong DB + a misread "missing: reviewing" that was really "missing: retro").

## Evidence

- #764: reviewer's `reviewing` recorded via global `t3` → landed in a different DB than `pr create` read; needed re-record from the worktree.
- #628: hand-created ticket + cross-DB attestation; ship-blocked.
- #778: maker's `lifecycle visit-phase 169 retro` ran against a non-worktree DB → gate saw `retro` missing; resolved only by re-recording from inside the worktree.

## Fix direction

Make `lifecycle visit-phase` and `pr create` resolve the correct isolated DB **deterministically from the ticket/worktree identity**, not the ambient CWD — or fail loudly with a clear "you are not in ticket N's worktree; its DB is X" message instead of silently writing to the wrong DB. A phase write/gate-read for ticket N must always hit ticket N's worktree DB regardless of where the command is invoked. Add a regression test: `visit-phase`/`pr create` for a ticket from a foreign CWD must either target the correct DB or refuse — never silently write elsewhere.

## Acceptance

- [ ] `visit-phase <id> <phase>` and `pr create <id>` operate on the ticket's own worktree DB regardless of process CWD, OR refuse with an explicit wrong-worktree error naming the correct path.
- [ ] Regression test for the foreign-CWD case.
- [ ] No silent cross-DB write/read remains on the lifecycle/ship path.

Recurring; HIGH (it has cost multiple multi-cycle misdiagnoses this session). Pairs with the worktree-DB-isolation handling and the #748 session-provisioning work.

souliane:
[...]

PR #4211, opened 2026-08-05
fix(mcp): delegate on an invisible claim, drop the agent SDK from django.setup(), and exercise the server in doctor

souliane:
## What was wrong

`t3 mcp serve` was dead for an entire session while `claude mcp list` reported only
`Connection closed`. Three independent faults, all presenting identically.

**Fault 1 (host env, fixed by reinstall before this PR).** The host `uv tool` env carried
`mcp 1.28.1` against a declared `mcp>=2,<3`; `teatree/mcp/server.py` needs the 2.x API.
Nothing was *missing*, so the existing missing-deps check was silent — the skew only
surfaced as an `ImportError` at the moment the server had to start. The container was
already current; only the host had drifted, three weeks behind `pyproject.toml`.

**Fault 2 — an invisible claim read as an absent one (#4041 class).**
`teatree.cli.mcp_owning_domain.owning_domain_wrapper` consulted
`ControlDbBoundary(CANONICAL_DB).read_write_allowed` to decide whether a container owns
the control DB. Ownership is a claim FILE beside the database — and the canonical
control-DB directory is a container-only mount BY DESIGN, so on the host it does not
exist at all. The claim cannot be stat'd, `read_write_allowed` answers `True` for a
database the container owns outright, and the server serves natively.

Its docstring claimed that failing to delegate was safe because the boundary guard would
still refuse the write loudly and with the remedy. That is false for this process: the
guard never gets to speak. Measured on the affected host, the natively-served server
answers `initialize` (so `claude mcp list` reports **connected**) and then fails **every**
ORM-backed read on `unable to open database file` — 4 such errors in one handshake.

`claim_is_observable` now splits "I cannot see whether a container claimed it" from
"nobody claimed it", and UNKNOWN delegates. Every genuinely safe serve-here case is
untouched: in-container, already delegated, observable-and-unclaimed, unresolvable clone,
missing or non-executable wrapper — so a plain clone with no container still serves
natively.

**Fault 3 — the agent SDK in every `django.setup()`.** `AgentsConfig.ready()` imported the
runners eagerly, pulling `agents.headless` → `agents.harness` →
`pydantic_ai.models.openai` → the whole `openai.types.*` pydantic tree. Both registries
are inversion seams that store a callable, so they now take deferring thunks and the SDK
is paid only by the code path that needs a model client.

## The fail-fast check

`t3 doctor check` gains `_check_teatree_mcp_liveness`. The pre-existing registration check
is a WARN by design, and a WARN is exactly what let this survive: one advisory line among
~twenty, most of the others benign host/container artifacts.

- **Hard FAIL**, gating the doctor exit code — not advice.
- **Exercising, not declarative.** It spawns the registered `t3 mcp serve`, speaks a real
  `initialize` frame over stdio, and keeps the stderr `claude mcp list` throws away.
- **Catches the false green.** `initialize` touches no ORM, so a server whose every real
  tool call fails still reports connected; a DB-unreachable signature in stderr fails the
  check even when the handshake succeeded.
- **Distinguishes the three causes** and names the exact remedy for each: stale tool env
  (`ImportError`), delegation failure (`OperationalError`), startup past the handshake
  budget.
- **Self-repairs the mechanical cause.** Declared-versus-installed version skew is detected
  (`teatree.utils.dep_skew`, real specifier semantics — the bootstrap-safe
  `dep_drift` deliberately cannot do this), repaired via
  `uv tool install --editable . --overrides uv-overrides.txt --reinstall`, and re-verified.
  It names **which side** of the host/container boundary is stale, because they drift
  independently. Refs #4049.

## Timings (same box, interleaved A/B, load ~4.6)

| | before | after |
|---|---|---|
| `t3 --help` | 2.56s | 1.78s |
| `import teatree.cli` + `ensure_django()` | 2.41s (openai loaded) | 1.54s (openai **not** loaded) |
| `t3 mcp serve` handshake | 3.22s, **4** DB errors | 2.40s, **0** DB errors |

Under real contention (load 19, cold) the same `import teatree.cli` + `ensure_django()`
measured **18.4s**, ~10s of it that one import — enough to exceed a client's handshake
window on its own.

## Mutation evidence — every new guard shown RED

| mutation | test that went red |
|---|---|
| delegation predicate reverted to pre-fix | `test_an_invisible_claim_delegates_rather_than_serving_here` |
| `ready()` imports the runner eagerly again | `test_django_setup_loads_none_of_the_heavyweight_agent_modules` (names openai, pydantic_ai.models.openai, agents.harness, agents.headless) |
| the FAIL finding degraded to an advisory WARN | `test_each_cause_fails_and_names_its_remedy` ×3 |
| false-green branch removed (handshake trusted) | `test_a_connected_server_whose_db_is_unreachable_is_a_failure` |
| handshake budget disarmed | `test_a_slow_but_successful_handshake_still_fails` |
| skew degraded to a missing-only check | `test_a_dist_below_its_declared_floor_is_skew` |

The import-weight guard runs in a **fresh interpreter**: asserting against the test
process's own `sys.modules` would be vacuous, since the suite imports the harness
elsewhere.

## Test evidence

`2434 passed, 9 skipped` across the affected surface (agents, doctor, core dispatch,
utils, cli mcp). Full prek gate chain green on commit — including `check-chokepoints`,
which correctly refused a raw `subprocess.run` and sent the spawn through
`teatree.utils.run`.

## A fourth fault the fix itself surfaced

Delegating is not enough on its own. `deploy/t3` **refuses** to dispatch when its host
cwd sits inside a checkout the container cannot see — a guard against a cwd-SENSITIVE
command silently resolving against the container's own copy of the source. A client
launches the MCP server from whatever directory the session happens to sit in, which is
routinely such a checkout, so naive delegation would have traded a server that cannot
open the DB for one the wrapper declines to start: the same bare `Connection closed`.

`mcp serve` is the cwd-INSENSITIVE case the wrapper's own comment carves out — a stdio
server over the control DB, resolving no tree from cwd — so the handoff now takes the
escape the refusal names and declares a container-side cwd (the control-DB directory:
container-only, always mounted, carrying no checkout to be wrong about). An
operator-set `TEATREE_INVOCATION_CWD` is left alone.

Measured from the deploy checkout, before and after the declaration:

```
deploy/t3 mcp serve, no declaration   -> 0 bytes of stdout (wrapper refusal)
deploy/t3 mcp serve, with declaration -> {"jsonrpc":"2.0","id":1,"result":{...}}
```

## End-to-end evidence — the false green is real

`claude mcp list` currently reports `teatree: t3 mcp serve - ✔ Connected`. That green is
false. Driven over real stdio with the connection held open:

```
PRE-FIX  (the server claude mcp list just called Connected)
  initialize            -> OK
  tools/call loop_stats -> {"isError": true, "content": [{"text":
                              "Error executing tool loop_stats: unable to open database file"}]}
  db errors in stderr   -> 4

AFTER    (through the domain this PR delegates to)
  initialize            -> OK in 3.7s
  tools/call loop_stats -> {"result": {"tasks": {"pending": 14, "claimed": 3,
                              "completed": 988, "failed": 329}, "dead_letter": 0}}
  db errors in stderr   -> 0
```

The registration exists, the client says connected, and every ORM-backed tool call
fails. That is exactly the state `_check_teatree_mcp_liveness` refuses to call healthy —
and running it live against that server produces:

```
FAIL  teatree MCP server is NOT usable (delegation-failure, 3.2s) — ... NOTE: this cause
      can present as CONNECTED — `initialize` needs no DB, so the handshake succeeds and
      every ORM-backed tool call fails afterwards.
      Captured stderr (`claude mcp list` never shows this):
        django.db.utils.OperationalError: unable to open database file
```

while against the healthy path it prints `OK  teatree MCP server answered initialize in 1.9s.`


souliane:
## Cold review — HOLD

The server fix works. I drove the server over real stdio, held the connection open, sent
`initialize` + `notifications/initialized`, then `tools/call loop_stats`, against this PR's code:

```
initialize            -> OK
tools/call loop_stats -> real ORM rows returned, isError: false
db errors in stderr   -> 0
```

Real ORM read, real rows, zero DB errors. All six guard mutations reproduce RED, so the new guards
are not vacuous. `django.setup()` heavy-module residency drops from 4 heavy modules / 2665
`sys.modules` / 1.63s to 0 / 1298 / 0.55s.

One control worth recording: my first probe ran from a git *worktree* and returned
`no such table: teatree_task`, which looks exactly like the fix failing. It was a probe artifact — a
worktree has a `.git` file, so `resolve_data_dir` (`src/teatree/paths.py:286`) auto-isolates it onto
a per-worktree DB (#779) and `CANONICAL_DB` resolved somewhere else entirely. `paths.py` is
byte-identical to main, so this was the harness, not the PR. Re-running from a non-worktree
`git archive` export with `CANONICAL_DB` matching main's resolution gave the result above. Without
that control I would have filed a false red.

### Findings, most severe first

**1. BLOCKER — CI red, PR-caused, deterministic.** `src/teatree/core/mcp_liveness.py` lands as a new
flat leaf, taking `src/teatree/core/` from 108 to 109 against `PINNED_FLAT_CORE_MODULES = 108`. Fails
`tests/quality/test_no_flat_core_regrowth.py:228` on shard 5, which fails the aggregate `test (3.13)`
gate. Verified locally: HEAD 109 leaves, main 108, ceiling 108. Either raise the pin to 109 with a
justification comment (that file has an established precedent block) or move the module into a
subpackage — a new flat leaf in `core/` is what that guard exists to resist, so prefer the subpackage
absent a concrete reason.

**2. MAJOR — a read-only check mutates the operator's environment.**
`_check_teatree_mcp_liveness` (`src/teatree/cli/doctor/checks_mcp.py:229`) unconditionally runs
`uv tool install --editable . --overrides uv-overrides.txt --reinstall` on skew.
`run_doctor_checks(*, repair: bool = False)` exists and every other mutating check threads it —
`_check_t3_shim_receipt(repair=repair)` at `app.py:475`, `run_bootstrap_checks(repair=repair)` at
`app.py:600`. Neither `_check_teatree_mcp_liveness()` nor the new `_run_mcp_checks()` (`app.py:418`)
takes the flag. So a bare `t3 doctor` hits `run_doctor_checks(repair=False)` (`app.py:673`) at
SessionStart and silently reinstalls the host tool env — replacing the console script the session is
running under — with no `--repair`.

**3. MAJOR — the headline self-repair has zero coverage.** `_repair_version_skew`
(`checks_mcp.py:191`) is patched out in both tests that reach it
(`tests/teatree_cli/doctor/test_mcp_liveness_check.py:86` and `:102`) and exercised nowhere.
Untested: the attempted-guard, `resolve_repair_plan` handling, the `run_allowed_to_fail` spawn, both
failure branches, the re-verify. It is the only path here that mutates the operator's machine, and
the PR body advertises it as a feature.

**4. MODERATE — a docstring guarantee that does not hold.** `_SKEW_REPAIR_ATTEMPTED_ENV` is written
to the current process's `os.environ` (`checks_mcp.py:207`) and read at the top of the same
one-call-per-process function. There is no re-exec anywhere under `src/teatree/cli/doctor/`, so the
guard is unreachable in production and its docstring claim — "so a repair that does not take reports
itself instead of reinstalling on every doctor run" — is false. With finding 2, an unrepairable skew
reinstalls on *every* doctor run. Needs a durable marker, not an env var.

**5. MINOR, latent.** `find_version_skew` (`src/teatree/utils/dep_skew.py`) checks
`requirement.marker` only on the not-installed branch, so a marker-excluded dep that happens to be
installed out of range is reported as skew — which would then trigger the auto-reinstall. Not live:
0 of 20 declared deps carry markers.

**6. OBSERVATION — budget headroom.** `HANDSHAKE_BUDGET_SECONDS = 10.0`, spawn timeout 20s. Steady
state measured 4.4–5.2s across six runs, but one ~11.6s handshake was observed during a
container-restart window, which would have hard-FAILed a server that was in fact coming up. The
delegated path pays a `docker compose exec` per doctor run, so the margin is roughly 2x on a quiet
box and thinner under load.

**7. UNRESOLVED — second CI failure.** `test-shard (3.13, 7)` fails on
`tests/integration/provisioning/test_teatree_self.py::TestTeatreeSelfProvisioning::test_two_worktrees_provision_serve_concurrently`
— `ReadTimeout: timed out (after 20 retries)` at `tests/integration/provisioning/_base.py:137`. No
coupling to the diff found, and this PR strictly *reduces* `django.setup()` cost, so causation is
implausible — but this is PLAUSIBLE flake, **not confirmed**. Needs a re-run to settle.

**8. PR body is stale on Fault 2.** `origin/main`'s inline predicate is logically identical to this
PR's `claim_is_observable` extraction, and main already delegates (probed live). The extraction plus
the missing unreachable-DB test is the right resolution, but the description reads as if it fixes
something main still carries. The live outage is Fault 4 — the wrapper refusing to run when the
checkout is not visible inside the container, producing 0 bytes of stdout. That is the
`Connection closed`, and it reproduces exactly on main.

### Standard bar

Clean. No migrations. No raw `subprocess` in the new `src` files — the spawn goes through
`teatree.utils.run.run_allowed_to_fail`, so no chokepoint is bypassed. No `typer.Exit` misuse; the
checks return `bool`. All 11 added `noqa` are `PLC0415` deferred-import with stated reasons. The
`_run_mcp_checks` refactor preserves aggregation correctly (each check is on the left of `and ok`, so
none is short-circuited away). The `dep_skew`/`dep_drift` split is justified: `dep_drift` is
deliberately zero-non-stdlib for bootstrap, `dep_skew` needs `packaging`.

### CI state (paginated, `?per_page=100`)

All 43 runs complete: **37 success, 3 skipped, 3 failure** — `test (3.13)`, `test-shard (3.13, 5)`,
`test-shard (3.13, 7)`.

### What unblocks this

Finding 1 is a one-line ceiling bump or a module move. Findings 2–4 are one knot: thread `repair`
into `_check_teatree_mcp_liveness`, make the marker durable, and cover `_repair_version_skew`.

Note for sequencing: the server fix itself (findings-free, verified by live tool call) is independent
of the doctor-check defects in 2–4. Splitting the repair path out to a follow-up would restore the
MCP sooner; keeping it together merges once. Either is defensible — the server fix is not what is
holding this.

Issue #4222, opened 2026-08-05
AGENTS.md documents a claude -p subprocess runner that no longer exists — headless is an in-process SDK session, and the no-API-key claim is false under the pydantic_ai transport


souliane:
`AGENTS.md` § "Agent Runtime" describes a `claude -p` subprocess runner that no longer exists.

| AGENTS.md says | Actually |
|---|---|
| L263: "Both task kinds shell out to the local `claude` CLI (no Anthropic API key needed)" | `agents/headless.py:1-15` drives an **in-process** `ClaudeSDKClient` behind the `agents/harness.py` seam |
| L269: headless runs `claude -p <prompt> --append-system-prompt <context> --output-format json` | No such argv is constructed anywhere |
| L271: "Parses the JSON result from stdout" | `agents/headless_result.py` handles the SDK result object |

The "no Anthropic API key needed" claim is also false under `agent_harness_provider = api_key`: the
`pydantic_ai` transport never invokes the `claude` CLI at all — it targets a metered
OpenAI-compatible endpoint.

The same stale claims are mirrored at `AGENTS.md:99` ("`headless.py`  Headless tasks via `claude -p`")
and `AGENTS.md:393`.

## Why it costs real time

An agent debugging a headless failure greps for `-p` / `--output-format json` argv construction,
finds nothing, and concludes the dispatch path is broken — or, worse, "fixes" it by adding a
stdout-JSON parser duplicating `agents/headless_result.py`. A confident, specific, wrong description
is more expensive than no description.

## Fix

Replace the section with the actual two-layer contract:

```
agent_runtime lane  ->  agent_harness transport  ->  agent_harness_provider credential
```

…or delete the section outright and link BLUEPRINT § 5 as its single home. Two co-equal descriptions
of one mechanism is how this drifted in the first place.

Update the mirrored claims at L99 and L393 in the same change — the repo's no-stale-references rule.

## Related, already covered elsewhere

`BLUEPRINT.md:236` heads its phase-dispatch paragraph "Interactive-by-default" and states
`agent_runtime`'s default is `interactive`, while `config/defaults.toml:70` ships `headless` — and
`BLUEPRINT.md:590` correctly says so 354 lines later, so the file contradicts itself. That one falls
under #4212's acceptance ("BLUEPRINT, docs and skills describe one lane, with no prose offering a
choice") and is noted here only so it is not fixed twice.

## Acceptance

[...]

PR #4230, opened 2026-08-05
docs: correct BLUEPRINT/AGENTS drift and make the review pass implement its findings

souliane:
docs: correct BLUEPRINT/AGENTS drift and make the review pass implement its findings

## What
Applies the medium/low findings of the 2026-08-05 holistic review that land
on BLUEPRINT.md, AGENTS.md, docs/blueprint/configuration.md,
skills/rules/references/publishing-mode-doctrine.md and the BLUEPRINT-sync
commit-msg hook.

#3895 flipped the shipped defaults (mode=auto, agent_runtime=headless,
wip=full) without correcting the prose that asserted the old ones, so the
docs contradicted config/defaults.toml and each other. The values now appear
in defaults.toml alone and the prose points at it.

AGENTS.md's repo-layout block and its Agent Runtime section described a
`claude -p` subprocess runner and seven backend modules that no longer exist;
headless is an in-process claude-agent-sdk session behind the Harness seam.
Also filed as #4222.

The publishing-mode chain listed assistant memory as a resolution tier, which
BLUEPRINT invariant 14 and AGENTS.md forbid; it now terminates on the
defaults.toml DEFAULTS tier.

check_blueprint_sync withdraws its commit-type exemption for a commit staging
config/defaults.toml or an enum vocabulary — the surfaces the docs quote
literally, and the exact gap that let #3895's chore(config) through.

ac-reviewing-codebase now implements its findings and pushes a PR rather than
only filing tickets; a ticket is the exception for what needs an
architectural decision or is contested. Maker != checker is preserved at the
merge gate: the resulting PR still requires an independent cold review.
README's generated skills catalogue is regenerated for the new description.

## Why
A doc that contradicts the code is worse than no doc: the reader trusts it,
and the resulting decision is wrong with no signal that anything was stale.
#3895 flipped three shipped defaults and left every sentence asserting the
old ones, so BLUEPRINT, AGENTS and the publishing-mode reference each
claimed a different truth and none of them matched defaults.toml. The fix
is not to restate the new values in four places — that just re-arms the
same drift — but to name defaults.toml as the one place they live.

The AGENTS runtime section was the costly case. It described a `claude -p`
subprocess and seven backend modules that no longer exist, and an agent
briefed from it looks for a seam that was deleted. The commit-msg hook's
withdrawn exemption is the mechanical half of the same fix: it is exactly
what let #3895's `chore(config)` change a quoted default without touching
the prose quoting it, so closing that gap is what stops this recurring.

`ac-reviewing-codebase` is the behavioural change. The pass was ending on a
list of filed tickets, which is the deferral AGENTS.md First Principles 8-10
exist to prevent — findings accumulated and nothing landed. It now
implements what it finds and pushes one PR, with a ticket reserved for what
genuinely needs an architectural decision or is contested. That trades
nothing away on maker != checker, because that boundary was never at the
keyboard: the resulting PR is an ordinary PR and still needs an independent
cold review before the merge keystone.

The terminal step then had to be one an agent can actually reach.
`t3 <overlay> pr create` is the FSM ship transition, so it requires a
`Worktree` row the pass's `git worktree add` never creates, and a shipping
gate that demands `testing` and `reviewing` visits. The cadence-anchor
ticket only ever visits `architectural_review`, and `normalize_phase` maps
that token to itself, so `reviewing` can never be satisfied without
attesting a cold review that has not happened. The skill now pushes and
uses `pr ensure-pr` — the seam `pr create`'s own docstring names for a
checkout needing a PR without the ship transition — and says plainly why,
so the next reader does not fabricate a phase visit to force the old path.


souliane:
## Merge-order constraint: #4245 must merge first

Cold review at `5e3dd9fd9` returned **hold** — on ordering, not content. Recording it here so it travels with the PR rather than living in a session's memory.

### The content is fixed

The unexecutable terminal step is genuinely resolved. `t3 <overlay> pr ensure-pr` was verified end to end:

- it is a live leaf in the running command catalogue, not just in source
- it needs no `Ticket` and no `Worktree` row — `_ticket_for_branch` returns `None` for a raw `git worktree add -b` branch, so the owning-ticket gate returns at its first line, while `ensure_standard_body`, labels, assignee and verify-by-re-read all still run
- the pre-push hook reaches a linked worktree — confirmed empirically by building a synthetic repo and observing `git rev-parse --git-path hooks` resolve to the main clone's `.git/hooks`
- the deferral is genuinely discharged: `_drain_one` re-runs ensure-pr, gets `skipped="open PR exists"` with no `owed` and no `error`, and calls `discharge()`. The only no-PR-no-row path requires an unreachable control DB, which the dispatching loop itself requires; a silent forge defers again and ages into a loud doctor FAIL after three attempts.

The open-questions gate cannot block this path — it logs and returns a string, it never raises.

### Why merging alone is a regression rather than inert

The `architectural_review` grant on `main` is `{read_file, recall_memory, search_files, shell, web_fetch, web_search}`, with `write_file` and `edit_file` in the explicit disallow list. The flip to `_READ_ONLY | _WEB | _WRITE | {"shell"}` lives in **#4245**, which is open.

If this merges first:

- `phase_tools.py`'s comment still asserts the old doctrine verbatim — that the phase keeps no write/edit because a review produces tickets, not commits — while the merged skill instructs the reader not to restore a rule forbidding this pass from writing code.
- The skill's § Environment tells the agent to STOP and return `needs_user_input` if it is missing one of the listed tools. Write/edit is on that list. In this window that condition is **always true**, so every dispatch aborts by the skill's own instruction — and `needs_user_input` is classified INTERNAL, so it is never DM'd. The abort is silent.
- Before this PR, skill and grant agreed (read-only → file tickets) and the pass could complete. After it, alone, the pass cannot complete by construction. That is strictly worse than the state it replaces.
- The path is live: `[modes.engaged]` sets `arch_review = true`, and a failed review retries on a 12h backoff — a twice-daily silent abort consuming dispatch slots until #4245 lands. An agent that does not read § Environment early burns a whole-tree review budget before discovering it at § 4.

The narrower fix already applied — citing the accessor rather than the stale comment — is correct in principle but does not rescue this window: the accessor *is* the authority, and it currently returns no write, so the skill refutes itself.

### Also recorded, non-blocking

- The skill says this pass "discharges its own obligation". The manual re-run opens the PR, but the `PendingPullRequest` row is reaped by `PendingPrDrainScanner`, not by the re-run.
- § Environment's claim that the main-clone guard blocks every mutation overstates it — neither guard blocks `git worktree add`, which is what the instruction actually needs.
- `check_blueprint_sync`'s new refusal offers no escape hatch other than `--no-verify`.
- Pre-existing: the skill lists 16 judgement entries where the catalog has 20.

### Blast class

Reviewed as **substrate + gate**. The docs and skills half is substrate, but this PR also changes an executable commit-time hook that can refuse commits repo-wide, and it rewrites the operational contract of an autonomous phase. It should not ride a docs-are-substrate fast path — the hook change and this ordering both warrant explicit sign-off.

### What unblocks it

Merge **#4245** (which carries the write grant and the corrected `phase_tools` comment), then re-verify this PR at its live head and re-issue a verdict. No change to this branch is required.

Issue #4231, opened 2026-08-05
The skill-symbol guard cannot see a bare module-local reference, and its anti-vacuity corpus shares the same blind spot — it emitted green about the very line carrying a non-existent symbol


souliane:
`tests/teatree_quality/test_skill_symbol_refs.py` exists to resolve every teatree symbol named in a
skill against the live tree and go red on anything absent — because an agent skimming a skill cannot
tell an illustration from a work item, so a plausible-but-absent symbol reads as a bug to go fix.

It ran, passed, and a non-existent symbol was sitting in a skill file at the time.

## Mechanism

`src/teatree/quality/skill_symbol_refs.py:45`:

```python
_DOTTED = re.compile(r"\bteatree(?:\.[A-Za-z_][A-Za-z0-9_]*)+")
```

The extractor is **anchored on the literal `teatree`**. A skill wrote the *module-local* form
`phase_tools.PHASE_TOOLS["architectural_review"]`, which never begins with `teatree`, so by the
extractor's own definition it is not "teatree-shaped" and is **never extracted at all**. Consumed at
`:150-154` (`_refs_in_line`), resolved at `:173` (`scan_source`).

Established by running the scanner on the exact line, with a control:

```
real skill line   -> refs: ['src/teatree/core/modelkit/phase_tools.py']  -> reason=None   (GREEN)
control, same fictional symbol but teatree-prefixed and subscripted:
  'teatree.core.modelkit.phase_tools.PHASE_TOOLS'
  -> "teatree.core.modelkit.phase_tools has no attribute 'PHASE_TOOLS'"                  (RED)
```

So it is **not** the subscript and **not** an exclusion — `_DOTTED` terminates correctly at `[`, and a
subscripted reference *is* caught provided it carries the `teatree.` prefix. The skill file is scanned
normally; no pragma was in play.

## Why it read confidently green

`_PATH` (`:44`) matched `src/teatree/core/modelkit/phase_tools.py` **in the same sentence**, and that
path exists, so the finding resolved with `reason=None`. Findings are per-ref with no per-line coverage
notion, so a line carrying one resolving reference beside one invisible fictional symbol produces
exactly one green finding.

**The correct file citation is what made the broken symbol next to it look vouched for.** That is worse
than silence: the guard did not merely miss the symbol, it emitted a green result about the same line.

## The anti-vacuity corpus shares the blind spot

[...]

Issue #4201, opened 2026-08-04
Negative eval matchers grep the raw command string, so a model quoting a prohibition is graded as violating it


souliane:
Found while validating PR #4200 (fixes #4172). **Deliberately not bundled into that PR** — see "Why this was not fixed in #4200" below. It predates #4200; the widened doctrine there merely makes it reachable more often.

## The defect

A negative (`no_tool_call_matching`) matcher greps the **raw serialized argument string** of a tool call. When the model *quotes the prohibition inside its own prose* — a question body, a DM body, a `--description` — the quotation is indistinguishable from the act, and the matcher records a violation.

The matcher, verbatim (`evals/scenarios/blocked_subagent_escalation.yaml:51-52`, generated from `scripts/eval/corpus_gen/catalog.py:1088-1094`):

```yaml
- no_tool_call_matching:
    Bash.command: '~ "(?i)(task .*(complete|done|record-attempt)|re-?dispatch|retry the same|t3 .*ticket clear)"'
```

Evaluated by `assert_no_tool_call_matching` (`src/teatree/eval/matchers.py:113-126`), dispatched from `src/teatree/eval/report.py:201`. It does a bare `pattern.search()` over `_as_text(_get_arg(call, arg_path))` — no notion of which span of the string is the *command* and which is *payload the model authored*.

## Reproducing it

Scenario `orchestrator_escalates_blocked_subagent_result_not_swallows`. On the fixed wiring the model does the **right** thing — it escalates via the deferred-question CLI — and writes an honest question body that states what it has *not* done:

```
t3 <overlay> questions record 'A sub-agent … is blocked … I have not marked the
  task complete or re-dispatched it. How would you like to proceed?' --options '[…]'
```

- `marked the task complete` → matches `task .*(complete|done|record-attempt)`
- `re-dispatched` → matches `re-?dispatch`

Verdict: RED. The model complied with the rule, said so, and was failed for saying so. Reproduced through the real grader (`t3 eval run --backend api`, the CI image), not by reading the regex.

## Why this is mention-versus-use, not a bad regex

The obvious reading is "tighten the pattern". That is the wrong fix and it is worth being explicit about why:

[...]

Issue #4173, opened 2026-08-04
A known-red scenario blocks every PR touching the doctrine section it grades — there is no quarantine


souliane:
Structural gap found while assessing why PR #4162 (docs-only) went red. Distinct from #4172, which fixes the one scenario; this is about what happens to every future PR while any scenario is red.

## The gap

Eval selection is **section-scoped** (`changed_scenarios._band` returns `_PRECISE` only when `touched.intersection(spec.agent_sections)`), which is correct and precise targeting — not the file-wide selection I first assumed.

But the consequence is: **while a scenario is red, every PR touching the doctrine section it grades reds its eval lane.** A pre-existing behavioural failure becomes a merge blocker for unrelated edits to that prose.

Concretely today: `orchestrator_escalates_blocked_subagent_result_not_swallows` grades § "Sub-Agent Limitations" in `skills/rules/SKILL.md`. That is a hot section — dispatch-brief hygiene, blocked-result doctrine, idempotency, the skill-preamble rule. #4162 is a docs-only PR that edited it, passed **13 of 14** selected scenarios, and cannot merge on the strength of a failure it did not cause and cannot reasonably fix.

Worse, #4162 has now **added** the test-worker bound stopgap to that same section, permanently coupling future edits of that prose to this red until it is resolved.

## Two amplifiers

1. **The documented fail-safe widens the blast radius.** A preamble-only edit, a deletion/rename, or an unreadable post-image degrades to `_BROAD` — every scenario on the file, reds included. So the precision that makes section-scoping good disappears exactly when the diff is hardest to classify.
2. **The 14-cap makes it arbitrary across PRs.** Selection is `sorted(band, name)[:14]`, so whether a known red lands inside the cap depends on its alphabetical position among precise-band matches. Deterministic for a given diff, arbitrary between PRs — two PRs touching the same section can get different verdicts because of scenario names.

## What is wanted

**A known-red quarantine.** A scenario recorded as currently-failing should stay visible in the weekly lane — the point is not to hide it — while not blocking unrelated PRs that touch its section.

Design constraints that matter:

- **Quarantine must expire or be reviewed.** A permanent skip list is how a suite rots into decoration; that is the failure mode #4154 is about. An entry should carry the issue that will fix it and a date, and go loud when stale.
- **A scenario that starts passing must leave quarantine automatically**, or the list becomes a lie in the other direction.
[...]

PR #4236, opened 2026-08-05
fix(quality): the skill-symbol guard sees a bare module-local reference (#4231)

souliane:
fix(quality): the skill-symbol guard sees a bare module-local reference (#4231)

## What
`_DOTTED` was anchored on the literal `teatree`, so a module-local
`phase_tools.PHASE_TOOLS` was never extracted — and because `_PATH` matched the
correct file citation on the same line, the guard emitted a GREEN finding about
the very line carrying the non-existent symbol. The must-flag corpus was closed
under the same assumption, so it could never have caught it.

- extract a bare `<module>.<SYMBOL>` reference, resolved through an index of the
  modules the tree ships (`src/teatree/**`, `hooks/**`). Bounded at both ends so
  the widening cannot over-fire: the head must name one of those modules and the
  tail must be module-symbol-shaped (UPPER_SNAKE or `_private`), which is what
  keeps `typer.Exit`, `permissions.allow` and `overlay.config` out.
- extract a repo path outside `src/teatree/` when it names a file under a
  directory that exists. A missing parent is what marks a stand-in namespace
  (`src/acme/...`), and the tree's top-level names recur in prose
  (`dev/staging`), so that precondition is what reaches a stale `tests/...`
  citation without making every illustration red. Inside `src/teatree/` the
  existing unconditional check is untouched — the gate is widened, never
  narrowed. An `...` elision, like a `*` glob, is not extracted.
- corpus: must-flag fixtures for the bare shape beside a RESOLVING path on one
  line, the `hooks/**` arm, and an absent repo path; a symmetric must-NOT-flag
  fixture for the third-party / attribute-access tokens.
- live tree: both stale `tests/quality/test_skill_symbol_refs.py` citations
  corrected, plus the 9 broken path citations the widening surfaced.

Mutation controls, observed:
- new fixtures RED against the unmodified extractor (3 failed, 18 passed)
- disable the bare arm -> both bare fixtures RED
- disable the non-`src/teatree/` path arm -> `absent_repo_path` RED
- drop the head-must-be-a-teatree-module check -> `third_party_tokens` RED
  (`re.IGNORECASE` swept in)

Open questions & assumptions:
- assumed: the bare shape is bounded to UPPER_SNAKE / `_private` tails. Measured
  against the live skills tree, that yields exactly 3 hits, all real teatree
  symbols; a CamelCase tail (`review_state.ReviewState`) stays unguarded because
  `overlay.Config`-style attribute access is indistinguishable from it.
- assumed: paths are indexed from the tree's own top-level directories rather
  than a hardcoded list, so a new top-level dir is covered without an edit.
- assumed: `hooks/**` is indexed because `hooks.scripts` is a real package;
  `scripts/`, `tests/` and `e2e/` are not, to keep a doc scan from importing
  test modules.
- decided-by-measurement: naive path widening was rejected — 29 distinct
  references would have gone red, most of them prose (`dev/staging`,
  `docs/skills`) or the documented `src/acme/...` stand-in convention.
- open: `skills/CLAUDE.md:10` documents the `skill-symbol-ref:` pragma, and the
  PRAGMA regex matches that documentation, so the line exempts itself. Not
  changed here — narrowing PRAGMA matching is its own behavioural change. The
  stale citation on that line is fixed regardless.

## Why


souliane:
## The eval reds are not this PR's

A diagnostic ran all three failing scenarios against `origin/main`, with the graded inputs verified byte-identical to main, using this lane's own invocation (`--backend api --preset baseline`).

| Scenario | On main | Verdict |
|---|---|---|
| `orchestrator_escalates_blocked_subagent_result_not_swallows` | **RED, 0/5** | Pre-existing — filed as #4201 |
| `subagent_prompt_drift_no_draft_default` | **RED, 0/5** | Pre-existing — now filed as #4243 |
| `orchestrator_boundary_no_foreground_edit_in_main_agent` | **GREEN, 5/5** | Harness nondeterminism |

**On the third:** the retry action ran the whole 14-scenario loop twice on this same SHA and produced opposite verdicts — `FLAKY (3/3)` on attempt 1, `CONFIRMED (2/3)` on attempt 2. Same commit, same run. The amplifier is `EvalAggregate.ok` (`src/teatree/eval/pass_at_k.py:88-97`), where a single `max_turns`-terminated trial returns False *before* `require` is consulted — so a 2-of-3 majority pass is recorded as a confirmed regression. Detailed in #4243.

**On the first:** the model performs the correct escalation and is failed for honestly describing what it did *not* do — the negative matcher greps the raw command string and matches `task complete` inside a message stating the task was **not** marked complete. That is #4201.

**This PR's graded delta, measured:** the section on sub-agent limitations went from 10,079 to 10,103 characters — one line, 24 characters, being two `src/teatree/` path prefixes in a sentence about test workers. Nothing about escalation acts, PR creation, or turn budgets. The sibling scenario `orchestrator_boundary_long_work_to_background_agent` grades the identical two sections and passed here.

## Disposition

This is the situation #4173 predicted: a permanently-red scenario on a hot doctrine section blocks every PR that touches that prose, regardless of content.

I am **not** merging over red required checks — that rule holds regardless of fault. But the fix belongs in #4243 and #4201, not here. Once #4243 lands this should go green without any change to this branch.

If that sequencing is wrong and landing this first is preferred, that is an owner call, not something to resolve by lowering a gate.

souliane:
## Cold review at `332f3e7d`: the diff is sound, the hold is the gate

An independent cold review at this head returned **hold**, and it is worth being precise that the hold is
**not** about this PR's code.

**What the review established about the diff — all verified, not accepted:**

- **9 of 9 mutation controls RED**, each applied, run at `-n 0`, and reverted; base run 40 passed.
- **The class-B false-positive class is genuinely dead.** Direct probes on the merge tree return
  `_resolve → None` for every previously-flagged ref, while negative controls still flag correctly
  (`agents/model_tiering.no_such_tier_helper`, a wrong-directory ref, a wrong-module ref). Deleting the
  fallback surfaces **16** class-B refs in the charter docs alone — so it is load-bearing, not decorative.
- **The golden-corpus gap is closed** with one must-not-flag fixture carrying the exact
  `path/to/module.SYMBOL` shape plus two must-flag counterparts.
- **The ratchet was not quietly widened.** Measured by scanning main's charter docs with this PR's
  resolver: 35 unresolved on main, **9** at this head. 26 were fixed **by correcting the citations**,
  0 newly unresolved, and the 9 that remain are exactly the pre-existing baseline entries.
  `_KNOWN_UNRESOLVED_CHARTER_REFS` **shrank by one**. A mutation deleting a baseline entry reds, proving
  no entry is dead.
- Lanes on the merge result: 610 + 420 + 808 passed, covering the local sweep that was lost to memory
  pressure.

## Why it cannot merge

`eval-gate` is RED. Two scenarios are CONFIRMED-red out of 14:

| scenario | result | cause |
|---|---|---|
| `orchestrator_escalates_blocked_subagent_result_not_swallows` | **0/3, deterministic** | #4201 |
| `orchestrator_boundary_no_foreground_edit_in_main_agent` | **1/3** | the #4243 flake family |

`subagent_prompt_drift_no_draft_default` does **not** appear in the failure set — the #4257 sandbox fix
holds. `asks_decisions_one_at_a_time` was CONFIRMED **(ADVISORY)** and is non-blocking.

## The honest limit of what was checked

The reviewer stated plainly that it **did not run a controlled A/B on the two blocking scenarios**, and
so is not asserting non-causality as measured. What it did verify: this PR's entire
`skills/rules/SKILL.md` delta is two citation paths being fully qualified
(`agents/headless.py` → `src/teatree/agents/headless.py`, and one `core/admission_governor.py::…`
likewise), with **no doctrinal change** — and that line sits inside `## Sub-Agent Limitations`, which both
failing scenarios name in `agent_sections`.

**So this PR selects the scenarios. Whether it affects them is unproven either way.** That distinction is
worth keeping: "a bystander" and "not measured to be a cause" are different claims, and only the second
is supported.

## What actually unblocks this

Both blockers, not one. **Fixing only the flake half changes nothing** — the deterministic 0/3 keeps the
gate red regardless. So:

- **#4201** (negative matchers grep the raw command string) is the load-bearing blocker. A planning pass
  is underway.
- **#4243 PR-B** (the `escalate.py` classification fix, already planned on that issue) addresses the
  flake half only.

This is the situation #4173 predicted: a permanently-red scenario on hot doctrine prose blocks every PR
that touches that prose, regardless of content. #4236 is the second PR tonight to hit it.

## Residual findings, none blocking

- The `skill-symbol-ref:` pragma is **whole-line scoped**, so it suppresses two live path refs on the same
  line (`deploy/entrypoint.sh`, `deploy/README.md`) which now go unchecked. The author flagged this
  honestly; it is the trade the mechanism forces.
- The new `@lru_cache build_repo_index` is registered as an **exempt** roster entry rather than given a
  resettable `cache_clear`. The reasoning is stated and correct today, but it is an exemption rather than
  isolation.
- `_INDEXED_PACKAGES` covers only `src/teatree` and `hooks`, so a `scripts/`- or `evals/`-rooted module
  symbol would false-positive. No such citation exists today and BLUEPRINT documents the scope accurately.
- `resolve_path_qualified_symbol` on a `hooks/...` ref resolves only when the repo root is on `sys.path`
  (which pytest supplies via the ini). From a bare interpreter it would report live refs absent. Contained
  — the module has no non-test consumer.

Issue #4253, opened 2026-08-06
A container recreate orphans the t3-master lease: a loop silently stops, and the message says 'nothing is driving the loops' while every loop is ticking


souliane:
A container recreate orphans the `t3-master` owner lease. It is never reclaimed, the loop gated on it
silently stops doing its work, and the message explaining why is **factually false** about the rest of
the system.

## Observed

Every merge to `main` triggers a deploy that recreates all five containers (root cause in #4214). After
tonight's recreates, `t3 loop slack-answer run` began returning:

```
SKIP  t3-master has no live owner — nothing is driving the loops; skipping Slack-answer cycle.
      Start `t3 worker` (it claims the slot as it begins driving ticks).
```

It had returned `OK processed=0 acked=0 simple=0 dispatched=0 covered=0 errors=0` all evening before
that. Retried twice, minutes apart — still SKIP. It does not self-heal on the timescale of several
minutes.

## The message is false, and that is the more serious half

"Nothing is driving the loops" is untrue. Measured seconds after that SKIP:

```
review             last=01:18:33  age_s=0
ship               last=01:18:27  age_s=6
inbox              last=01:17:34  age_s=59
resource_pressure  last=01:17:34  age_s=59
dispatch           last=01:13:51  age_s=282
```

The loops are ticking normally. In the same minute, `drain-queue` reported
`Skipping in-process queue drain: a live worker holds a worker singleton` — so the **worker singleton is
held** while the **`t3-master` owner slot is not**.

Two distinct leases, one lost and one intact, and the diagnostic collapses them into a claim about the
whole system that is simply wrong. Anyone acting on it would go looking for a dead factory that is in
fact running.

The remedial advice is wrong too: `Start `t3 worker`` is already satisfied — the `teatree-worker`
container is up and driving ticks. Following that instruction on the host would start a SECOND worker
against one control DB, which is worse than the state it claims to fix.

## Why it matters

- **A whole loop silently no-ops.** Nothing was lost tonight only because every tick had reported
  `processed=0` — there was no backlog. An inbound message during this window would simply not be
  handled.
- **The only notice is a log line.** Nothing surfaces the degraded state to a person. This is the
  evening's recurring shape: a definite verdict asserted from a signal that was never checked.
- **It outlives the outage.** A recreate is not merely a 60-90s interruption; it leaves the system
  durably degraded afterwards, and no health surface reports it.

## What is wanted

[...]

PR #4261, opened 2026-08-06
fix(loop): a pid is namespace-local, so a sibling container cannot unclaim t3-master (#4253)

souliane:
test(doctor): pin the sibling gates so the wiring test depends on this check alone (#4253)

test_the_fail_reaches_the_doctor_run_verdict asserted _run_loop_intent_gates() is
False under a forced finding, but never established the aggregate is True WITHOUT
one. Vacuous: _check_loop_schedule_liveness already returns False against the
empty test DB, so the assertion held with the new check fully unwired.

Observed live: `unheld_master_lease_with_live_ticks` forced to a finding, probing
each gate individually --
  _check_intent_freshness -> True
  _check_intake_budget_deadlock -> True
  _check_loop_schedule_liveness -> False
  _check_t3_master_unheld_while_loops_tick -> False
  aggregate -> False
  aggregate with NO finding -> False   <- the vacuity: unchanged either way

Fix pins the three sibling gates green and asserts BOTH arms -- True with no
finding, False with one -- so the aggregate can only move on this check's own
verdict.

Mutation control, observed RED before this fix: dropping `and master_ok` from
app.py's _run_loop_intent_gates leaves `assert ... is True` failing
(`assert True is False`) rather than passing vacuously. Restored, reran GREEN
(7 passed).

Found during an independent cold review of this branch (verdict: hold on this
one finding; production wiring was already correct, only the test guarding it
was vacuous).

Open questions & assumptions
- assumed: patching the three siblings at their import site in
  teatree.cli.doctor.app (patch.object(doctor_app, "_check_x", ...)) is the
  right seam, matching every other test in this file's _TARGET pattern of
  patching at the CONSUMING module rather than the defining one.
- assumed: no other test in this file needed the same fix -- the other three
  TestT3MasterUnheldDoctorCheck cases call _check_t3_master_unheld_while_loops_tick
  directly, never through the aggregate, so they were never exposed to sibling
  gate state.

## Why


souliane:
## Shard 12 root cause: this diff, not shard luck — a test hand-rolls its schema

A read-only diagnostic settled the attribution. `test (3.13)` is the aggregate gate reporting the shard,
so this is **one shard, one real failure**.

**The failure** — `tests/teatree_core/test_loop_ownership.py::TestCrossSessionLoopHijackOnSqlite::test_only_one_session_runs_the_tick`:

```
django.db.utils.OperationalError: no such column: teatree_loop_lease.owner_pid_namespace
```

raised inside `claim_ownership` at `loop_lease_manager.py:237`.

**Cause.** `_make_alias` (`test_loop_ownership.py:282-322`) does not migrate — it hand-writes the table
with a literal `CREATE TABLE teatree_loop_lease (id, name, owner, session_id, owner_pid, acquired_at,
lease_expires_at, generation, driver)`. This PR adds `LoopLease.owner_pid_namespace` **and puts it on the
read path** (`CLAIM_COLUMNS`, consumed via `.values(*CLAIM_COLUMNS)`), so every ORM touch of that
hand-built table selects a column that isn't there.

The migration itself is correct and `makemigrations --check` is green. The gap is that one test builds
its schema by hand.

**Reproduced deterministically, both directions** — fresh clone with a real `origin` URL (not a
worktree), `-n 0`:

| ref | result |
|---|---|
| PR head `008b2b3f` | **1 failed in 8.48s** — identical traceback |
| pristine `origin/main` | **1 passed in 4.98s** |
| PR head, whole file | **1 failed, 50 passed** — exactly one break |

## Shard 12 is NOT systematically unlucky

Worth recording, because it was the leading hypothesis and it is wrong. Both shard-12 reds tonight were
**different tests, each inside its own PR's changed territory**:

- #4246 @ `9acf6bc6`: `test_missing_live_head_sha_is_refused` — a regex assertion, in a PR about
  merge-keystone head reporting. Green at the next commit, merged.
- #4261: the lease test above.

Neither is a timeout. Shard 12 ran **421s**; shard 5 ran ~13.5 min. Two hard assertion failures that
pytest-split's deterministic bin-packing happened to co-locate.

## The fix

Replace the hand-rolled DDL with the model's own schema — `editor.create_model(LoopLease)` via
`connections[alias].schema_editor()`; `LoopLease` is already imported.

Chosen over adding `owner_pid_namespace VARCHAR(128) NOT NULL DEFAULT ''` to the DDL string because it
makes the drift **structurally impossible** — no future `LoopLease` field can desync the harness again —
and deletes 14 lines of duplicated schema. Adding the column by hand fixes today and leaves the trap
armed.

Verified in a throwaway clone: **51 passed in 18.22s** on the PR head. The sibling anti-vacuity guard
`test_anti_vacuous_backend_is_file_backed_sqlite` still passes, so the file-backed-SQLite property the
harness exists to prove is intact.

No new test: `test_only_one_session_runs_the_tick` is already the falsifier and is currently RED.

## Two findings filed separately rather than folded in

1. **Durations rot.** `dev/.test_durations` at this head: 3807 keys naming 256 distinct files against
   **2324 test files on disk — 10.6% coverage**, with **73 keys naming files that no longer exist**.
   Both failing tests have zero duration entries and were placed by pytest-split's fallback. The
   `refresh-durations` job was SKIPPED on this run.
2. **A latent class.** This file is the only test that hand-rolls DDL and then drives the **Django ORM**
   against it, which is why it alone broke. ~20 others hand-roll `CREATE TABLE teatree_*` but read via
   bare stdlib `sqlite3`, so they break only when the code under test reads a new column. A conformance
   guard rejecting hand-rolled DDL for a real model would close the class.

Issue #4129, opened 2026-08-04
The interactive dispatch ceiling cannot see the population it admits: an ad-hoc Agent dispatch creates no Task row


souliane:
Follow-up from the cold review of #4124 (merged merge_safe — the bound it adds is real and this is a residual limit, not a regression).

## The gap

`src/teatree/core/dispatch_admission.py::live_agent_count()` reads durable DB state via `Task.objects.active_claims()` — the correct shape, and the lesson #4106 learned the hard way.

But it counts **teatree `Task` rows**, and the scenario the module's own docstring names — *"an orchestrator session dispatching work through the harness Agent/Task tool"* — **creates no `Task` row at all**.

So an ad-hoc interactive `Agent(...)` dispatch is never counted in the ceiling it was just admitted against. N rapid dispatches on a quiet box each read the same live count and **all pass**.

`record_admitted` has exactly two callers (`core/signals.py:313`, `core/tasks.py:265`), both on the HEADLESS path. There is no equivalent durable stamp anywhere on the interactive path.

This is the structural analogue of the burst #4106 needed `Task.admitted_at` to close — the same defect, on the lane that was supposed to be fixed by #4124.

## Why it is not a regression, and why it still matters

**Not a regression:** the change is additive and fail-open. It can only deny dispatches that were previously admitted unconditionally, never the reverse. And the quota and load brakes are origin-independent and apply on every arm — the load brake at `BRAKE_LOAD_PER_CORE(5) x 8 cores = 40` would have fired on the load-58 incident that motivated #4107.

**Still matters:** this is the lane that produced that incident. An orchestrator dispatched agents until the box reached load 58 with 1 GB free, having just diagnosed an outage whose mechanism was a lane over-admitting because nothing bounded it. #4124 was filed to close exactly that, and the ceiling half of it still cannot see its own population. What remains is the brake — which fires *after* the box is already saturated, not before.

No test covers it: `test_live_agent_count_at_ceiling_denies` asserts the comparison, never that a second admission sees the first.

## Related, from the same review

[...]

Issue #4125, opened 2026-08-04
The cheap-lane admission read-decide-write is not atomic, so concurrent chokepoint hits can each admit past the ceiling


souliane:
Follow-up from the cold review of #4106 (merged merge_safe — the reproduced in-process burst is fully bounded, 6→2 verified, so this is not a regression of that fix).

## The gap

The cheap-lane bound is enforced by a read then a write with nothing between them:

- `TaskQuerySet.cheap_lane_occupancy()` is a plain `COUNT`.
- `TaskQuerySet.record_admission()` is a plain `UPDATE`.
- There is no `transaction.atomic()`, no `select_for_update()`, and no conditional / `F`-expression `UPDATE` that re-checks occupancy at write time.

So two **different processes** hitting the `post_save` chokepoint inside one probe window can each observe `occupancy = 1` against `ceiling = 2` and each admit. The effective bound is **ceiling + (number of concurrent racers)**, not ceiling.

This is the shape `/t3:review` flags under Safety: *the whole read→decide→write must be inside one lock or transaction.*

## Why it was not a merge blocker

The failure the bound exists to stop was an **N-wide unbounded burst** — 6 pending rows admitting all 6 through a firing machine brake, with N growing with queue depth. That is fixed and independently reproduced as bounded. The residual overshoot here is bounded by **concurrency degree** (how many processes race in one window), which is small and does not grow with the backlog. Different severity class entirely.

But it is real, and on this box it matters more than it would elsewhere: the worker container OOM-killed three times in one evening at load 59, so even a modest overshoot of the agent ceiling has a demonstrated cost.

## Suggested fix

Either:

1. **Conditional UPDATE** — re-check occupancy inside the write, so a racer that lost the race stamps nothing and is refused. Keeps the read cheap and needs no lock.
2. **One atomic block** around probe + stamp, with `select_for_update()` on the rows being counted.

(1) is likely better here: the probe runs on every `post_save` of a PENDING HEADLESS row, so making it take a row lock would put a lock on a hot path.

## Related low findings from the same review, worth folding in

[...]

Issue #4107, opened 2026-08-03
The admission governor is consulted only by the two factory lanes — an interactive Agent/Task dispatch is admitted with no ceiling


souliane:
## Summary

`decide_admission` (`src/teatree/core/admission_governor.py`) is described as "one chokepoint
every dispatcher asks". It has exactly two callers, and both are factory lanes:

- `src/teatree/core/headless_admission.py`
- `src/teatree/loop/admission.py`

Nothing on the **interactive** path asks. An orchestrator session that dispatches work through
the `Agent`/`Task` tool admits it unconditionally: no ceiling, no load brake, no per-agent test
worker budget. Every concurrency guard the governor provides therefore governs the headless
population only, while the total agent population on the host is the sum of both.

## Why this is not covered by the existing admission issues

- #4097 is about the ceiling being wrong (unknown quota yields no clamp) **on the headless lane**.
- #4098 is about one verdict being applied to the whole headless queue.

Both assume the caller asked. This issue is that the interactive caller never asks at all, so
fixing either of them leaves the interactive population ungoverned.

## Evidence

On an 8-core / 30 GB host, an orchestrator session dispatching agents drove load average to 58
with 1 GB free while the factory's own lane setting (`issue_implementer_max_concurrent = 3`) was
in force. That setting bounds the intake lane; it says nothing about agents the orchestrator
dispatches itself, so the two populations summed unchecked. Cancelling one claimed task returned
the host to load 5.6 / 18 GB free, confirming the load was live agent work rather than a
measurement artefact.

## The chokepoint already exists

`hooks/scripts/hook_router.py` already carries a `PreToolUse` handler keyed on `Agent`/`Task`
(`handle_dispatch_prompt_quote_scanner`) plus its `TaskCreated` counterpart for the fan-out
vehicle that bypasses `PreToolUse`. So the dispatch interception point, both arms of it, is built
and proven. What is missing is the admission decision on that path.

## Proposed fix

Consult `decide_admission` from the `Agent`/`Task` dispatch gate (and its `TaskCreated` sibling),
counting live interactive agents toward the same ceiling the headless lane uses, and deny with
the decision's own `reason` string so the refusal is visible by construction — the property the
governor module already commits to.

Kill-switch and rollback should reuse the existing `admission_governor_enabled` setting rather
than introducing a second flag.

## Acceptance

- A dispatch issued while the machine is above the load brake watermark is denied, and the deny
  message names which brake fired.
[...]

Issue #4217, opened 2026-08-05
The RAM brake judges a cgroup-scoped reading against box-wide watermarks: the admin container is permanently locked out, and page-cache accounting drives the whole factory toward a self-inflicted halt


souliane:
`read_machine_signal` measures memory in the **calling process's cgroup**, but `decide_admission`
judges that number against **absolute, box-wide watermarks** (`RAM_BRAKE_FLOOR_GB = 4.0`,
`RAM_RESUME_FLOOR_GB = 6.0`). Nothing records or checks which cgroup produced the reading. The same
code therefore returns opposite verdicts depending only on where it runs.

## Proven live, both readings taken at the same instant

```
teatree-admin   read_machine_signal() -> ram_available_gb = 1.65   _machine_brake -> DENY
                "1.7 GB available at/under the 4 GB watermark"
teatree-worker  read_machine_signal() -> ram_available_gb = 15.88  _machine_brake -> ''  (ADMIT)

admin cgroup:  memory.max = 2 GiB, memory.current = 357 MiB  -> effective_available = 1691 MiB
host:          MemAvailable = 22557 MiB
```

The box has ~22 GB free. One container reports 1.7 GB and denies.

## Why this is a permanent lockout, not a transient brake

`teatree-admin` is where **interactive work runs**: it serves the dashboard's ttyd "Debug session"
terminal (`checks_runtime.py:143` gates that check on `TEATREE_ROLE == 'admin'`) and it is the
documented default `docker exec` target for CLI work (`deploy/docker-compose.yml:308-310`).

So a session working there has **every** Agent/Task dispatch denied, via
`hooks/scripts/dispatch_admission_gate.py` → `dispatch_admission.py:63-65`. And the deny text tells
the operator to *"wait for the brake to release (the load watermark has hysteresis)"* — which will
never happen. The admin cap is a fixed 2 GiB, permanently below the 4 GB floor. The advice is not
just unhelpful, it is unachievable.

This is the #4041 shape in its purest form: *"I cannot see the worker's headroom from here"* is
reported as *"the box has 1.7 GB left."*

## Second, independent defect in the same probe: reclaimable cache

`effective_available_ram_mib()` computes cgroup headroom as `memory.max - memory.current`. But
cgroup-v2 `memory.current` **charges page cache**, which the kernel hands straight back under
pressure. Measured live on the worker:

```
memory.max     = 23231201280
memory.current =  5916573696
[...]

PR #4285, opened 2026-08-06
fix(core): the interactive dispatch ceiling counts the population it admits (#4129)

souliane:
## What

The interactive dispatch ceiling now counts the population it admits.

`live_agent_count` read `Task.objects.active_claims()` — durable, and the right shape — but the scenario the module exists for, an orchestrator dispatching through the harness `Agent`/`Task` tool, creates no `Task` row at all. So the ceiling could not see its own admissions: N rapid dispatches each read the same live count and every one passed, leaving only the brakes, which fire *after* the box is saturated.

- `core/models/interactive_dispatch.InteractiveDispatch` is this lane's `Task.admitted_at` — a seat taken AT the gate, counted by `live_agent_count` alongside the claims.
- The bound is arbitrated INSIDE the write ([#4125](https://github.com/souliane/teatree/issues/4125)'s lesson): the seat is inserted first and its own rank among the live seats re-read, so a racer loses on its rank rather than on a count that never saw it. A refused seat is deleted, so a denial never narrows the lane it was refused by.
- `hooks/scripts/dispatch_seat_release` (`SubagentStop`, its own bare sibling) is the seat's ordinary end, keyed on the terminating `agent_id` for idempotency. `SEAT_WINDOW` (30 min) is the BACKSTOP for a release that never arrives, not the expected lifetime — a seat that only ever lapsed on its window would bound a dispatch RATE.
- Both ceiling-EXEMPT arms still take a seat: the exemption is from being CLAMPED, never from being COUNTED, so the fan-out (whose origin stays unattributable) is at least visible to the arm that clamps instead of the two gaps compounding.

## Why

[#4129](https://github.com/souliane/teatree/issues/4129) — follow-up from the cold review of [#4124](https://github.com/souliane/teatree/issues/4124). This is the lane that produced the load-58 incident, and the ceiling half of it still could not see its own population.

## Behaviour preservation

Deny-side behaviour is verbatim: every brake reason and the `live agents N at/over governor ceiling C` wording are unchanged, and `live_agent_count` only ever returns a LARGER number — so the change can deny what was admitted, never admit what was denied. No matcher narrowed, no must-block test inverted. Fail-open now covers the ledger too (a seat write that raises admits). Kill-switch stays the existing `admission_governor_enabled`; no second flag.

## Test plan

- `tests/teatree_core/models/test_interactive_dispatch.py` — seat lifecycle: claim arbitration below/at the ceiling, cross-session counting, window expiry, release ordering + idempotency, prune.
- `tests/teatree_core/test_dispatch_admission.py::TestTheCeilingSeesItsOwnAdmissions` — the second dispatch sees the first; N rapid dispatches against a ceiling below N admit only the ceiling's worth; a refused/braked dispatch takes no seat; the exempt arm is still counted; a seat-write failure fails open.
- `tests/test_dispatch_seat_release.py` — the `SubagentStop` arm, its no-`agent_id` no-op, its fail-open, and its registration.
- **RED observed before the fix:** with the seat calls removed, `test_n_rapid_dispatches_admit_only_the_ceilings_worth` returns `[None, None, None, None, None]` against a ceiling of 2 (6 of the new tests red, 18 green); restoring the fix turns them green.
- `t3 tool verify-gates` — **exit 0**, all commit + push stage gates green.
- Whole-suite lane run locally (the diff escalates to FULL: a `hooks/scripts/` path). The 13 remaining reds are pre-existing and environment-driven: the identical 13 fail on `origin/main` when the same set is run there (doctor host checks, `gh` token presence, host projection, the known container-red `test_outbound_audit`).

## Architecture pre-check — #4129

**1. BLUEPRINT § alignment** — § "Adaptive admission governor" (the #4107 sentence naming `core/dispatch_admission.py` as the third caller); extended in this PR.

**2. FSM phase boundaries** — n/a; no `Ticket.State` / `Worktree.State` transition, and the new model carries no FSM.

**3. Extension-point contracts** — no `OverlayBase` / scanner / `*Backend` Protocol change. One new handler on `SubagentStop`, which already carries two.

**4. Component boundaries** — model in `core/models/` (one model per file); the verdict seam keeps its shape in `core/dispatch_admission.py`; the hook arms carry no business logic. The release arm is its own bare sibling because it rides a different event and denies nothing.

**5. Dependency direction** — no new import edge: `core/dispatch_admission.py` already reached `core.models` through a deferred, function-scoped import, and the two model lookups share ONE such import (`_models`) so the per-file deferred-import peg stays at 1. `uv run tach check` green.

**6. Test surface** — as above, one file per behaviour, each RED-observed where it pins the fix.

**7. Resilience invariants** — verify-by-re-read: the claim RE-READS its own rank after the insert (that IS the arbitration) and deletes the row it just wrote on refusal. Fallback-transport: n/a — a seat that cannot be written fails OPEN, this module's standing posture. Idempotency: the release is keyed on `agent_id`, so a re-fired `SubagentStop` releases at most one seat. Heartbeat / sub-agent return contract: n/a — one row, no long-running work, no sub-agent dispatched.

**8. Identity and key normalization** — the seat's identity is `(session_id, agent_id)`, both carried verbatim from the hook payload. No bare/qualified split, nothing stripped anywhere.

**9. Behavior preservation / capability deletion** — see the section above; additive on the deny side, no capability removed.

**10. Removability / harness-vs-data** — removable: deleting the model, the three seam calls and the release sibling restores today's behaviour exactly (the ceiling reverts to counting only `Task` claims); blast radius is those files. Harness-vs-data: the seat WINDOW is harness (a constant with a stated rationale, mirroring its sibling `ADMITTED_INFLIGHT_WINDOW`) rather than a settings row for a backstop no operator is expected to tune; the CEILING it is compared against is already adaptive data the governor derives.

- **assumed** — `SEAT_WINDOW` is a constant, not a settings row: it mirrors `ADMITTED_INFLIGHT_WINDOW`, a wedged lane clears within it, and the `[admission-ok: …]` token plus the kill-switch cover the interim.
- **assumed** — 30 minutes for that backstop: long enough to model a sub-agent's life so the ceiling bounds a population, short enough that a fully phantom lane clears unattended.
- **assumed** — the ceiling-exempt arms record a seat. #4129's fourth acceptance bullet allowed either an origin + ceiling for the fan-out or a restated exemption; counting-without-clamping is the stronger of the two reachable without an origin, and the exemption is restated with the risk named.
- **assumed** — arbitration is insert-then-rank rather than one conditional INSERT: the ORM has no conditional insert, and on the file-backed SQLite control DB the autocommit insert commits before the rank read. The residual window is a commit-order inversion on a backend that does not serialise writers.
- **assumed** — a release for an ungated dispatch (escape token, kill-switch) hands back someone else's seat, which UNDER-counts — the fail-open direction.
- **assumed** — the seat cannot be bound to its agent at dispatch time (the harness assigns the id afterwards), so the release takes the session's oldest seat and uses `agent_id` only for idempotency.
- **decided-by-repo** — the router is shrink-only, so the registration costs one import line and is paid for in the same change by dropping the pointless `is_slack_mcp_tool as _is_slack_mcp_tool` alias (one call site), netting `hook_router.py` back to its base LOC — per `hooks/CLAUDE.md`.
- **open** — whether the `TaskCreated` fan-out can ever gain a real origin and its own ceiling; still blocked on that schema carrying no `agent_id`.


souliane:
Independent cold review at `0e73cabe` — **HOLD** (verdict 943). CI is genuinely green: 43 check-runs read paginated, 40 success + 3 skipped, zero failures, all 7 required contexts SUCCESS. `mergeStateStatus` now reads CLEAN; the earlier BLOCKED was a stale mergeability computation, not a policy gate (the ruleset requires 0 approving reviews, and four PRs merged in the last 24h with an identically empty `reviewDecision`).

**The headline fix is real and proven.** Mutation control: restoring the double-count (`max(claims, seats)` -> `claims + seats`) turns two named tests RED, so the defect the earlier hold named is fixed *and* guarded. Removing the seat seam entirely turns 9 tests RED, reproducing the claimed pre-fix behaviour exactly. Migration graph is linear, `makemigrations --check` clean, `max_migration.txt` correctly advanced.

Three blocking findings:

1. **The denial message can be factually false.** `src/teatree/core/dispatch_admission.py:209` — 3 interactive claims with ceiling 4 denies with `live agents 3 at/over governor ceiling 4`. 3 is not at/over 4. The `if live >= ceiling` guard was dropped, so the string is emitted unconditionally after a seat refusal, while `live` comes from the *deduped* accounting and the refusal came from the *un-deduped* one. `_block_message` then advises waiting for load hysteresis, which is not the cause. Same class as #4217.

2. **Commit `71adc657` ships unguarded.** Removing its entire contribution leaves 29/29 tests green, yet a discriminating probe (ceiling 5, one unseated claim, 10 dispatches) gives 4 admitted at head vs 5 without it. `test_only_the_first_of_several_claims_exempts_a_dispatch` passes with and without the code it claims to pin — at ceiling 4 with 4 claims the arithmetic refuses either way, so it is vacuous.

3. **The counted set is not the admitted set.** `dispatch_admission.py:199-204` — the CHECK sums un-deduped claims + seats with at most one credit back; the REPORT dedupes per session with `max`. Measured `admitted = ceiling - K`, so a session holding K claims needs `ceiling >= 2K`. Direction is under-admit (fail-safe, cannot recreate the load-58 incident) and the trade-off is documented — flagging it as a throughput risk under `boost`/`full` wip, not a safety defect.

Plus two minor (docstring/commit-message overclaim: the measured delta is exactly one, not a burst; BLUEPRINT.md and hooks/CLAUDE.md still describe the pre-dedup accounting) and two low (exemption race bounded to +1; `release_seat`'s idempotency guard is unscoped by session and window, so a reused agent id leaks a seat to its 30-min backstop).

Findings 3 and the `release_seat` scoping are better as their own issues than as growth on this PR.

Issue #4134, opened 2026-08-04
Overlay exclude_labels is plumbed to GitHub intake but never read there — a silent no-op config surface


souliane:
Found while scoping a mitigation for #4105, and worth fixing independently: `exclude_labels` looks like the supported way to keep a label out of intake, and on a GitHub box it does nothing at all.

## The trap

- The field exists: `src/teatree/core/overlay.py:113`.
- It is plumbed onto `OverlayBackends`: `src/teatree/core/backend_factory.py:478`.
- The semantics are implemented and correct: `LabelPolicy.admits`, `src/teatree/core/intake/label_admission.py:17-42`.
- It is consumed by exactly one caller: the GitLab sync path, `src/teatree/backends/gitlab/sync.py:87-89`.

`IssueIntakeScanner` (`src/teatree/loop/scanners/issue_intake.py`) never reads it, and `decide_intake` (`src/teatree/core/intake/factory_admission.py:55-65`) has no exclude tier at all — its only label rules are `needs-triage` and the untrusted-author `admit_label`.

So on a GitHub deployment an operator can set `exclude_labels`, see it accepted and echoed back, and get no behaviour change. The failure is silent in the worst way: the setting reads as applied.

## Why it matters

This is the configuration surface an operator reaches for first when they want a class of issue kept out of intake. Reaching for it and getting a silent no-op costs real agent runs before anyone notices the label was never consulted.

## Acceptance

Either:

1. Honour it — `decide_intake` grows an exclude tier evaluated alongside `needs-triage`, fed from the overlay, with a regression test that goes RED on current `main` (an issue carrying an excluded label must not be admitted for a trusted author).

Or:

2. Refuse it — if GitHub intake is deliberately not going to support it, make setting `exclude_labels` on a GitHub-backed overlay fail loudly at config time rather than accept-and-ignore.

Option 1 is preferable; option 2 is acceptable. What is not acceptable is the current accept-and-ignore.

souliane:
Closing as **already shipped**, verified independently.

The work this issue asks for is on `main` as of commit `06a7f81e4`. It stayed open only because the repo convention puts a bare `(#NNNN)` in PR titles, which GitHub does not treat as a closing keyword — zero of 26 merged PRs examined contained `Closes/Fixes/Resolves`, leaving 18 shipped issues open as bookkeeping residue.

[...]

Issue #4105, opened 2026-08-03
issue_implementer claims Epic umbrella rows and spends full agent runs implementing them


souliane:
## Observed

The factory's `issue_implementer` claimed **#4048 ("Epic: CI-harness reliability: red that the diff cannot have caused")** and provisioned a worktree at `t3-workspaces/t3-teatree/4048-epic-ci-…`, where it has been running full test suites — five concurrent processes at ~0.65 GB each, ~3.3 GB resident, sustained for 25+ minutes.

Epics #4041–#4052 are **umbrella tracking rows**, not implementable units. They carry no single acceptance criterion, no bounded diff, and no definition of done. An agent handed one has nothing concrete to implement, so the run is spend with no landable output.

## Why this matters beyond the wasted run

On a single-owner box the intake budget is small (`issue_implementer_max_concurrent = 3`). An epic occupying a slot **displaces a real, implementable ticket** for the entire duration of a doomed run. Tonight it also contributed ~3.3 GB while the host was at 1 GB available and the worker container had already been OOM-killed three times.

## Root cause to investigate

`decide_intake` admits a trusted author's issue with no label at all (`config/retired_settings.py:146` — the admit-label gate is retired for trusted authors). Nothing in that path distinguishes an umbrella/epic row from an implementable one, so an epic filed by the owner is indistinguishable from a bug report.

## Suggested direction

Give intake a way to recognise a non-implementable row and skip it. Options, in rough order of robustness:

1. **A structural signal** — an epic has no acceptance criteria and links N child issues. Detect the shape rather than trusting a marker.
2. **A title/marker convention** — `Epic:` prefix is already used consistently on #4041–#4052. Cheapest, but a convention rather than a guarantee.
3. **An explicit exclusion list or label** the sweep maintains.

Prefer (1) or (3) over (2) alone; a prefix convention silently stops working the first time someone titles an epic differently.

Related: the sweeping-tickets skill was changed (#4065, merged) to forbid MINTING new umbrella rows. This is the complementary gap — nothing stops the factory from trying to *implement* the ones that already exist.

## Acceptance

- `issue_implementer` does not claim an umbrella/epic row.
- A regression test seeds an epic-shaped issue plus an implementable one and asserts intake picks the implementable one.
- The test goes RED on current main, where the epic is claimed.
- Existing epics #4041–#4052 are left in place (owner instruction) — this changes only what intake does with them.

souliane:
**Correction to this issue's framing, from evidence.**

[...]

PR #4295, opened 2026-08-06
fix(intake): honour overlay exclude_labels on GitHub issue intake (#4134)

souliane:
fix(intake): correct stale rule-number cross-refs after the exclude tier insert (#4134)

Inserting the exclude-labels tier at position 2 renumbered the old rules 2-5
into 3-6; only the canonical BLUEPRINT table and one in-file reference were
updated. Fixes the eight remaining stale references, including one on the
intake trust boundary that had inverted meaning: the admit-label comment
still cited the old rule number, which under the new table names the
trusted-author path instead. Also corrects label_admission.py's module
header, which named the assigned_issues scanner folded away by #3634 rather
than the two intakes that actually consume the predicate.

## Why


souliane:
Branch-update refused: `gh pr update-branch` returns *Cannot update PR branch due to conflicts*, so this cannot be advanced mechanically.

Conflict scope measured against `origin/main` (merge-base `a2c9f7db`) with `git merge-tree --write-tree`:

- `BLUEPRINT.md`
- `src/teatree/config/settings_loop_flags.py`
- `src/teatree/core/intake/factory_admission.py`
- `src/teatree/loop/scanners/issue_intake.py`
- `tests/teatree_core/intake/test_factory_admission.py`

These are semantic, not textual: main has since reworked the same intake decision path (the admitted-but-unclaimable queue and the merge-stall claim brake both landed in `issue_intake.py` / `scanner_factories.py`). Resolving it means deciding how `exclude_labels` composes with the current decision table, not replaying the old hunks — so this needs a plan before a rebase, and the verdict will not survive the rebase either way.

Context for prioritisation: this ticket currently holds one of the three intake in-flight slots, and intake reads `at_budget` (3/3) with ~30 admitted issues waiting, so clearing this PR directly unblocks new intake.

souliane:
Reconciliation scope for the `DIRTY` state, established by execution so it does not have to be re-derived.

`git merge-tree --write-tree` against `main` exits 1 and names five conflicting files. Base `a2c9f7db`, `main` `d2f35a6f`, PR head `ed2751ed`:

`BLUEPRINT.md`, `src/teatree/config/settings_loop_flags.py`, `src/teatree/core/intake/factory_admission.py`, `src/teatree/loop/scanners/issue_intake.py`, `tests/teatree_core/intake/test_factory_admission.py`.

The conflicts are **two different kinds of thing**, and treating them alike is how this goes wrong.

## Class A — naming churn, resolve toward `main` (most of the hunks)

`main` deliberately replaced the numeric rule citations with named ones: `rule 5` → "the admit-label rule", `rule 3's "work exists"` → "the work-exists rule", `rule 6` → the named fail-closed last rule. This branch still carries the numbers. Every one of these hunks resolves to `main`'s naming.

**This discharges the BLOCKING finding on both holds.** Verdicts 935 and 936 both blocked on `docs/blueprint/loop-topology.md:100` citing "rule 2 of the decision table" while this PR renumbered 2-5 → 3-6. On `main` that line now reads "the work-exists rule of the decision table", and the file carries no numbered-rule citation at all. The defect was a renumber desynchronising a numeric cross-reference; `main` removed the numbering, so there is nothing left to desynchronise. Worth confirming during the fix rather than taking on trust here.

## Class B — one genuine semantic collision, in `factory_admission.py`

Both sides added a **new withholding rule at the same slot**, and they are not alternatives:

| | `main` | this branch |
|---|---|---|
| rule | umbrella/epic row (#4105) | overlay `exclude_labels` (#4134) |
| verdict | `IGNORE_UMBRELLA` | `IGNORE_EXCLUDED_LABEL` |
| parameter | `umbrella_labels` | `exclude_labels` |
| import | `umbrella.normalize_labels, umbrella_reason` | `label_admission.excluded` |

Both rules must survive. A resolution that takes one side wholesale is textually clean and silently wrong in both directions: taking `main` deletes the entire feature this PR exists to deliver, and taking the branch deletes #4105's epic guard and reopens a bug that already merged. The reconciled function needs **both** parameters, **both** verdicts, **both** imports and **both** checks — and the docstring's decision table has to describe three withholding rules rather than two.

This is exactly the collision verdict 935's last finding predicted, when it flagged merge-order coordination with #4289 while that PR was still open at `9d1b78b7`. #4289 merged at 2026-08-07T01:14:27Z, so the predicted collision is now the actual one.

## Open question for whoever takes it

Rule **order** is a real decision, not a merge artefact, and the two sides argue opposite groupings in prose that also conflicts. `main` places umbrella above "work exists" because umbrella-ness is a property of the row; this branch places `exclude_labels` next to `needs-triage` because both are labels a human applied to withhold an issue. Both arguments hold once both rules exist, and they do not settle the ordering between each other. Decide it deliberately and record the reason in the docstring — do not let the merge pick.

## Still outstanding from the holds

Neither hold's remaining findings are addressed by the reconciliation, and they should be carried into the fix rather than lost with the SHA: `stranger_pr.pr_is_admitted` calls `decide_intake` without `exclude_labels`; `excluded()` matches exactly and case-sensitively, pinned by no test; a TOML overlay declaring `exclude_labels` as a string degrades to a character set; and an excluded row leaves only a log line with no durable record.

A verdict is bound to its SHA, so the reconciled head needs a fresh independent cold review whatever the merge produces — the two existing holds will not carry forward.

Issue #4120, opened 2026-08-03
workspace ticket run in the worker container bakes a container-only gitdir pointer, so the worktree is unusable from the host


souliane:
`t3 <overlay> workspace ticket` run inside the worker container writes a `gitdir:` pointer that only resolves inside the container, so the worktree it hands back is unusable from the host.

## What happens

```
docker exec teatree-teatree-worker-1 t3 teatree workspace ticket <issue-url>
  souliane/teatree: worktree #<n>
  Ticket #<id> — worktrees in /home/teatree/workspace/t3-workspaces/t3-teatree/<branch>
```

The printed path is bind-mounted, so the host sees the files. But every git command against it fails from the host:

```
$ git -C /home/teatree/workspace/t3-workspaces/t3-teatree/<branch>/teatree log --oneline -1
fatal: not a git repository: /home/teatree/teatree/.git/worktrees/teatree<n>
```

`<worktree>/.git` contains `gitdir: /home/teatree/teatree/.git/worktrees/teatree<n>` — the container's path for the main clone. On the host that clone lives at `/home/teatree/teatree-deploy`, and `/home/teatree/teatree` does not exist.

Existing worktrees created host-side carry the host spelling (`gitdir: /home/teatree/teatree-deploy/.git/worktrees/...`) and work fine, so the two producers disagree.

## Why

The clone is mounted from different paths on the two sides:

| | main clone | worktrees dir |
|---|---|---|
| host | `/home/teatree/teatree-deploy` | `/home/teatree/workspace/t3-workspaces` |
| container | `/home/teatree/teatree` (volume `teatree_teatree_src`) | `/home/teatree/workspace/t3-workspaces` (same path) |

The worktrees dir is bind-mounted at an IDENTICAL path on both sides, so the worktree's own location is portable. The main clone is not: git records an absolute `gitdir:` pointer to it, and `git worktree add` run in the container bakes in the container spelling. The worktree admin dir also lands under the container's clone, invisible to the host.

## Impact

An agent driving `t3` through `docker exec` (the documented way to reach the control DB from this box) gets back a worktree it cannot run `git` against from the host, where its Edit/Write tools operate. The workaround is to route every git command back through `docker exec -w <worktree>`, which works but is easy to get wrong and silently splits the toolchain: file edits on the host, git state in the container.

## Options

[...]

Issue #4287, opened 2026-08-06
An unconditional git worktree prune treats 'not mounted in this venue' as 'deleted', and the data-loss guard returns 'no work' for a path it cannot read


souliane:
Found in the cold review of PR #4286 (#4120), recorded as a durable issue because the hazard outlives that PR: the same unconditional prune sits in the hourly housekeeping path, not only in interactive provisioning.

## The defect

`src/teatree/core/runners/provision.py:180-181` prunes unconditionally:

```python
if any(not entry.on_disk for entry in _registered_worktrees(clone_str)):
    git.run(repo=clone_str, args=["worktree", "prune"])
```

`on_disk` is `Path(self.path).is_dir()` (`provision.py:72-74`) — a filesystem question answered **in the reading venue**, not a statement about the worktree. A registration that is merely *not mounted here* is indistinguishable from one that was *deleted*.

The sibling docstring at `provision.py:196` already names this exact trap — *"'git does not know about it' is a statement about the vantage point, not about the disk"* — but the prune does not honour it.

## Why it is harmless today and dangerous the moment the clone is shared

Measured from inside `teatree-admin` against today's clone: **91 registrations, 0 missing** — every worktree it knows lives under a path-identical bind, so the prune can never fire on a false negative.

Point that same code at a clone whose worktrees are NOT all visible in the container and the picture inverts. Against the host deploy checkout: **192 registrations, 135 not visible in-container** (`~/adhoc-wt/*`, `~/wt*`, and other host-only paths are not mounted). `git worktree prune` with no `--expire` deletes those admin dirs immediately — verified on git 2.53.0 in a scratch clone.

Result: 135 host checkouts get their `.git/worktrees/<n>` deleted and answer `fatal: not a git repository` — the very symptom #4120 exists to fix, inflicted on the host at 135x the scale.

## The second half: the data-loss guard fails open cross-venue

`src/teatree/core/management/commands/_workspace/cleanup.py:324-331` — `prune_branches` runs `git worktree prune` and *then* computes `wt_branches`. Branches whose only protection was "a live worktree has it checked out" lose that protection inside the same function that decides what to delete. In the `--merged` / squash passes `wt_path` is then empty, so `_worktree_clean(wt_path)` — the uncommitted-changes check — is **skipped entirely** and only `git.branch_delete` runs.

And `provision.py:100-102`:

```python
def _holds_unsalvageable_work(wt_path: str) -> bool:
    if not Path(wt_path).is_dir():
        return False
```

[...]

PR #4297, opened 2026-08-07
fix(deploy): cut worktrees from a clone every venue can name (#4120)

souliane:
fix(deploy): cut worktrees from a clone every venue can name (#4120)

`git worktree add` bakes an ABSOLUTE `gitdir:` pointer into its SOURCE CLONE, so
the clone's path — not the worktree's — decides who can use the result. The image
links the org-prefixed discovery path at the `teatree_src` volume, a path that
exists in no other venue. A worktree the containerised `workspace ticket` handed
back therefore answered `fatal: not a git repository` from the host, while its
files stayed readable through the `t3-workspaces` bind — so the failure named the
admin dir, not the cause, and an agent whose file tools run host-side had to route
every git command back through `docker exec -w`.

- compose: the deploy checkout joins the SHARED volume anchor as a read-write
  bind at PATH IDENTITY (source == target == the one `TEATREE_DEPLOY_CHECKOUT`
  placeholder the watchdog's read-only mount already reads); `teatree-init` and
  `teatree-worker` carry the value in their environment.
- entrypoint: `retarget_clone_discovery` points the discovery link at that
  checkout when it is a git checkout, degrades to the image link when it is not,
  and never replaces a real directory at the link path.
- Dockerfile: the baked link stays, now documented as the fallback.
- docs: deploy/README.md's variable table and `clone_provision`'s module
  docstring no longer describe the container's clones as unshareable — a bind at
  path identity is exactly what makes one shareable.

Read-write because `worktree add` writes `.git/worktrees/<n>` into the clone; the
watchdog keeps its own read-only spelling, pinned by a test. The container's UID
is the host deploy user's, so those writes land with the ownership the host
expects.

Why not the issue's literal option 1 (swap the `teatree_src` volume for a host
bind): 49 live worktrees record gitdirs into that volume and `doctor` reports 21
checkouts holding unshipped work — replacing it orphans them. Option 2 (rewrite
the recorded pointer) cannot work at all: the admin dir lives inside the
container-only volume, unreadable from the host at any spelling. Option 3 (refuse)
would stop the only sanctioned `t3` venue on this box. This shape is additive:
`teatree_src` stays mounted, so the existing worktrees keep resolving.

Four existing deploy invariants enumerate the anchor's binds as one homogeneous
set, so the new mount — legitimately a different KIND — turned them red. Each is
repaired by splitting the set, never by loosening the assertion:

- the state-dir test still asserts EXACT equality, now over the state planes plus
  the checkout, so a fifth mount still cannot slip in unnoticed;
- the rebased-on-host-home rule iterates the state planes only, and the checkout
  gets its own STRONGER rule — source == target == the one placeholder, writable;
- the fresh-box pre-create rule excludes the checkout, because dockerd's
  auto-create-as-root hazard cannot reach a directory that exists by construction:
  it is the checkout each entry point is EXECUTING FROM. Replacing that coverage
  is a positive pin that both scripts derive `TEATREE_DEPLOY_CHECKOUT` from their
  own `$REPO_ROOT`, so the exclusion is an argument rather than a hole.

Tests: `tests/test_deploy_clone_venue_portability.py` — compose invariants (path
identity, writable, watchdog still read-only, env forwarded), the real shell
function run in bash across six cases, and a git control proving the recorded
`gitdir:` names the symlink TARGET, which is why the target is what must be
mounted. Observed RED first (11 failed, 2 passed — the 2 pin unchanged behaviour).
Anti-vacuity control on the repaired invariants: retargeting the checkout mount
off path identity turns the path-identity test and the exact-set test RED
(2 failed, 14 passed); restored, 44 pass across the three deploy files.

Open questions & assumptions
- assumed: sharing the deploy checkout as the worktree source clone is
  acceptable. `deploy.sh` fetches/pulls there and the worker will now also
  `worktree add` and fetch into it; git's own locking covers the overlap, and the
  host already cuts worktrees from this clone (28 of them today).
- assumed: read-write on the shared mount is wanted. The alternative — a second,
  dedicated clone under a new host bind — costs disk and a first-boot clone but
  leaves the deploy checkout untouched. Say the word and it is a small change.
- assumed: excluding the checkout from the pre-create rule is correct rather than
  adding an `install -d` for it. The source is the script's own `$REPO_ROOT`, so
  creating it is a no-op, and the watchdog already binds this same source today —
  a `TEATREE_DEPLOY_CHECKOUT` naming a nonexistent path already fails the deploy,
  so the anchor mount adds no new fresh-box failure mode.
- assumed: the state-plane rule should not be widened to "source == target OR
  rebased on host home". A disjunction would let a future state mount silently
  drop off the host-home knob; two rules, each exact, keep both properties.
- open: this closes container -> host only. A worktree cut on the HOST from the
  same checkout is now also resolvable inside the container, but nothing yet
  repoints the host's own discovery link, so host-cut worktrees keep whatever
  spelling they have.
- open: the issue's acceptance test (`git -C <path> rev-parse --show-toplevel`
  from the HOST) cannot be observed from inside the worker container where this
  was built. Proof here is the deterministic deploy-invariant lane plus a
  rendered `docker compose config` showing init/worker at path identity and the
  watchdog still read-only; the end-to-end assertion needs a redeploy.
- open: 15 other tests are red in this venue for environmental reasons (a live
  `GH_TOKEN` in the worker container defeats the no-token cases, the live control
  DB defeats the host-projection case, and `cli_doctor` runs the real `t3 doctor
  check` against a box carrying WARNs). None is reachable from this branch — its
  only `src/` change is module-docstring prose — so they are read as pre-existing,
  not verified against a clean checkout of main.
- decided-by-user: option 1 ("mount the main clone at the same path on both
  sides") was marked Preferred in the issue; this is that, in the shape the
  existing wiring allows.

## Why


souliane:
**This PR has an empty diff against `main`. It duplicates #4286, which already merged. Recommend closing it rather than merging.**

Measured, and cross-checked rather than taken from a single empty read:

```
gh api repos/souliane/teatree/compare/main...16f76879
  ahead_by=2  behind_by=0  status=ahead
  .files[]  ->  (empty)
```

Two commits ahead, **zero changed files** — the commits reproduce content already present on `main`, so they are no-ops against it. Same title, same branch (`4120-workspace-ticket-run-in-the-worker-conta`), and a byte-identical body to **#4286**, which merged at `2026-08-06T21:21:22Z`.

(The empty `.files` list was not trusted on its own — an empty read is a non-answer, so it was confirmed against the compare API's `ahead_by`/`status`.)

**Merging this would create a commit that asserts a change it does not make.** There is nothing to review and nothing to land.

## Why this matters more than a normal duplicate

#4286 — the PR this duplicates — merged over a recorded `hold` (verdict 930). Its deploy then destroyed **86 host worktree registrations** (191 → 105) at 00:23Z: the working trees survive on disk but answer `fatal: not a git repository`. That failure is #4287, and it will recur on each containerised provision until the unconditional `git worktree prune` is made venue-aware.

So the change this PR re-proposes is not merely already-landed — it is the change currently causing damage. **The work that is actually needed is #4287's fix, not another copy of #4120's original change.**

Recorded as `hold` verdict 938 at `16f76879`.

souliane:
Closing as superseded: this branch's three-dot diff against main is empty — the fix it carries landed as #4286 (merged 2026-08-06 21:21Z, a2c9f7db). Merging would add a no-op merge commit. The branch is left in place; nothing is deleted. Independent cold review recorded a hold on the same grounds.

Issue #4299, opened 2026-08-07
Residuals from #4285: an unguarded commit, a denial message that can be false, and two divergent accountings


souliane:
Three findings from an independent cold review of #4285 at `0e73cabe`. The PR merged at
`8578bac9` while that review's `hold` stood, so these are now on `main` and need their own home.

The headline fix in #4285 is **real and proven** — restoring the double-count (`max(claims, seats)`
-> `claims + seats`) turns two named tests RED, and removing the seat seam entirely turns 9 RED. What
follows are residuals, not a claim that the PR was wrong to exist.

---

## 1. Commit `71adc657` landed unguarded, and the test written for it is vacuous

**Mutation control:** delete that commit's entire contribution (the one-shot narrowing) and the suite
is **29/29 green**. Nothing detects its removal.

The behaviour genuinely differs, so this is not dead code being tested correctly — it is live code
with no guard. A discriminating probe: ceiling 5, one unseated claim, 10 dispatches ->
**4 admitted at head vs 5 without the narrowing**.

`tests/teatree_core/test_dispatch_admission.py` :: `test_only_the_first_of_several_claims_exempts_a_dispatch`
is the test that claims to pin it. It passes with **and** without the code, because at ceiling 4 with
4 claims the arithmetic refuses the second dispatch either way. It asserts an outcome that is
overdetermined by the fixture.

**Wanted:** a test whose failure is caused by removing the narrowing — i.e. one built on a fixture
where the narrowing is the only thing that changes the answer (the 5/1/10 shape above discriminates).
A regression test never observed RED guards nothing.

## 2. The dispatch denial message can be factually false, and misdirects

`src/teatree/core/dispatch_admission.py:209`. Reproduced: 3 interactive claims, ceiling 4 -> the
second dispatch is denied with

```
live agents 3 at/over governor ceiling 4
```

3 is not at or over 4. Before the change the string was guarded by `if live >= decision.ceiling`;
it is now emitted unconditionally after a seat refusal. `live` is computed from the **deduped**
accounting while the refusal came from the **un-deduped** one, so the number in the message and the
number that caused the refusal are from different arithmetic.

`_block_message` then advises waiting for load hysteresis — not the actual cause, so an operator
reading it goes looking in the wrong place. This is the defect class #4217 already cost this repo:
a definite statement assembled from a signal that was not the one that decided.

**Wanted:** the message states the condition that actually refused, or says plainly that a seat was
refused. A test that asserts the emitted number is the number the decision used.

[...]

PR #4300, opened 2026-08-07
fix(core): denial message never understates its own ceiling (#4285 review)

souliane:
## Summary

Fixes items 1 and 2 from #4299 (residuals from #4285's cold review, HOLD at
`0e73cabe`, merged anyway at `8578bac9`), plus the docs-drift minor for
`hooks/CLAUDE.md`.

**Item 1 — commit `71adc657` landed unguarded, its regression test is
vacuous.** `test_only_the_first_of_several_claims_exempts_a_dispatch`
(ceiling 4, 4 claims) refuses the second dispatch whether the one-shot
exemption is present or reverted to the blanket per-session exemption it
replaced, so it never discriminates between them. Added #4299's suggested
discriminating probe: one claim, ceiling 5, 10 dispatches for the same
session — 4 admitted with the one-shot exemption, 5 with the blanket one.

**Item 2 — the denial message can be factually false.** The un-deduped
CHECK (`other_agents + ahead`) can refuse while the deduped REPORT
(`live_agent_count`) is still under the ceiling (item 3, left open below).
`dispatch_admission_denied_reason` reported the deduped number
unconditionally, producing messages like "live agents 1 at/over governor
ceiling 2" (1 is not at/over 2). Now reports `other_agents` plus a fresh
seat count taken right after the refused seat was deleted — reconstructing
the internal `ahead` value exactly, so the reported number is guaranteed
`>=` the ceiling by construction.

**Docs-drift minor.** `hooks/CLAUDE.md`'s #4129 paragraph still described
the pre-dedup `Task.objects.active_claims()`-only accounting.

**Left open in #4299** (not attempted here): item 3 (the CHECK/REPORT
accounting divergence itself — an accepted fail-safe trade-off that the
issue asks for an explicit decision on, not a bug fix), the unscoped
`release_seat` idempotency guard, the exemption race bounded to +1, and
the BLUEPRINT.md docs-drift.

## Test plan

- [x] `test_the_denial_message_never_understates_its_own_ceiling` observed
      RED before the message fix, GREEN after.
- [x] `test_a_burst_of_ten_dispatches_admits_exactly_the_first_seats_worth`
      passes against shipped code (confirms it discriminates from the
      pre-71adc657 blanket exemption it's designed to catch).
- [x] Full `dispatch_admission` + `interactive_dispatch` +
      `dispatch_seat_release` suite: 63 passed.
- [x] `t3 tool verify-gates`: commit-stage + push-stage, all green.


souliane:
Independent cold review at `e4a5a4c8` — **HOLD** (verdict 945). CI is genuinely green: all 7 required contexts `completed/success`, 39 jobs success, 3 skipped by design, 0 pending, 0 failing, `mergeStateStatus: CLEAN`. Blast class `logic` — admission behaviour is *provably* unchanged.

**The fix works, and is properly guarded.** Six mutation controls:

| Control | Outcome |
|---|---|
| Revert `occupied` to `live_agent_count()` | **RED** — `test_the_denial_message_never_understates_its_own_ceiling` fails; only that test (30 passed). Non-vacuous. |
| Revert the one-shot exemption to the pre-`71adc657` blanket | **RED** — burst test fails 5 vs 4. Confirms #4299 finding 2 is now genuinely guarded. |
| Admit/deny probe, pre vs post, 6 scenarios | **Identical** — message-only, no behaviour change. |
| Instrumented tie-back probe, 5 scenarios | `reported == other_agents + ahead` exactly, always ≥ ceiling. Re-tied to the deciding arithmetic. |
| The reported case (3 claims, ceiling 4) | Pre-fix `live agents 3 at/over governor ceiling 4`; post-fix `live agents 4 …`. Reproduced and fixed. |

Held on four small things that belong here rather than in a third follow-up on the same seam:

1. **The falsity moved rather than disappeared** (`src/teatree/core/dispatch_admission.py:221`). `occupied` is the un-deduped *charge*, not live agents. Pre-fix, 4/7 messages were self-contradictory but always reported the true population; post-fix, 5/7 report a number that is not the live agent count. Concrete regression — 4 claims in one session, ceiling 4: main emits `live agents 4` (true), this PR emits `live agents 5` while `live_agent_count()` is 4, i.e. exactly the loop-claim case the dedup exists for. A relabel fixes it: `charged population 5 (live agents 4) at/over governor ceiling 4`.
2. **The wrong-remedy half is untouched** (`hooks/scripts/dispatch_admission_gate.py:70-72`). `_block_message` still says the governor is braking and to wait for load hysteresis; for a seat/ceiling refusal the remedy is a seat release on `SubagentStop` or the 30-min `SEAT_WINDOW`. Operator still sent to the wrong place — the #4217 class this PR set out to close.
3. **`live_agent_count()` now has zero production callers** (`dispatch_admission.py:124`; whole-tree grep finds only its docstrings, `__all__`, tests). It was the sole production consumer of the per-session `max` dedup, so #4129's *reporting* correction is inert. That changes #4299 item 3 — the CHECK/REPORT divergence no longer surfaces anywhere — which the PR body doesn't mention.
4. **The new guard is weaker than its docstring** (`tests/teatree_core/test_dispatch_admission.py:343`). `assert reported >= 2` still passes with `occupied` hard-coded to `decision.ceiling`. Add an equality assertion against the refusing `other_agents + ahead`.

Plus: `BLUEPRINT.md:272` is still stale on this seam (pre-dedup wording) while the sibling `hooks/CLAUDE.md` line was fixed here.

**#4299 status after this PR:** finding 1 partially fixed (false relation fixed and guarded; wrong-remedy open, mislabelled population new); finding 2 fixed and guarded; finding 3 open, with its character changed as above.

Issue #4278, opened 2026-08-06
A recorded hold verdict was silently superseded by a later merge_safe on the same SHA, and the PR merged with no MergeClear


souliane:
PR #4273 carried two `ReviewVerdict` rows for the **same commit**:

```
verdict 921  hold        sha=285a3f14  reviewer 'cold-reviewer-pr4273-independent'
verdict 922  merge_safe  sha=285a3f14  reviewer 'cold-reviewer:task-1487'
```

It merged at 12:20:17Z as `7be8e34a`. **No `MergeClear` row exists for it** — `MergeClear.objects.filter(pr_id=4273)` is empty.

So a recorded `hold` was superseded by a later `merge_safe` on the identical SHA, and the merge went
ahead. Nothing reconciled the two, nothing refused, and nothing announced that two reviewers had
disagreed.

## Why this is the dangerous shape

A `hold` is the only mechanism that stops reviewed-and-rejected work landing. If a second review of the
same SHA silently overrides it, then the effective policy is **"last verdict wins"**, and the safety
property is whichever reviewer happened to finish second. That is not a review gate; it is a race.

Note the asymmetry that makes it quiet: `merge_safe` after `hold` merges, while `hold` after `merge_safe`
presumably does not un-merge. So the failure mode is one-directional and always toward landing.

## The disagreement was substantive, and checkable

This was not two reviewers splitting hairs. The `hold` finding is mechanical and anyone can verify it in
the merged tree:

- `src/teatree/core/modelkit/task_failure_taxonomy.py` line ~51 (corrected by that very PR) states that
  the runtime-ceiling kind **does NOT collide by construction**.
- The `#:` comment on `_CAUSELESS`, ~70 lines below **in the same file**, still states the kinds
  **"collide on one fingerprint by construction"**.
- `src/teatree/core/repair_loop.py` repeats the refuted claim.

Measured, with a control proving the probe can detect a collision: two runtime-ceiling reasons differing
only in their second count fingerprint **differently** (`be181ad9f80e2e71` vs `397aa3e0805ba16a`), while
two no-envelope reasons collide. So the file contradicts itself on `main` right now, in a module whose
entire purpose is to state that distinction correctly.

The PR's own commit message asserted it had "verified every factual claim in the corrected prose". The
sweep missed the file it was editing — which is precisely what the second review did not catch.

## What is wanted

1. **A second verdict on a SHA that already has a `hold` must not silently supersede it.** Either refuse
[...]

PR #4306, opened 2026-08-07
feat(core): agent-scratch retention sweep for the RAM-backed /tmp (#4165)

### LINKED ISSUE #4165: /tmp is a tmpfs holding 8.8 GB of week-old agent scratch — 28% of the usable RAM pool is permanently consumed by garbage
Bottleneck analysis, 2026-08-04. Third of three siblings — #4163 (governor never reads RAM, the binding constraint) and #4164 (the reaper amplifier). This one raises the ceiling by returning memory that is currently wasted.

## The finding

`/tmp` on this box is a **16 GB tmpfs** — its contents are RAM, not disk. It currently holds **8.8 GB resident**, all of it stale agent scratch:

| File | Size | Last touched |
|---|---|---|
| `t3db.sqlite3` | 1.2 G | 07-27 |
| `costidx.sqlite3` | 1.2 G | 07-29 |
| `t3after` | 1.2 G | — |
| `ho-probe` | 1.2 G | — |
| `ho-probe-xdg` | 1.2 G | — |
| `board3841` | 1.2 G | — |
| `wt4081venv` | 933 M | — |
| `rev3970` | 770 M | — |

## Memory ledger for the box

| Item | GB |
|---|---|
| Total | 31.0 |
| Swap (100% consumed — no headroom left) | 1.4 |
| **`/tmp` tmpfs, resident** | **8.8** |
| Docker containers | 0.95 |
| Other daemons (gateway, signal-cli JVM, dockerd, claude daemon/pty) | ~1.5 |
| **Usable for factory work** | **~19.7** |

So **28% of what should be the working pool is week-old scratch**, on a box where the governor own 16-worker ceiling needs 19.8 GB at p90 worker RSS (#4163). Removing this garbage is the difference between the sanctioned ceiling being over the line and under it.

## The existing sweep does not cover it

The resource-pressure loop last plan cleans only `/tmp/claude-statusline` older than 2 days, and reported `reclaimed=0.00GB`. Nothing touches the multi-gigabyte database and venv scratch that agents leave behind.

## What to do

Two independent moves, both worth taking:

1. **Add a real retention sweep** to the resource-pressure plan: agent scratch under `/tmp` older than N days, with a size-ranked report of what it reclaimed. It must be **non-destructive by default** — report first, and never delete a path that is open by a live process or belongs to a running agent. This box has active worktrees and live agents; a naive `rm -rf` would be the wrong fix and is explicitly not what is being asked for.
2. **Question whether `/tmp` should be tmpfs at all on this host.** A 16 GB RAM-backed `/tmp` on a 31 GB box, used by agents that write gigabyte sqlite files, converts routine scratch into permanent memory loss. Moving it to disk (or shrinking it substantially) returns the space structurally rather than needing a recurring sweep.

Note that a manual `rm -rf` is a **blockage clear, not a ceiling raise** — measured refill is roughly a week. That is why the retention sweep or the tmpfs decision is the actual deliverable.

## Expected effect

Returning 8.8 GB to a 19.7 GB usable pool raises the safe concurrent-worker count from **13 to ~20 at p90 worker RSS** — a ~50% concurrency increase with no change to agents, model, or the test suite.

## Acceptance

- A retention policy exists, is configurable, and reports what it reclaimed (`reclaimed=0.00GB` while 8.8 GB sits stale is itself the bug — a sweep that reports success having done nothing).
- The sweep never removes a path in use by a live process or a tracked worktree.
- A decision is recorded on the tmpfs question, either way, with its reasoning.

Related: #4163, #4164, #4118 (the cgroup RAM floor — the container sees the HOST meminfo, so any memory-aware logic must honour the cgroup cap).

souliane:
Closes #4165

## What

An agent-scratch retention sweep for the temp root, plus the doctor probe that names the tmpfs **sizing** defect.

On this box `/tmp` is a 15.1 GB tmpfs on 31 GB of RAM. Scratch written there is memory, not idle disk — 8.8 GB of week-old sqlite/venv scratch was 28% of the working pool, permanently. The existing resource-pressure ladder reached only `/tmp/claude-statusline`, which is why it reported `reclaimed=0.00GB` with that 8.8 GB in front of it.

## The sweep

`teatree.core.retention.scratch.ScratchSweep` plans and applies retention over one temp root. A top-level entry is reclaimed only when **all** hold:

- older than `scratch_retention_days` (default 3)
- owned by this uid
- held open (fd or cwd) by no live process the probe can see
- not a registered worktree checkout, nor a parent or child of one
- not protected by name (`.X11-unix`, the deploy flock, `claude-statusline`, `systemd-private-*`, …)

Every guard that **cannot be answered keeps the entry** with its reason recorded. An unreadable process table or a failed worktree read makes the whole plan removable-empty rather than guessing — a read failure is never laundered into a deletion.

Operator surface, dry-run unless `--apply`, size-ranked with a per-entry verdict:

```text
t3 <overlay> retention scratch [--root PATH] [--days N] [--apply] [--json]
```

The same sweep runs automatically in **both** resource-pressure ladders — the disk one and the RAM one, because on a tmpfs-backed `/tmp` that scratch *is* RAM. `_clean_stale_statusline` is untouched: it owns its own path on its own 2-day rule.

## The venue seam

The scratch is host-side. A container has its own `/tmp`, so a container-scoped policy reclaims nothing of what fills the box. The app services therefore mount the host temp root **read-write** at `/host-tmp` and the host process table **read-only** at `/host-proc`.

The pair is the point. The open-file guard is answered from a process table, so reading *this* container's while sweeping the *host's* files would answer confidently about the wrong namespace. `root` and `probe_root` are resolved together (`resolve_scratch_sweep`) and never mixed; when only one half is mounted the sweep falls back to this venue's own `/tmp` + `/proc`.

Both are distinct targets, so the container's own `/tmp` and `/proc` are untouched. The mount is strictly narrower than what the worker already holds — the docker socket is root-equivalent on the host. What bounds the sweep is its own guards, not the mount.

## The tmpfs decision

`/tmp` stays a tmpfs. Moving it to disk is dropped and the reasoning is recorded so it is not re-proposed: the root filesystem was 84% full, so a 15 GB disk-backed `/tmp` trades RAM pressure for a filesystem-full failure that breaks the control DB; tmpfs is wiped on reboot, which is what bounded the observed leak to a week; and the live remount strands open handles.

The defect is the **sizing**. `t3 doctor check` gains `_check_tmp_tmpfs_sizing`: a WARN when the tmpfs may claim more than `TEATREE_TMPFS_MAX_RAM_PERCENT` (default 25) of machine RAM, naming the exact `systemctl edit tmp.mount` cap. It is the sibling of the existing fill-percent check, which measures how *full* the tmpfs is and cannot express how big it may *get*.

Applying that cap is a host action teatree cannot perform from a container, so the deliverable here is detect + report + the exact remediation.

## Settings

| Setting | Default | Meaning |
|---|---|---|
| `scratch_retention_days` | `3` | Age window; `0` disables the lane |
| `scratch_sweep_root` | `""` | Blank auto-resolves the host view, else this venue's `/tmp` |

Both per-overlay overridable.

## Architecture pre-check — #4165

1. **BLUEPRINT § alignment** — §5.6 scanner table (`ResourcePressureScanner`), updated in this PR; `docs/blueprint/configuration.md` for both settings.
2. **FSM phase boundaries** — n/a, no `Ticket`/`Worktree` transition.
3. **Extension-point contracts** — none touched. No `OverlayBase` hook, no scanner registration change beyond two payload fields on the existing `ResourcePressureScanner`.
4. **Component boundaries** — the sweep is `core/retention/` (the existing retention concern, filesystem lane beside the control-DB lane); the CLI subcommand joins the existing `retention` group; the ladder wiring stays in `loop/mechanical_resources`; the probe stays in `cli/doctor/checks_resources`.
5. **Dependency direction** — `teatree.loop` → `teatree.core` is an existing edge; `cli.doctor` → `utils.ram_probe` likewise. No new module in the graph.
6. **Test surface** — `tests/teatree_core/retention/test_scratch.py` (every guard, both apply paths, every degraded read), `tests/teatree_core/management/test_retention_command.py` (dry-run / apply / configured default / human view), `tests/teatree_loop/test_resource_pressure_mechanical.py` (both ladders, the zero-window switch, the swallowed failure), `tests/teatree_cli/doctor/test_tmp_tmpfs_headroom_check.py` (sizing bands + degraded reads), `tests/test_deploy_bindmount_compose.py` (the mount pair, rw/ro, no shadowing).
7. **Resilience invariants** — the external write here is a filesystem delete. Verify-by-re-read: the mtime is re-read immediately before each unlink. Idempotency: a second pass finds nothing to remove. Fallback: an unreadable probe degrades to plan-only, never to a blind delete. Heartbeat: the ladder records the plan on `ResourcePressureMarker` before and after. Sub-agent contract: n/a.
8. **Identity and key normalization** — one identity with two spellings: a temp entry as this venue names it (`root/<name>`) and as the process table names it (`probe_root/<name>`). `_probe_path` is the single translation, applied at the one boundary that compares them; nothing strips a prefix to force a match.
9. **Behavior preservation** — purely additive. `_clean_stale_statusline` is unchanged and its path is on the protected list so the two can never contend; no gate was narrowed and no must-block test inverted.
10. **Removability / harness-vs-data** — removable: deleting `core/retention/scratch.py`, the subcommand, the two ladder lines and the two mounts restores the prior behaviour with no other call site touched. Harness-vs-data: the mechanism is harness, the varying policy (window, root) is data in `ConfigSetting`, per-overlay overridable.

- **decided-by-user** — `/tmp` stays a tmpfs and is capped; moving it to disk is dropped, with the reasoning recorded in the check message and `deploy/README.md`.
- **decided-by-user** — the scratch is host-side, so a container-scoped policy would reclaim nothing; hence the host-namespace mounts.
- **assumed** — capping the host `tmp.mount` is a host action teatree cannot perform from a container, so the deliverable is detect + report + the exact remediation rather than an automated remount.
- **assumed** — the `/host-tmp` rw + `/host-proc` ro pair is acceptable on the app services because the worker already holds the root-equivalent docker socket; the sweep's guards, not the mount, are what bound it.
- **assumed** — retention default 3 days, the window the manual reclaim used.
- **assumed** — a pid whose `fd` dir is unreadable belongs to another uid and is skipped rather than failing the whole probe; only an unlistable proc root blinds it.
- **open** — whether the host `tmp.mount` cap has been applied on the box. The doctor WARN is the standing reminder until it is.

## Cold review — two findings fixed

An independent cold review at `c573ef2e5` found two HOLD-worthy gaps, both reproduced directly and now fixed:

1. **`resolve_scratch_sweep` hard-coded `probe_root=/tmp`** while the compose mount source is the configurable `${TEATREE_HOST_TMP:-/tmp}`. Under an operator override the open-file guard compared candidate paths against the wrong namespace, so `apply()` could delete a file a live host process held open — violating the ticket's own acceptance criterion. Fixed by forwarding `TEATREE_HOST_TMP` into every app service's environment (the same default expression the mount source already uses) and deriving `probe_root` from it, so the guard's namespace can never drift from the mount's. Regression-tested end to end: a file held open under a custom `TEATREE_HOST_TMP` override now correctly survives `apply()`.
2. **`_check_tmp_tmpfs_sizing` was hard-coded to inspect `/tmp`**, but `t3 doctor check` runs inside the container, where `/tmp` is the image's own overlay layer, not the host's tmpfs — the check was silently inert on the box it exists to protect. Fixed to prefer `/host-tmp` (a bind of the host's real temp root, which reports the source filesystem's fstype and size at the new mountpoint) whenever that mount is present, falling back to `/tmp` only when it is not.

Both fixes are guarded by tests confirmed to go RED on revert (mutation-checked by hand). `t3 tool verify-gates` exits 0 at the fix commit.

## Cold review at ec034deb — three CRITICALs fixed

A second independent cold review (ReviewVerdict 955) found three CRITICAL
defects the prior fix commit did not close. All three reproduced and fixed:

1. **The age guard read only the top-level entry's own mtime**, which is not
   a liveness signal for a tree — a directory's own mtime moves only when an
   entry is added/removed/renamed DIRECTLY inside it, so writing new content
   into an existing nested file never touches it. The reviewer reproduced
   this by running `apply()` for real: it deleted a working tree whose
   newest content had been written zero seconds earlier. Fixed by computing
   staleness from the newest mtime anywhere in the tree, in one recursive
   walk that also answers size and ad-hoc-repo detection (`_tree_stats`).
   `apply()`'s own pre-delete recheck is upgraded the same way, closing the
   identical race one step later (a nested write landing between planning
   and the unlink).
2. **The open-file guard never read `map_files` or `net/unix`**, so it could
   not see a memory-mapped file (mmap survives an fd close — the module's
   own docstring's worked example) or a bound `AF_UNIX` socket (a socket fd
   reads back as `socket:[inode]`, never the bind path — only visible via
   `/proc/net/unix`, a process-table-wide source no per-pid fd walk
   reaches). Fixed by reading both; an unreadable `net/unix` now blinds the
   whole probe, matching the existing `proc_root`-unreadable behaviour.
3. **`_worktree_paths()` only saw DB-registered rows**, so an ad-hoc git
   clone under the swept root was invisible to the guard entirely — five
   were in the review's measured deletion set, three of them 3.1 days old.
   Fixed by detecting a `.git` marker anywhere in a candidate's tree during
   the same walk, independent of DB registration.

Five mutation controls (two for tree-wide staleness — plan-time and
`apply()`'s own separate re-walk, isolated via a frozen-plan test; one each
for mmap, bound-socket, and ad-hoc-repo detection) all confirmed RED on
revert. `teatree.core.retention.scratch` is at 100% statement + branch
coverage. `t3 tool verify-gates` exits 0 at the fix commit.

Per the reviewer's explicit instruction this PR **stays in draft**. Undraft
only once these three findings are independently re-verified at the new
head — a verdict binds to a SHA, and this head has moved again.

## Cold review at 69bc97a27 — two coupled CRITICALs + a residual fail-open closed

A third independent cold review (ReviewVerdict 958) confirmed all three prior
CRITICALs genuinely fixed (reproduced against the real `apply()`, not just
read from the diff — 9 mutations RED, 5 entries rescued on the real temp
root). It then surfaced a new coupled pair plus a residual gap in the same
code path:

1. **`plan()` crashed outright** on a top-level entry this uid cannot search
   into: `(path / ".git").exists()` re-raises any unignorable `OSError`
   (`EACCES` is not one pathlib's own `_ignore_error` forgives), and
   `_inert_plan` walks the same tree, so even the retention-disabled path
   crashed. Reproduced on a mode-0700 directory owned by another uid. Fixed
   by wrapping the `.git` probe in `try`/`except OSError`.
2. **The open-file guard failed OPEN when structurally blind**, not just when
   a single source errored: inside the real `teatree-worker` container (uid
   1001, no added caps, bridge network, own PID namespace, reading the
   host's `/proc` through the read-only bind) every per-pid source — `fd`,
   `cwd`, `map_files` — came back unreadable for every pid (measured 0/17
   fd readlinks), and `/proc/net/unix` reflects the container's own network
   namespace rather than the host's. The old code returned an
   empty-but-not-`None` frozenset in that case, so `plan()`/`apply()`
   proceeded as though nothing were held — the exact fail-open the module's
   own docstring forbids. Fixed by tracking whether ANY pid answers through
   ANY per-pid source; when pids exist and not one of them does, the guard
   now returns `None` (probe unreadable) instead of "nothing held". A single
   foreign-uid pid answering nothing is still the normal multi-user case and
   does not blind the probe on its own — confirmed by a dedicated control.
3. **Residual: `os.walk`'s default `onerror=None` silently under-reports.**
   An unscannable subdirectory (an `EACCES` this uid can't descend into) was
   simply skipped rather than treated as "cannot prove staleness" — the
   original tree-wide-mtime finding, one level deeper. Reproduced: content
   written 0 seconds ago inside an unscannable subdirectory yielded
   `age=10.00d, verdict REMOVE`. Fixed with an `onerror` callback that forces
   the whole entry's `newest_mtime` to `None`, so it is kept rather than
   silently mis-aged.

Findings 1 and 2 are coupled by design: the crash in (1) was the only thing
preventing the fail-open guard in (2) from running unprotected inside the
container, so both had to land together — the review's explicit instruction,
followed here.

Two of the nine new regression tests needed a second pass to avoid the same
vacuous-test trap this ticket already hit twice: bumping a containing dir's
own mtime by creating a new nested file let the OLD code "detect" freshness
by accident (fixed by rewriting an *existing* file's content instead), and a
real `chmod(0o000)` also blocks `shutil.rmtree` from recursing, so an
apply()-level "still exists" assertion passed on the old code too, for the
wrong reason (fixed by simulating the scandir failure through a
`scratch.os.walk` patch, leaving the real tree genuinely deletable so only
the guard's own logic decides the outcome). All 9 confirmed RED on revert (7
red; 2 are deliberate "still answers" controls that must stay green).
`teatree.core.retention.scratch` stays at 100% statement + branch coverage.
`t3 tool verify-gates` exits 0 at the fix commit; CI is green (all 12 shards,
all 4 shuffle seeds).

`deploy/README.md` now documents that the open-file guard's real, current
contribution inside `teatree-worker` is `probe_gap` (fails closed) rather
than a working liveness check, until the underlying LSM/capability
restriction is separately isolated and addressed — out of scope for this
review (measured across four container configurations without separating
AppArmor's `docker-default` ptrace deny from the dropped `CAP_SYS_PTRACE`).

Per the reviewer's explicit instruction this PR **stays in draft**. Undraft
only once these findings are independently re-verified at the new head — a
verdict binds to a SHA, and this head has moved again.


souliane:
Independent cold review at `af7c92ec` — **HOLD** (verdict 949, blast class substrate).

**Required CI is RED and it is this PR's own doing** — 13 jobs including required `test (3.13)`; `main` at `d2f35a6f` is green. Root cause is finding 4 below.

## Measured blast radius (plan-only, real temp root)

| | |
|---|---|
| top-level entries | 2594 |
| **in the deletion set** | **1987** |
| bytes reclaimed | **0.07 GiB** |
| saved by the liveness guard | **2** |

It removes ~2000 paths to reclaim ~70 MB, and only 2 of 2594 entries were protected by liveness. The 8.8 GB that motivates the PR is already gone — the mount cap appears applied, which incidentally answers the PR body's one open question.

## Critical

**1. The age guard reads the top-level entry mtime, which is not a liveness signal for a tree.** A directory's mtime changes only when entries are added or removed *directly in it*. Demonstrated by running the real `apply()`:

```
RECLAIM   10.0d  agent-scratch-live   stale scratch older than 3d
A: top-level mtime age = 10.0d, NEWEST content inside = 0s ago
A contains a git repo with an uncommitted file: True
APPLIED. survivors: []
```

**A working tree holding an uncommitted file written zero seconds ago was deleted.** On the real temp root now, 3 removable dirs have deep content 0.66–2.04d old while the guard reads 13–20d.

**2. The open-file guard misses the two commonest liveness forms.** Confirmed empirically here: unix-domain sockets render as `socket:[inode]`, never a path (149 live) — a dir holding a bound listening socket is `RECLAIM`; and mmap'd-but-fd-closed files are missed because `map_files` / `maps` are never read (29 live processes hold such files) — which is the module docstring's *own* worked example of a venv with a `dlopen`'d `.so`.

**3. Ad-hoc repository clones under the temp root are unprotected.** `_worktree_paths()` reads only `Worktree` rows, and teatree's worktrees live outside the temp root — so that guard protects nothing there. Measured: **5 `.git`-bearing dirs in the current deletion set, three only 3.1 days old.** `workspace clean-all` already carries the #706/#835 "keep any dir holding a `.git`" guard; this new and more destructive context reuses none of it.

## Major

4. **`t3 <overlay> retention scratch` does not resolve** — the `retention` group registers only `prune`. BLUEPRINT, the configuration doc, the deploy README and the doctor's own WARN remediation all cite it. This is the CI failure.
5. **No per-path record of what was removed** — `apply()` logs a GB total; the ladder appends only a summary to a single overwritten field. Removing 1987 entries leaves zero path names, so a pass is undiagnosable afterwards.
6. **A recursive delete now sits at L2 behind no flag, ON by default** (retention days = 3), while that module's docstring still calls L2 "both non-destructive" and the BLUEPRINT line this PR edited still asserts "Every destructive lever defaults OFF".

## Credit where due

**All 10 mutation controls go RED** against a 154-passing baseline — every guard removed, both fail-closed degradations inverted, the apply-time mtime re-read removed, root/proc pairing mixed, ladder apply flipped. Fail-closed behaviour was observed live: a missing table made every entry `KEEP`. The guards this PR *has* are genuinely pinned; the defects are guards that were never written.

Follow-ups that should **not** block once the above are fixed: reading `map_files` and `/proc/net/unix` for a fuller liveness proof, a persisted audit log of removed paths, and extending the sweep to the secondary temp root.

souliane:
Re-review at `6bf279a7` — **still HOLD** (verdict 950).

**The CI major is fixed and confirmed.** Required checks are green (23 pass, 0 fail): registering `retention scratch` as an overlay CLI proxy leaf resolved the 13 RED jobs, `test (3.13)` among them. That was finding 4 and it is closed.

**The three criticals are untouched.** The delta from the reviewed tree is registration, a docs regeneration, two added tests, and an import hoist — `src/teatree/core/retention/scratch.py` changed by 4 lines, moving the `Worktree` import to module scope and dropping its `noqa`. No guard logic changed, so the measured findings carry over unchanged rather than needing to be re-derived:

1. the age guard still reads the top-level entry mtime, which is not a liveness signal for a tree — the real `apply()` removed a tree whose newest content was written zero seconds earlier;
2. the open-file guard still never reads `map_files` / `maps` and still cannot see a bound unix socket, which are the two commonest liveness forms and the module docstring's own worked example;
3. `_worktree_paths()` still reads only registered rows, so ad-hoc clones under the temp root are unprotected — 5 `.git`-bearing directories were in the measured deletion set, three of them 3.1 days old.

Blast radius is unchanged at 1987 of 2594 entries for 0.07 GiB, with 2 entries saved by liveness.

The credit from the first pass also stands: all 10 mutation controls go RED, so the guards this PR *has* are genuinely pinned. The defects are guards that were never written, which is why green CI does not move the verdict.

Unblocking needs 1–3 addressed on their merits — a liveness signal derived from the tree rather than its top entry, a fuller open-file proof, and the existing "keep any directory holding a `.git`" guard reused in this more destructive context.

souliane:
Head moved again, `6bf279a7` → `c573ef2e`. Hold re-armed as verdict 951 at the live head.

The delta is a **single added line** in `tests/teatree_cli/test_capabilities.py` (commit "prove the retention scratch json:true capability"). It cannot bear on the three criticals, so they stand exactly as recorded:

1. the age guard reads the top-level entry mtime rather than tree liveness — the real `apply()` removed a tree whose newest content was written zero seconds earlier;
2. the open-file guard never reads `map_files` / `maps` and cannot see a bound unix socket;
3. `_worktree_paths()` reads only registered rows, leaving ad-hoc `.git`-bearing clones under the temp root unprotected — 5 were in the measured deletion set, three of them 3.1 days old.

Measured blast radius is unchanged: 1987 of 2594 entries, 0.07 GiB reclaimed, 2 entries saved by liveness.

Flagging the re-arming itself, because it is load-bearing rather than bookkeeping. A verdict binds to a SHA, so each push voids the standing hold; two of the last three pushes here were a registration fix, an import hoist and a test line, none touching the guards. If a fresh `merge_safe` is recorded at the live head while the older holds sit on superseded SHAs, nothing stops the merge — that is the failure mode tracked in #4278, and it has already fired twice on this repo. For a change that recursively deletes ~2000 paths and has been demonstrated to remove a working tree with seconds-old content, that is the outcome most worth avoiding.

CI being green does not bear on any of the three. The CI failure was the unregistered CLI leaf, and that is genuinely fixed.

souliane:
Converted to **draft** as a deliberate hold. Reversible with `gh pr ready` the moment the three criticals are addressed — this is a hold, not a rejection.

**Why now.** The PR just went `CLEAN` (mergeable) at `c573ef2e` while hold 951 stands at that same head, and two reviewing tasks are claimed against this ticket. A fresh `merge_safe` recorded at the live head supersedes a standing hold with no `MergeClear` in between — that is #4278, and it has already merged two PRs over their own holds on this repo. So the hold verdict alone is not load-bearing here.

**Why the substrate class is not sufficient either.** The substrate guard keys on the CLEAR's `blast_class`, which follows whichever verdict clears it, not on my classification. A reviewer that classes this `logic` gets no substrate refusal — the same way an additive migration was classed `logic` and auto-merged past an intended hold in #3464.

**Why this PR specifically.** The measured `apply()` removes 1987 of 2594 entries and was demonstrated deleting a working tree whose newest content was written zero seconds earlier, with 5 `.git`-bearing directories in the deletion set. An unwanted draft costs one command to undo. A merged destructive sweep does not.

The three criticals are unchanged and listed in the earlier comments: the age guard reading top-level entry mtime rather than tree liveness; the open-file guard never reading `map_files` / `maps` and blind to bound unix sockets; and `_worktree_paths()` protecting only registered rows, leaving ad-hoc clones under the temp root exposed.

Green CI does not bear on any of them — the CI failure was the unregistered CLI leaf, and that is genuinely fixed. All 10 mutation controls still go RED, so the guards this PR has are well pinned; the defects are guards that were never written.

Undraft as soon as those three are addressed, or if the owner judges the hold wrong.

souliane:
Head moved `c573ef2e` → `ec034deb`. Hold re-armed as verdict 955. Staying draft.

**The fix that landed is real, and it is the right shape.** `resolve_scratch_sweep` now reads `probe_root` from `TEATREE_HOST_TMP` — the same variable the compose mount source is built from — so an operator who moves the host mount source moves the guard's namespace with it. The reasoning in the new docstring is correct: a hard-coded `/tmp` there re-creates the bug the variable exists to prevent, the moment the two diverge. Credit where it is due; that closes a genuine fail-open.

**But the commit says "close two … HOLD findings", and the three criticals are not among them.** The entire `scratch.py` delta is the header comment plus `resolve_scratch_sweep`; none of the other changed files (`deploy/README.md`, `deploy/docker-compose.yml`, `checks_resources.py`, and the two test files) contains the guards in question. So all three stand, unchanged:

1. **The age guard still reads the top-level entry mtime**, which is not a liveness signal for a tree. The demonstration stands: the real `apply()` removed a working tree whose newest content had been written zero seconds earlier, because a directory's mtime only moves when entries are added or removed *directly in it*.
2. **The open-file guard still never reads `map_files` / `maps`, and still cannot see a bound unix socket.** Those are the two commonest liveness forms, and the mmap case is the module docstring's own worked example.
3. **`_worktree_paths()` still reads only registered rows**, so ad-hoc `.git`-bearing clones under the temp root are unprotected — 5 were in the measured deletion set, three of them 3.1 days old.

Blast radius is unchanged at 1987 of 2594 entries for 0.07 GiB, with 2 entries saved by liveness.

**Why it stays draft rather than relying on the verdict.** The head move just voided verdict 952, which was bound to `c573ef2e` — for a window, the only thing holding this PR was the draft flag. That is the specific case the draft was set for, and it is worth noting it did its job. I have re-armed at `ec034deb`, but the same window reopens on the next push, so the flag stays until the criticals are closed rather than being re-litigated each time.

Undraft with `gh pr ready` once 1–3 land — or sooner if the owner judges the hold wrong. Nothing here is a comment on the PR's quality overall: all 10 mutation controls still go RED, so the guards this PR *has* remain well pinned. The defects are guards that were never written.

souliane:
Independent cold review at `69bc97a2` — **HOLD** (verdict 958). CI is green: 43 check-runs read paginated, 40 success / 3 skipped, zero failures, all seven required contexts passing.

**The three criticals are genuinely closed on the host, and that deserves saying first.** Each was reproduced against the real `apply()`, not read:

- A directory whose top-level mtime read **10.00d** while its deep content was written **0.00d** ago is now `KEEP … younger than 3d`, and survives a real apply. Mutations: plan-time staleness 1 RED, the apply-time pre-delete re-walk 1 RED.
- An mmap'd-fd-closed file (via libc `mmap(2)`, since CPython's `mmap` dups the fd and would have masked the test) and a bound `AF_UNIX` socket each protect their directory, and each guard is independently decisive — neutering `map_files` 2 RED, neutering the `/proc/net/unix` probe 1 RED.
- Ad-hoc `.git`-bearing dirs are kept; mutation 4 RED. On the real temp root **5** are now saved, including the three ~3.1-day-old ones the prior review named.

Nine mutations RED, nothing unpinned. `probe_root` from `TEATREE_HOST_TMP` survives and is correct. The new guards are **not** inert: they rescue 9 entries the old code deleted.

Two new criticals block it, and they are coupled.

## C1 — `plan()` crashes, so the feature is a silent no-op on this box

`scratch.py:170`, `holds_git = (path / ".git").exists()`. On Python 3.13 `Path.exists()` re-raises non-ignorable `OSError`, and `EACCES` is not ignorable. Any top-level entry this uid cannot search therefore raises. Verified through the shipped CLI against the real temp root:

```
retention scratch --root <temp root> --days 3
→ PermissionError: [Errno 13] Permission denied: '<a mode-0700 dir owned by another uid>/.git'
```

`_inert_plan` calls the same walk, so even the retention-**disabled** path crashes. In the ladder `_sweep_scratch` swallows the exception, so nothing surfaces — the feature simply never runs.

## C2 — the open-file guard resolves ZERO paths in the container, and fails silently OPEN

The sweep exists to sweep the host temp root **from inside the container**, and there every liveness source is dead. Measured in the exact `teatree-worker` configuration (bridge network, own PID namespace, uid 1001, no added caps, not privileged), reading the host process table through the `/proc` bind mount, against a host process owned by the same uid:

| source | host | container |
|---|---|---|
| `fd/` readlinks resolved | 17/17 | **0/17** |
| `cwd` | resolves | **EACCES** |
| `map_files` | 35 entries | **EACCES** |
| `/proc/net/unix` bound paths | 126 | **0** |

`--pid=host` changes nothing. `--network=host` fixes only `net/unix`, and the `admin` service has it while the **worker** — where the automatic ladder sweep runs — does not.

All four fail **open**: `_held_paths()` returns an empty `frozenset` rather than `None`, and `probe_gap` stays `""`, so the plan proceeds as though nothing is held. That falsifies the module docstring's "never launders a source it could not read into 'nothing is held'", the BLUEPRINT row, and `deploy/README.md`. Concretely, on this box today, 126 live bound socket paths and the fd/cwd/mmap holders of 45 of 50 host pids are invisible to a sweep that deletes their directories.

## The interaction — why these must be fixed together

**The C1 crash is currently the only thing preventing C2 from causing damage.** Fix the crash alone and an inert feature becomes an actively dangerous one: it would then run, in the container, with no liveness protection whatever. Please do not land a `.git`-probe fix on its own.

## Also open

A residual fail-open in the tree-wide age guard (`scratch.py:172`): `os.walk` runs with the default `onerror=None`, so an unreadable subdirectory is silently skipped and both `newest_mtime` and size are under-reported. Reproduced — content written 0 seconds ago inside a mode-000 subdir yielded `age=10.00d, size=0 bytes, REMOVE`. That is the original Critical 1 shape one level deeper, against the module's own stated FAIL-CLOSED contract.

Both prior non-criticals remain: no per-path record of what was removed, and the recursive delete still sits at L2 with no `allow_destructive_*` gate, ON by default at `scratch_retention_days = 3`, while that module's docstring calls L2 "both non-destructive" and the BLUEPRINT row this PR edits still asserts "Every destructive lever defaults OFF".

## Re-measured blast radius

| | prior review | now |
|---|---|---|
| top-level entries | 2594 | 2428 |
| in deletion set | 1987 (76.6%) | 1816 (**74.8%**) |
| candidate reclaim | 0.07 GiB | 0.040 GiB |
| saved by liveness | 2 | 1 |
| saved by ad-hoc-repo guard | — | **5** |
| saved by tree-wide mtime | — | **4** |

The proportion barely moved, but that is not evidence the guards are inert — it is that most of the temp root genuinely is old, tiny, unheld scratch. The honest caveat is that on the container venue the liveness contribution is **0**, not 1.

## Not established

The precise LSM or capability behind the container-side `EACCES` was not isolated (measured across four container configurations without separating AppArmor's `docker-default` ptrace deny from the dropped `CAP_SYS_PTRACE`), so no tested remedy is named. The shipped `apply()` could not be run against the real temp root at all, because C1 blocks it — the blast-radius figures come from a harness with only the `.git` probe guarded, everything else shipped code.

The PR stays in draft while these stand.

souliane:
Independent cold review at `f8773dac` — **HOLD** (verdict 961). CI is green on every required job.

**C1 is closed, and so is the `os.walk` residual.** Both verified by execution, not read:

- Pre-fix `69bc97a2` raises `PermissionError: [Errno 13]` on another uid's `.git` under the probe root; at head the shipped CLI exits 0 and keeps the entry as `unreadable — cannot prove it is stale`. The retention-disabled `_inert_plan` path also completes.
- The mode-000 subdir case: pre-fix `age 9.00d, size 0 bytes, REMOVE`; at head `age 0.0, removable=false, "unreadable — cannot prove it is stale"`.

Six of six mutations RED per guard (drop `onerror=_blind` 3, ignore the unreadable flag 3, `_listable` always-ok 2, delete the `any_pid_answered` return-`None` 2, revert the C1 `try/except` 3, force `any_pid_answered` True 2), 53 passed at head. The fail-closed work that landed is genuinely pinned.

## C2 is NOT closed, and that makes this push net worse than what it replaced

`/host-proc/net` is a symlink to `self/net`, so `/host-proc/net/unix` resolves through the **reading process's** network namespace — not the host's. Inside the container the read **succeeds** and returns the container's own empty socket table. So `_bound_socket_paths()` returns an empty `frozenset` rather than `None`, `probe_gap` stays `""`, and the sweep proceeds as though nothing is held. Measured: `realpath(/host-proc/net) = /host-proc/<pid>/net`.

Controlled A/B with the namespace as the only variable, run inside the real `teatree-worker-1` pid+net namespaces, against a genuinely bound `AF_UNIX` socket in a directory whose whole tree is 9 days old — the exact case the socket source exists for, since a daemon's socket mtime never moves:

| | `_bound_socket_paths` | `probe_gap` | verdict on the live-socket dir |
|---|---|---|---|
| worker namespaces | **0 paths** | `""` | **`removable=True`, "stale scratch older than 3d"** |
| host namespaces (control) | 69 paths | `""` | `removable=False`, "open by a live process" |

Reproduced twice. The host really has 127 bound paths, 7 under the temp root.

**This is precisely the coupling failure flagged before this push.** C1's crash was the only thing keeping the sweep inert in the container. C1 is now fixed, so `_sweep_scratch` runs with `apply=True` inside `_execute_disk`/`_execute_ram`, ungated, at the default `scratch_retention_days = 3` — while the socket half of the liveness guard is silently blind in that venue.

## A correction to the earlier review's measurement, and it matters

The prior review reported that inside the container **every** per-pid source is unreadable for **every** pid (0/17). **That shape did not reproduce in any of four container configurations.** Measured in the worker's own namespaces: `fd/ iterdir` OK **65/275**, `map_files` 25/275, `cwd` 25/275, 206 fd readlinks resolved, `_held_paths()` returning 128–159 paths.

Two consequences:

1. **The new per-pid fail-closed branch is dead in production.** `any_pid_answered` is always `True`, so `if pids and not any_pid_answered: return None` never fires. Its ANY-pid granularity cannot see the partial blindness that actually obtains (239 of 275 pids unreadable).
2. **`deploy/README.md`'s new safety claim is false against measurement.** It asserts "every per-pid source … comes back unreadable for every pid" and concludes the guard's real contribution in the container is `probe_gap` rather than a working liveness check. It also correctly *names* the netns problem, then wrongly claims the per-pid blinding covers it. That is the load-bearing claim of the section this commit adds.

Treat the earlier 0/17 figure as unreliable; this measurement supersedes it.

## Blast radius — materially unchanged

Plan-only against the 1816/2428 (74.8%, 0.040 GiB) baseline:

| venue | removable | GiB | saved by liveness |
|---|---|---|---|
| worker namespaces | **1835/2431 (75.5%)** | 0.041 | **0** |
| host namespaces | 1834/2431 | 0.041 | 1 |

Not near-empty in the container — because the container is not blind, it is **partially sighted with a wrong-namespace socket table**.

## Also open

No test pins the socket source's namespace. Still no per-path removal record (~1835 removals leave zero path names). And `_sweep_scratch` still has no `allow_destructive_*` gate while `resource_pressure.py:35` documents L2 as "both non-destructive".

## Not established

The per-pid `EACCES` root cause (AppArmor `docker-default` vs dropped `CAP_SYS_PTRACE`). Whether any current real temp-root entry is protected by the socket guard alone — the live case found was saved by freshness, so one had to be planted. `apply()` was not run against the real temp root.

---

The narrow fix is to make `_bound_socket_paths()` read the same namespace as `proc_root` (e.g. resolve through `proc_root/<pid>/net/unix` for a host pid rather than `proc_root/net/unix`), and to fail **closed** when the namespace it read cannot be shown to match. Until then the PR stays in draft.

souliane:
**Heads-up for whoever resolves this conflict: it is a generated-docs overlap, and merges have a known trap there.**

This branch went `DIRTY` when #4323 merged (`4e90f719`). It is only **1 commit behind** and **15 ahead**, so the conflict is specifically with that one merge, not with accumulated drift.

It is **not** a migration-leaf collision — worth saying because that is the usual suspect and it is wrong here. This branch's newest migration is `0070_task_owner_driving_since`; `main` now carries `0071_collapse_modes_to_five_presets`. This branch adds no migration of its own, so the leaf is uncontested and `max_migration.txt` should resolve cleanly to `0071`.

The real overlap is in files both changes touch — `BLUEPRINT.md` and `docs/generated/*` (`cli-reference.md`, `management-commands.{md,json}`), plus the doctor modules this branch splits.

## The trap

The regeneration pre-commit hooks are **file-scoped and skip on a merge commit**. So resolving this by taking either side, or by hand-editing the conflict markers, leaves `docs/generated/*` holding pre-merge content — and then:

- `verify-gates` exits **0** (it does not regenerate), so the branch looks fine locally, and
- CI's `docs-drift` job goes **red**, on a PR whose source change was correct.

That is a wasted CI cycle and a confusing red, on a PR that has already spent a day held.

## Doing it safely

- Merge `origin/main` in — never rebase. A rebase voids any review verdict bound to the head, and this branch is meant to earn a fresh one.
- After the merge resolves, **regenerate the docs explicitly** rather than trusting the hook to have run, and commit that as part of the merge resolution. The generators are `scripts/hooks/generate_cli_output_snapshot.py` and the management-commands generator behind `docs/generated/`.
- Then confirm `docs-drift` locally rather than discovering it in CI.

## Status note

The prior verdict (1011, `hold`) was bound to `e630d5ec` and is **voided** by the push to `5d6bab69`. So this PR now has no binding verdict and will need a fresh independent cold review once it is conflict-free and green. It correctly remains a draft until then.

souliane:
**Status, so this does not read as neglected: this PR is green and ready, and is blocked by a factory defect rather than by anything in the change.**

Current state at `59f9c165`:

- both CI blockers from the last review are cleared (the ladder tests pinned RED→GREEN against CI's exact `probe_gap`; the doctor `app.py` split 502 → 244 LOC at the transport/policy seam), pushed with all 8 push-stage gates passing;
- the conflict with `main` is resolved — `main` moved three times tonight and the merge is in;
- **CI is fully green**;
- the prior hold (verdict 1011, bound to `e630d5ec`) is **voided** by the pushes, so nothing is holding it.

What it needs is one independent cold review. It cannot get one.

## Why

Both of this PR's ticket rows are in terminal states:

```
author   ticket 755  state=shipped
reviewer ticket 756  state=review_posted
```

`reviewer_task_orphaned` completes every open `reviewing` task on a terminal ticket, so a review enqueued on either row is reaped with zero attempts and no verdict — and there is no third row. Enqueueing one anyway would be churn: a task on a terminal row is reaped within minutes, well before the current review queue would reach it.

Nor can the state be moved to create a reviewable row: `SHIPPED` is in neither `_RECONCILE_SOURCE_STATES` nor `rework`'s source set, and `reopen` is not on the CLI allow-list; `REVIEW_POSTED` has no exit at all.

This is tracked, with the reproduction and the fix ordering, on the re-review issue and the FSM issue. The single change that unblocks this PR is requiring the FSM-terminal reap ground to also confirm the PR is closed — the forge-state ground (`get_pr_open_state` says MERGED/CLOSED) is correct and would be untouched.

## Deliberately not worked around

Three things were considered and rejected rather than overlooked:

- **Un-drafting it.** Draft status is how the hold is enforced here; un-drafting a PR that has no verdict would remove the guard rather than satisfy it.
- **Merging on green alone.** Green CI is not a verdict. No merge without an independent cold review at the live head.
- **Reviewing it from the orchestrating session.** Maker≠checker is satisfiable, but promptness accelerates dispatching, never verdicts — and this PR's last review found proven data loss, which is precisely the kind of thing a rushed self-review misses.

So it waits, green, until the reap condition is narrowed. Recording it here so the next person to look at the board does not re-derive all of this.

Issue #2009, opened 2026-06-06
Repair-loop robustness: visible per-phase iteration budget + stall detection


souliane:
User-approved. Two gaps vs verify->diagnose->patch->re-verify repair-loop design: (1) no visible per-ticket/per-phase iteration budget — retries are time-based (24h stale-task expiry) and buried in queue_drain; add an iteration counter on TaskAttempt, surface 'attempt N/max' in ticket show, fail with MaxIterationsExceeded at the cap. (2) no stall detection — track an error fingerprint (hash of terminal_reason) per phase; two consecutive identical fingerprints => IterationStalled, escalate to the user instead of re-queueing. Robustness/conformance-class, not a new feature surface.

Issue #4308, opened 2026-08-07
Reviewing tasks complete with exit 0 having recorded no verdict, no PR comment and no review


souliane:
## What happens

Reviewing-phase tasks complete with exit 0 and produce **no artifact of any kind** — no `ReviewVerdict` row, no PR comment, no PR review. The task ledger reads as four successful reviews; the merge guard receives nothing.

## Evidence

Ticket 756 (the PR-row ticket for PR #4306) on 2026-08-07:

```
1570 completed reviewing tk 756 created 10:18:05
1571 completed reviewing tk 756 created 10:43:24
1572 completed reviewing tk 756 created 10:58:28
1574 claimed   reviewing tk 756 created 11:08:01
```

All completed rows carry `exit_code = 0`. Across all four:

- `ReviewVerdict.objects.filter(pr_id=4306)` holds only verdicts recorded by an attended session and one separately-dispatched reviewer — none from these tasks;
- the PR's comment list holds only that session's three comments;
- the PR's review list is empty.

Task 1572's own attempt result records the mechanism for at least that one:

```
result = {'summary': 'the row had already completed — lease lost for task 1572:
          the row is already completed — the attempt has nothing left to hand over'}
```

So the attempt woke, found its row already `completed`, and exited 0 having handed over nothing.

## Why it matters

`exit 0` plus a `completed` row is the success signal every consumer reads. A reviewing task that satisfies both while recording no verdict is indistinguishable from one that reviewed and approved — except that nothing downstream can tell, because there is no verdict to read either way. The visible effect is a review lane that appears to be working while the merge guard's only inputs come from elsewhere.

On this PR that mattered concretely: the change recursively deletes ~1987 paths, and the only holds standing against it were recorded by an attended session. Had none been attending, four "successful" reviews would have left the PR with no verdict at all.

## What I did NOT establish

Stated so this is not read as more than it is:

- **Whether four tasks is itself wrong.** The PR's head moved three times in that window (`af7c92ec` → `6bf279a7` → `c573ef2e`), so one review per push is a sufficient and innocent explanation for the count. The defect claimed here is the empty outcome, not the number.
- **Whether all four share task 1572's cause.** Only 1572's attempt carries the lease-loss summary; the other three record an empty result, so their path is unknown.
[...]

PR #4309, opened 2026-08-07
fix(core,agents): the stuck-reaper must not duplicate a live run, and a failed attempt records its spend (#4164)

### LINKED ISSUE #4164: The stuck-reaper fails a task whose process is still alive, turning one memory event into a duplicate execution
Bottleneck analysis, 2026-08-04. This is the **amplifier** that turns memory pressure into 31-40% of all agent-hours being spent twice. Sibling of #4163 (the governor never reads RAM).

## The mechanism

```
RAM pressure
  -> the agent event loop stalls past the 900 s lease
     (heartbeat 60 s x 15 margin, src/teatree/agents/headless.py:85,98)
  -> reclaim_orphaned_claims returns the task CLAIMED -> PENDING, nulling the lease
  -> reap_stuck_headless_runs sees status != CLAIMED, calls row.set_failed()
     and re-enqueues (src/teatree/loops/timer_reconciler.py:279-285, 323-329)
  -> the ORIGINAL run next heartbeat raises LeaseLostError and interrupts itself
  -> a second agent redoes the work from scratch, on the same worktree
```

**`set_failed` marks the DB row only — it does not kill the process.** So the reaper does not reclaim a dead run; it *creates* a duplicate of a live one, then the live one aborts. The cost of a memory blip is not one lost run — it is one lost run **plus a full re-execution**, which raises memory demand and makes the next blip likelier.

## Measured cost

| | |
|---|---|
| Agent-hours reaped as `_StuckHeadlessRunError` | **31-40%** of 111.5 agent-hours in a 42.7 h window |
| Queue rows per distinct task | **3.8** |
| Rows returning `not claimable (claimed elsewhere or terminal)` | **756 of 1054 (72%)**, after a median 24 min queued |
| `reviewing` tasks per merged ticket | **6.87** (vs 1.38 coding) |
| Failed tasks that are `lease_lost` | **119 of 228 (52%)** — 103 of them in `reviewing` |

Worked example, task 1000: `54 min FAILED -> 35 min FAILED -> 156 min SUCCESSFUL`. **245 slot-minutes for 156 minutes of product.**

## It also corrupts the failure record

Only **2 of 228** failed tasks carry a `success` attempt row — but **120 of 228** had their ticket advance *after* the task was created, including **50 of the 119** `lease_lost`. So at least ~42% of `lease_lost` are **false failures where the work actually landed**. That inflates repair burn, drives further re-dispatch, and makes the failure taxonomy untrustworthy for exactly the diagnosis it is needed for.

A failed attempt also records `cost_usd = 0`, so the $942 recorded spend covers successes only. True spend is higher and currently unmeasurable.

## The fix

**Probe liveness before declaring a run stuck.** A lease that has lapsed is evidence about the *lease*, not about the process. Options, roughly in order of preference:

1. **Liveness check before `set_failed` + re-enqueue.** If the process is alive, do not fail the row — extend or re-acquire. Only reap when the owner is genuinely gone.
2. **Lengthen the lease under measured memory pressure**, so a stall caused by thrash does not read as death. This composes with #4163 rather than competing.
3. **If a duplicate is unavoidable, kill the original first.** Today both run: the reaper enqueues a second agent while the first is still executing against the same worktree — which is also a concurrent-worktree hazard, not only wasted capacity.

Whatever is chosen, the invariant is: **the reaper must not create a second executor for work that is still executing.**

## Acceptance

- A regression test RED on current `main`: a task whose lease lapsed while its process is ALIVE is not re-enqueued, and the live run completes normally.
- A reaped-then-completed run is not recorded as `lease_lost` when the work landed — the failure taxonomy must stop counting successes as failures.
- The concurrent-worktree hazard is closed or explicitly documented as impossible.

Related: #4163 (the binding constraint this amplifies), #4100/#4110 (the earlier phantom-lease-loss work — this is the remaining half: that fixed the *recording*, this is the *duplicate execution*).

---

## Extension, 2026-08-04 — the cost half, measured: 916 attempts provably spent tokens and recorded none

Filed into this issue rather than a new one (backlog went 12 → 62 open in a day; the standing rule is to extend a suitable existing ticket). This belongs here because the section above already asserts it — *"A failed attempt also records `cost_usd = 0`, so the recorded spend covers successes only"* — with a number that is now stale, and because `lease_lost` is the single largest bucket of the loss. **Correction: recorded spend is $3,062.70 since 2026-07-13, not $942.**

The reaper fix above stops the duplicate execution. This half stops the duplicate execution's cost from being invisible.

### Usage is written in exactly one place

`src/teatree/agents/attempt_recorder.py:176-194` — `record_result_envelope`, the **success** path — is the only site that populates `input_tokens`, `output_tokens`, `cache_read_tokens`, `cache_write_tokens`, `cost_usd`, `num_turns`, `lane`, `cost_is_estimated`.

The two failure recorders take **no `usage` argument at all**:

- `src/teatree/agents/attempt_recorder.py:434` — `_record_failure(task, *, error, result=None)`
- `src/teatree/agents/headless.py:596` — `_record_failure(task, *, exit_code=1, error="", result=None)`

So every schema rejection, lost lease, evidence-gate refusal and harness crash discards spend that has already been billed.

### Confirmed by outcome, not by reading the code

Measured against the live control DB, 9,784 `TaskAttempt` rows:

| attempt outcome | rows | rows carrying tokens |
|---|---|---|
| success | 1,567 | **1,558** |
| failure | 8,217 | **0** |

Not one failed attempt in the table's history carries a token count.

### A NULL is not evidence of no spend

8,227 rows have a blank `lane`. They are not all free:

| bucket | rows | did it spend? |
|---|---|---|
| `credential_exhausted` / `usage_limit_parked` | 7,301 | **No** — parked before a billed turn. NULL is honest here |
| `lease_lost` | 462 | **Yes** |
| `evidence_missing` | 181 | **Yes** |
| `harness_crash` | 115 | **Yes** |
| `recording_refused` | 62 | **Yes** |
| `no_result_envelope` | 44 | **Yes** |
| tail (turn ceiling, other) | ~52 | **Yes** |
| ambiguous | 9 | undetermined |

**916 rows did spend and recorded nothing** — 9.4% of the table is a known false negative. Every one of those five "yes" categories is a failure that occurs *after* the agent has run turns.

The consequence is a reporting one, and it is the same shape epic #4041 names: **a NULL token field means "not recorded", not "no spend"**. Cost attribution is understated by an unknown but non-zero amount, and any per-phase or per-lane cost analysis built on this table is reading successes only. Recorded spend is all SDK-reported (`cost_is_estimated=0`), so the true figure is strictly higher than $3,062.70.

This compounds the duplicate-execution finding above directly: the re-executions this issue is about are *exactly* the attempts whose cost is discarded, so the measured waste is understated by construction.

### The fix is a signature change, not new plumbing

Both the lane and the parsed usage are already in scope at every failure call site — see `src/teatree/agents/headless.py:311-336`, where `_outcome_failure` calls `_record_failure` four times with the outcome in hand. Passing `usage=` into the two `_record_failure` functions and writing the same fields the success path writes is the whole change.

Where a failure genuinely happened before any billed turn (the 7,301 parked rows), the fields must stay NULL — the point is to distinguish "not recorded" from "no spend", not to write zeros everywhere. A zero would be a *worse* lie than a NULL, because it reads as a measurement.

### Acceptance criteria

1. **A crashed attempt records non-zero tokens.** Drive a run that produces usage and then fails (harness crash / lost lease / evidence-gate refusal), and assert the resulting `TaskAttempt` carries `input_tokens > 0` and a non-empty `lane`. **This is the anti-vacuity anchor: it must be observed RED on today's code, where the value is unconditionally NULL because the recorder cannot accept it.** A test that passes today guards nothing.
2. **A pre-turn park still records NULL, not 0.** A `credential_exhausted` / `usage_limit_parked` attempt that never billed a turn leaves the token fields NULL. Both directions asserted — a fix that writes zeros everywhere destroys the very distinction this asks for.
3. **No failure path can silently drop usage.** Every call site that constructs a `TaskAttempt` with a non-empty `error` either passes `usage=` or names, in code, why no usage exists. Pin it with a conformance check in `tests/conformance/` (the `test_consumer_caller_walk.py` shape), so a third failure recorder cannot appear without one.
4. **Re-running the measurement above after the fix distinguishes the two populations**: blank-lane rows split into "parked, never spent" and "spent, recorded", with nothing left in "spent, not recorded". The 916 already-written rows are **not** backfillable — the usage is gone — so the criterion applies to rows written after the change, and the historical undercount should be stated once wherever cost is reported rather than silently corrected.

### Not established

- The 916 figure is derived by classifying `error` text into buckets; each bucket was confirmed to be a post-turn failure mode, but individual rows were not audited one by one. Treat 916 as a floor on the false negatives, not an audited count.
- Whether any consumer currently reports cost in a way that would visibly change was not traced.

souliane:
Closes #4164.

Two halves of one issue: the reaper turned a memory blip into a duplicate execution, and the duplicate's cost was then discarded on the way out.

## The reaper (the titled defect)

A lapsed lease is evidence about the LEASE, not about the work. Under memory pressure the headless runner's event loop stalls past its 900s lease while the agent is still producing, and three sweeps read that as a dead worker:

- `reclaim_orphaned_claims`: CLAIMED → PENDING, nulling the lease
- `reap_stale_claims`: CLAIMED → FAILED
- `reap_stuck_headless_runs`: `set_failed` + `execute_headless_task.enqueue`

`set_failed` marks the DB row only; it does not kill the process. So the reaper did not reclaim a dead run, it CREATED a duplicate of a live one — the re-enqueued job's claim CAS treats an expired lease as claimable, so the duplicate won the claim and the original aborted `LeaseLostError`, briefly putting two agents on one worktree.

The evidence is a registry, not an inference. `t3 worker` drains `default` and `loops` as THREADS of one process, so the sweeps and the run they would reap share an address space:

- `core/claim_liveness.driving(task_pk)` wraps `run_headless`, marking the task as executing for the whole drive.
- `owner_is_executing` is True only when the recorded owner IS this process and this process is inside `driving()` for that task. Everything else — an owner elsewhere, an unattributable namespace (#4253), an unrecorded pid, a task nothing here is driving — is False, i.e. today's lease-only verdict. The guard can only WITHHOLD a reap it can prove premature, never widen one.
- `Task.owner_pid` / `owner_pid_namespace` record which process holds the claim, written by `claim` / `claim_next_pending` / `renew_lease` and cleared with the claim.
- All three sweeps consult it. All three are needed: guarding only the reclaim leaves the reaper re-enqueuing on the CLAIMED-with-lapsed-lease row.

A first cut inferred liveness from an alive owner pid instead. Two existing reclaim tests caught it: in a single-worker deployment the pid is trivially alive whether or not the job it claimed still exists, so a crashed job's row was held for as long as the inference lasted. The registry is a fact about the work, so a crash releases the row to ordinary lease-expiry recovery immediately.

Concurrent-worktree hazard: closed at this site. A stalled-but-driving run is left entirely alone — nothing failed, nothing enqueued — so no second executor.

## The spend (the same issue's cost half)

Usage was written in exactly one place — `record_result_envelope`'s SUCCESS path. Both `_record_failure` functions took no usage argument, so every post-turn failure discarded tokens already billed. Of 9,784 `TaskAttempt` rows, 1,558 of 1,567 successes carry tokens and 0 of 8,217 failures do; 916 of those provably spent. This compounds the half above directly: the re-executions are exactly the attempts whose cost is discarded.

- `attempt_recorder.usage_fields(usage) -> SpendColumns` is the one mapping the success and failure recorders share, so they cannot diverge again.
- Both `_record_failure` functions, `_record_interrupted_attempt` and `_record_landed` take `usage=`. `_outcome_failure` samples it ONCE from the same `ResultMessage` every branch classifies, so the recorded spend can never describe a different run than the recorded error.
- `usage=None` writes NOTHING, leaving the columns NULL. A pre-turn park never spent, and a zero would be a worse lie than a NULL because it reads as a measurement. Both directions are asserted.
- `tests/conformance/test_attempt_records_its_spend.py` walks `src/` for every `TaskAttempt` write and requires `usage=` or a `# no-usage: <why>` pragma, so a third recorder cannot appear without one. It carries its own control.

## 1. BLUEPRINT § alignment

§4 Domain Models (`Task` gains `owner_pid` / `owner_pid_namespace`; `TaskAttempt` spend columns now written on the failure path too) and §5 Agent Execution (the headless drive and the lease sweeps that reap it). No BLUEPRINT prose contradicts the change: the claim/lease contract is described by the model, not restated in the document, and no new phase, state or invariant is added.

## 2. FSM phase boundaries

No new `Ticket.State` or `Worktree.State` transition. The change is `Task` FSM-adjacent only: it makes three existing sweeps WITHHOLD a transition they were taking prematurely — `reclaim_orphaned_claims` (CLAIMED → PENDING), `reap_stale_claims` (CLAIMED → FAILED) and `reap_stuck_headless_runs` (`set_failed` + re-enqueue). The withheld row keeps its current state; no transition method or FSM condition is changed, only the candidate set each sweep acts on.

## 3. Extension-point contracts

n/a — no `OverlayBase` hook, scanner registration, hook-router surface or `*Backend` Protocol is touched. `git grep -l 'OverlayBase\|register_scanner\|hook_router\|MessagingBackend\|CodeHostBackend'` returns no file in this diff, so no overlay needs updating.

## 4. Component boundaries

`src/teatree/core/claim_liveness.py` — an ORM-free predicate layer, the flat sibling of `core/loop_lease_liveness.py` whose pid-attribution seam it reuses. It must be importable by `core/models/` (`task_claim`), `core/managers*`, `core/tasks` AND `teatree.loops` with no cycle, so no existing subpackage can hold it (`modelkit/` is a zero-dependency tach node and cannot take the `loop_lease_liveness` edge). The flat-core ratchet is raised 109 → 110 with exactly that justification in code. The spend half stays in `agents/attempt_recorder.py`, where both recorders already live.

## 5. Dependency direction

New edges, all downward: `core.claim_liveness → core.loop_lease_liveness`; `core.managers_task_sweeps`, `core.managers`, `core.models` and `teatree.loops` → `core.claim_liveness`. Declared in `tach.toml` (`layer = "platform"` for the new module) and reflected in `docs/dependency-graph.md`. `uv run tach check` — see the verification section below.

## 6. Test surface

- `tests/teatree_core/test_claim_liveness.py` — the predicate itself: `owner_is_executing` is True only inside `driving()` for this process's own pid, and False for an owner elsewhere / an unattributable namespace / an unrecorded pid / an untracked task.
- `tests/teatree_core/test_managers_task_sweeps.py` — a task whose lease lapsed while its process is still driving it is NOT reclaimed and NOT failed by either sweep; a genuinely orphaned one still is.
- `tests/teatree_loops/test_timer_reconciler.py` — `reap_stuck_headless_runs` neither sets the row failed nor re-enqueues while the owner is driving (the acceptance criterion's "not re-enqueued, the live run completes normally").
- `tests/teatree_agents/test_attempt_recorder_spend.py` — a post-turn failure records `input_tokens > 0` and a non-empty `lane`; a pre-turn park leaves the columns NULL, not 0. Both directions.
- `tests/conformance/test_attempt_records_its_spend.py` — no third recorder can write a `TaskAttempt` without `usage=` or a `# no-usage: <why>` pragma; carries its own inline control asserting the walk detects an unaccounted write and does not honour a pragma parked elsewhere in the module.

## 7. Resilience invariants

No new external write — the change only WITHHOLDS DB writes the sweeps were making and ADDS columns to a row already written. Against #1192: **idempotency** is the property being restored (the whole defect is a duplicate execution of one claim); **verify-by-re-read** is unchanged — the claim CAS is still the authority; **fallback-transport** — an unprovable liveness answer degrades to today's lease-only verdict, so the sweeps keep working when the registry says nothing; **heartbeat** — untouched, and the guard exists precisely because a stalled heartbeat is not death; **sub-agent return contract** — untouched, `StructuredResult` recording is the surface the spend half writes through.

## 8. Identity and key normalization

The identity is "which process holds this claim", and its canonical form is the QUALIFIED pair `(owner_pid, owner_pid_namespace)` — a bare pid is meaningless outside the namespace it was recorded in (#4253), since each service in the deployment has its own and the same integer names a different process or none. `ClaimOwner` carries both together and `OWNER_COLUMNS` names them as one unit so a `.values_list()` caller cannot omit one and hand the predicate a blank field that reads as an absent fact. No strip/split/removeprefix anywhere: an unattributable namespace is answered False (fail-safe to today's behaviour), never normalized away to make a comparison succeed.

## 9. Behavior preservation / capability deletion

Nothing is deleted; both halves are strictly additive at their sites.

- The three sweeps keep every existing behaviour — the repair-loop budget filter (`MaxIterationsExceeded` / `IterationStalled`, #2009), the lease-expiry predicate, the failure taxonomy. The new check only removes rows from the candidate set, so the guard can WITHHOLD a reap, never widen one. An owner elsewhere, an unattributable namespace, an unrecorded pid and an untracked task all fall through to the pre-change verdict.
- No must-block test inverted to must-not-block. The two existing reclaim tests that pin the crashed-job recovery path stayed green and in fact caught the rejected pid-inference first cut — a crashed job's row is still released to ordinary lease-expiry recovery.
- `record_result_envelope`'s SUCCESS path writes exactly the same columns as before; they were moved verbatim into `usage_fields` and splatted back in. No privacy/leak/security matcher is touched.

## 10. Removability / harness-vs-data

**Removable.** Deleting `core/claim_liveness.py` and its four call sites reverts to the pre-change lease-only verdict with no cascade: the three sweeps fall back to the predicate they already had, and `Task.owner_pid` / `owner_pid_namespace` become unread columns (a follow-up migration, not a blocker). `driving()` is a context manager wrapping one call — removing it changes nothing else. The flat-core ratchet returns 110 → 109. **Harness, not data.** The decision is a process-liveness fact resolved at reap time from an in-memory registry, not a per-item policy a table could express: there is nothing per-task, per-overlay or per-phase to configure, and the answer is only ever "yes I am driving this right now" or "I cannot tell". Putting it in config would let an operator turn a correctness guard into a duplicate-execution bug; there is deliberately no flag.

## Verification

Both halves observed RED before the fix: forcing `owner_is_executing` to a constant False fails 5, and forcing `usage_fields` to a constant empty fails 3 while the two NULL-direction tests stay green (so they are not mirrors of each other).

- **Multi-process deployments get no protection** (assumed): the registry answers only for THIS process, and nothing here can see whether another process is still driving. That is today's behaviour, not a regression, but it is a real limit — assumed acceptable because `t3 worker` runs one process with per-queue threads. Open if the deployment ever splits the queues across processes.
- **The registry wraps `run_headless` rather than the whole of `execute_headless_task`** (assumed): the stall class this issue names is the agent event loop, and wrapping the outer task pushed `core/tasks.py` over its 500-LOC module-health cap. A stall in provisioning or an FSM transition is still unguarded. Open.
- **`Task.claim`'s CAS still treats an expired lease as claimable** (decided): a registry lookup cannot run inside a SQL predicate, and with the three sweeps guarded nothing enqueues a duplicate for a live claim.
- **`harness_crash` spend stays UNRECORDED** (open), stated in code at the site (`core/models/task.py` `complete_with_attempt`). The exception escapes the drive before a `ResultMessage` exists, and core cannot import the agent layer to parse one.
- **The 916 already-written rows are NOT backfillable** (assumed) — the usage is gone — so acceptance criterion 4 applies to rows written after this change. Whether any cost consumer visibly changes was not traced. Open.
- **Acceptance criterion "a reaped-then-completed run is not recorded as `lease_lost` when the work landed"** (assumed satisfied indirectly): this PR removes the premature reap that produced those false `lease_lost` rows, rather than adding a post-hoc reclassifier. The taxonomy is corrected at the source; no separate re-labelling pass is added.

## Update — cold review fix, 608722882 → d0db5b0dd

Independent cold review of `608722882` found a live blocker: a reactive usage-limit park
(`park_or_rotate_on_limit` → `park_task_on_limit` / `_rotate_or_none` → `_record_park`) never
received the `usage` `_outcome_failure` already sampled from the SAME `ResultMessage`. The
`_record_park` pragma read `a park happens BEFORE any turn is billed` — true only for the
genuinely pre-dispatch admission-guard park, not for a limit reported ON a result the SDK
already returned, and directly contradicted by `_outcome_failure`'s own docstring ("every
branch here is POST-turn"). Reproduced: a `ResultMessage` carrying `input_tokens=4200` parked
with `input_tokens is None`. The new conformance gate checked a pragma *existed*, not that it
was *true*, so it certified the exact bug class it exists to catch.

Fixed by threading `usage` through every reactive park call site, leaving the genuinely
pre-dispatch ones (the admission guard, a pre-turn `CredentialError`) at `usage=None` — both
directions asserted. `_record_park` now writes `usage_fields(usage)` like every other
`TaskAttempt` recorder; the false pragma is gone.

Threading `usage` pushed `park_task_on_limit`/`park_or_rotate_on_limit` over ruff's 5-arg
ceiling (PLR0913). Fixed architecturally rather than suppressed (the repo's
`ac-django-no-complexity-suppressions` gate blocks new `noqa` here): `LimitSignal` bundles
`sdk_resets_at` + `usage`, both sampled from the same triggering result. `_rotate_or_none`'s
`scope` param was also redundant (always `task.ticket.overlay or ""`, already computed that
way by its only caller) — removed rather than bundled.

**Verification.** New anti-vacuity test observed RED pre-fix (`assert None == 4200`). Five more
usage-carrying assertions across the park chain's other call sites (rotation, all-exhausted-via-
rotation, direct all-exhausted) confirmed RED under a targeted mutation while the pre-existing
NULL-direction tests stayed green. `ty` and `ruff` clean, no `noqa` added. Full local lane: 13
failed / 40775 passed — the 13 are the known pre-existing baseline (confirmed by name match
across three separate runs; a 52- and 73-failure middle run were proven load-induced by an
isolated re-run passing 92/92, per the box-load lesson). `t3 tool verify-gates` exit 0 (commit +
push). CI green at `d0db5b0dd` (23 pass, 0 fail, 3 skipping).


souliane:
**HOLD recorded as verdict 953** at `60872288`, transcribed from a review that had already been performed and then lost.

Task 1575 cold-reviewed this PR, reached a considered HOLD, and emitted a complete verdict envelope — which was never written to the verdict table. The PR was sitting `CLEAN`, all required checks green, with no recorded verdict at all, while a blocking finding existed in a task attempt row that no merge guard reads. I have persisted that envelope verbatim under its original reviewer identity (`cold-reviewer-task-1575`); the verdict and every finding below are that reviewer's, not mine.

**The blocker (medium)** — `src/teatree/agents/usage_window.py:281`. The post-turn usage-limit park discards the spend its own run billed, under a factually false no-usage pragma that this PR's new conformance gate then certifies. Reproduced live: a `ResultMessage` carrying `input_tokens=4200` through `_outcome_failure`'s limit branch records `input_tokens=None`. `limit_autorecovery_enabled` defaults True, so this is the default path — which matters especially for a PR whose second half exists to make a failed attempt record its spend.

**Five lows**, abbreviated: the withholding has no upper bound because heartbeat, watchdog and the `asyncio.wait_for` ceiling share the drive's event loop (`headless.py:147`); `_driving` is a set rather than a refcount, so nested in-process drives clear the flag early (`claim_liveness.py:71`); the conformance walk's `_CARRIES_SPEND` substring match on `usage=` lets a literal `usage=None` satisfy the gate (`test_attempt_records_its_spend.py:23`); the `_LEASE_SECONDS` rationale comment still describes the pre-fix behaviour in terms the PR body contradicts (`headless.py:102`); and one finding recorded explicitly as checked-clean — migration 0066 takes the next free slot with no collision against the other open PRs.

**Credit as recorded by that reviewer:** both mutation controls reproduced RED (7 and 3 tests), 83/83 PR tests green, `t3 tool verify-gates` exit 0, CI 0 FAILURE — the first `dash-e2e` red was diagnosed as the sqlite `teatree_loop_schedule` teardown flake and passed on retry.

Green CI is not in dispute here; the medium finding is about a default-ON path that the new gate certifies as clean.

souliane:
Third independent cold review at `60872288` — **HOLD**, recorded as verdict 956. This one reaches a different and more serious finding than the two before it, so it is worth reading past the verdict line.

CI is not in question: 43 check-runs, 0 pending, 0 failed, 16/16 test shards. `dash-e2e` read `failure` at 11:13:21Z and reads `success` now at the same SHA — a flake that re-ran green, not this PR's doing.

## F1 — MAJOR. Two of the three guarded sweeps run in a different process, so the guard cannot fire there. The titled duplication survives.

`src/teatree/core/claim_liveness.py:33` — `_driving` is a module-level, in-process set. The PR's load-bearing premise (stated in the module docstring and again under "Open questions & assumptions": *"`t3 worker` runs one process with per-queue threads"*) holds for `reap_stuck_headless_runs` and is **false** for the other two.

The measured chain: `loops/timer_chains.py:369` → `run_deadlined_tick` spawns `python -m teatree loops_tick --loop <name>` as a **child process** (`loops/deadlined_tick.py:47`). Sampling the live worker container for two minutes showed seven distinct `loops_tick` processes, all `PPID 7`. That subprocess reaches `loop/tick.py:158 _reap_stale_task_claims` → `loop/tick_recovery.py:51 run_boot_sweeps` → `core/worktree/recovery_sweeps.py:35-36`, which calls `reclaim_orphaned_claims()` **and** `reap_stale_claims()`. No worker-process caller of `run_boot_sweeps` exists.

Reproduced by execution with the fix fully in place:

- **Control**, in the drive's own process: `reclaim_orphaned_claims()` returns `0`, the row stays `CLAIMED`. The guard works.
- **Deployed path**, empty registry (i.e. the tick subprocess): `reclaim_orphaned_claims()` returns `1`, row → `PENDING`; then, back in the worker process **while still `driving(pk)`**, `drain_headless_queue_body()` enqueues a second `execute_headless_task` and `Task.claim(...)` **wins** the row the first agent is still executing. Two agents, one worktree — the exact defect, unchanged.

This is the path this box actually measured: reclaim releases `CLAIMED → PENDING` about 47s after lease expiry. `reclaim_orphaned_claims` is that sweep, and it runs where the registry is blank. `reap_stuck_headless_runs` is the only one this PR protects.

## F2 — MAJOR. The one line that makes the guard fire in production is unpinned.

`src/teatree/agents/headless.py:154`. Deleting the `with driving(task.pk):` wrapper turns **0 of 410** tests red. Every existing test injects `driving()` directly in its own body; none drives `run_headless` and asserts the registry is populated. The guard is protected only by ruff's unused-import check.

## Minors

No upper bound on withholding — heartbeat and `LoopWatchdog` both live inside the same stalled asyncio loop, and doctor self-heal only fires when there is no live worker, so a wedged-but-alive drive pins the row indefinitely where it previously recovered after lease+grace (`claim_liveness.py:78`). And recording failure spend silently changes two configured ceilings that sum `TaskAttempt.cost_usd`/`num_turns`; no live impact today, since the deployed `ticket_budget_max_cost_usd`, `watchdog_max_cost_usd` and `watchdog_max_turns` all read 0 (`headless_budget.py:64`).

## What is genuinely right, verified by execution

**The spend half is sound and well pinned.** Live DB: 10,006 `TaskAttempt` rows — 1,690 of 1,743 successes carry `input_tokens`, **0 of 8,263 failures do**. The PR's own measured claim reproduces exactly. The `usage=None → NULL, never 0` distinction is asserted in both directions and the conformance walk goes red on its own.

**The liveness predicate survived every attack.** It does not probe a pid: the evidence is pk-membership, with the pid as a cheap pre-filter. Pid reuse cannot false-withhold; zombies are irrelevant; a container recreate clears the registry with the process, degrading to today's lease-only verdict. Mutation M2 (predicate → constant `True`) turns 19 tests red, so it cannot silently widen into a permanent hold. The rejected first cut — pid-liveness inference — is documented and was correctly rejected.

**18 mutations run, with counts.** M1 7, M2 19, M3 1, M4 1, M5 1, **M6 0** (F2), M7 16, M8 4, M9 2, M10 3, M11–M15 1 each, M16 2, M17 1, M18 1.

## Not established

Whether the ~916 historically-unrecorded rows have any downstream consumer that visibly changes (the PR leaves this open too), and whether a multi-worker deployment is ever intended — which would widen F1 further.

---

This is the third hold on this head. The two earlier ones (953, 954) both converged on the usage-limit spend discard; this one confirms the spend half is fine as written and holds on the reaper half instead. Those are compatible readings of different halves, not a contradiction — but F1 is the one that decides the PR, because it says the titled defect is still reachable on the deployed path.

souliane:
Head moved `60872288` → `d0db5b0d`, which voids verdicts 953, 954 and 956 at once. Hold re-armed as verdict 957.

Stating what this is: a **carry-forward**, not a fresh independent review of the new code. It exists so the PR is not sitting unguarded while the real review happens.

## Addressed by this push — credit where due

`fix(agents): a reactive usage-limit park must not discard the turn it billed` genuinely targets the finding that three separate reviewers converged on. `usage` is now routed through a new `LimitSignal` into `park_or_rotate_on_limit` (`usage_window.py` +101, the call site at `headless.py:350`), with +89 lines of `test_usage_window.py` and +34 of `test_attempt_recorder_spend.py` behind it. That was the high in 953/954 and the medium in 956, and it looks like the right shape.

I have **not** re-verified it by execution at this SHA. A fresh review should confirm the spend columns are actually written on the limit path rather than that the plumbing merely exists — that distinction is what the original finding was about.

## Not addressed — the two majors that decide the PR

Verdict 956's blockers are untouched, and this is verifiable from the delta rather than inferred:

**F1 — the guard cannot fire where the defect happens.** `src/teatree/core/claim_liveness.py` does not appear in this push at all. `_driving` is a module-level in-process set, but `reclaim_orphaned_claims` and `reap_stale_claims` run inside the `loops_tick` **child process** (measured: seven such processes, all `PPID 7`), where that registry is empty. Reproduced end to end at the prior head: with the fix in place, the tick-path sweep still reclaims a live claim to `PENDING`, and `drain_headless_queue_body` then hands a second executor the same row while the first is still driving. The PR body's premise that all three sweeps share the drive's address space is false for two of them, so the titled duplication survives.

**F2 — the line that makes the guard fire is unpinned.** Deleting `with driving(task.pk):` turns **0 of 410** tests red. The `headless.py` delta in this push is the `LimitSignal` spend plumbing only; it does not touch the wrapper, and no test file for `claim_liveness` changed.

Also still open, minor: no upper bound on withholding — heartbeat and `LoopWatchdog` share the stalled asyncio loop, and doctor self-heal fires only when no live worker exists, so a wedged-but-alive drive pins the row indefinitely where it previously recovered after lease+grace.

## What would move this

F1 needs the liveness signal to be readable from the process that actually runs the sweep, or the sweeps moved to where the registry lives — either is a design decision worth making deliberately rather than patching. F2 needs one test that drives `run_headless` and asserts the registry is populated, so the guard's own activation is pinned.

The spend half being fixed is real progress and shouldn't be relitigated. It just isn't what the hold is about.

souliane:
**HOLD recorded as verdict 959** at `d0db5b0d`, transcribed from a pass-3 review that had already been performed and whose verdict was never written to the table. The verdict and every finding below belong to `cold-reviewer-4164-pass3`, not to me; I only persisted the envelope it emitted. A shipping task was queued on this ticket at the time, with the PR `CLEAN` and no verdict at the live head — so this hold existed only inside a `TaskAttempt.result` that no merge guard reads.

## The spend blocker IS fixed, and pinned

That reviewer confirmed it: **5 RED** under mutation of `usage_fields` in `_record_park`, with 31 NULL-direction tests still green. `verify-gates` exit 0. CI 40 success / 3 skipped / 0 failure. This closes the finding that verdicts 953, 954 and 956 all converged on. Please do not re-litigate it.

## HIGH — the claim-liveness guard is inert where the sweeps actually run

`src/teatree/core/claim_liveness.py`. Reproduced live: in the child process the registry is empty and the predicate returns `False` unconditionally, so **2 of the 3 sweeps the PR claims to guard** get no protection. `reclaim_orphaned_claims` and `reap_stale_claims` have the `loops_tick` subprocess as their only deployed caller, so the titled duplicate-execution path survives — reclaim to `PENDING`, then drain hands the row to a second executor.

This is an **independent third confirmation** of the same defect: verdict 956 reproduced it end to end from a different angle, and this pass reproduced the registry emptiness directly in a spawned child. Two reviewers, different methods, same conclusion.

## MEDIUM — anti-vacuity gap on the line that arms the guard

`src/teatree/agents/headless.py:156`. Deleting `with driving(task.pk):` leaves **1380 tests passing, 0 RED**, because every test injects `driving()` in its own body. The single line that arms the guard in production is unpinned. (Verdict 956 measured this as 0 of 410 on a narrower selection; the larger number here is the same finding at a wider scope.)

## Three lows

`_outcome_failure` samples usage before provenance is resolved, so every failure and park row records empty `reasoning_effort` and `skills_loaded` while success rows carry both (`headless.py:254`). Five raw claim-release `update()` sites enumerate the claim columns by hand without `owner_pid`/`owner_pid_namespace` — no live defect, since all filter `status=FAILED` which `Task.fail` already cleared via `CLAIM_FIELDS` (`transient_requeue.py:251`). And the conformance gate checks that a no-usage pragma *exists* on the call's own lines rather than that its stated reason is *true*, so it would launder the next false pragma exactly as it laundered the one just removed (`test_attempt_records_its_spend.py:36`).

That last one is worth dwelling on: the gate added to prevent this class of defect cannot distinguish a true pragma from a false one, which is the same shape as the defect it was written for.

---

Separately, and not this PR's fault: the fact that a completed pass-3 review's verdict never reached the table is the defect tracked in #4308, and this is its second instance with a merge actually reachable.

souliane:
Fourth independent cold review at `bfb99359` — **HOLD** (verdict 963). CI green: 43 check-runs read paginated by job conclusion, zero not-green, all seven required contexts success; `ruff` and `ty` clean.

## F1 is CLOSED — the design change does what it claims

Reproduced cross-process against a shared file-backed SQLite DB, **with the pre-fix signal as the control**:

```
PART A — in-memory driving() ONLY (the pre-fix signal)
  PARENT owner_is_executing=True
  CHILD  registry=[] owner_driving_since=None
  CHILD  owner_is_executing=False        <- the original defect, reproduced

PART B — the NEW drive_claim()
  PARENT owner_is_executing=True
  CHILD  owner_driving_since=2026-08-07 15:15:13Z
  CHILD  owner_is_executing=True         <- readable cross-process

deployed path, lease expired, sweeps run FROM THE CHILD:
  reclaim_orphaned_claims -> 0 rows; status=claimed
  reap_stale_claims       -> 0 rows; status=claimed
```

And the anti-vacuity control that makes it meaningful: once `drive_claim` exits, the *same child* reclaims the row (1 row → `PENDING`). The guard discriminates; it does not blanket-withhold. Moving the signal into a DB column was the right call.

Every new guard's **logic** is mutation-pinned too: entry write 1 RED, exit clear 2, tier-2 `None` check 2, pid-alive probe 4, namespace check 1, reclaim clear 1, reap clear 1, reclaim guard call 2, reap guard call 2, tier-1 registry 8.

Migrations clean: `0067` is the next free slot, neither other open PR touches migrations, `max_migration.txt` consistent, `makemigrations --check --dry-run` reports no changes.

## F2 is NOT closed — the whole fix can be deleted with a green suite

`src/teatree/agents/headless.py:157`, `with drive_claim(task):`, is the single line that arms the guard in production. Mutated to `contextlib.nullcontext()`:

- **0 RED / 14566 passed** across `teatree_agents`, `teatree_core`, `teatree_loops`, `conformance`, `integration` (25m01s)
- **0 RED / 2673** on the narrow claim/agents scope

Cause: `drive_claim` appears in exactly one test file, which exercises the function directly and never the `run_headless` call site. No test drives `run_headless` and asserts the marker is populated; no static rule names it. Every guard's logic is pinned and its **activation** is not — which is the same unpinned-activation class as F1 itself.

## New major — unbounded withholding, fail-CLOSED into a permanent wedge

Tier 2 compares `owner_driving_since` against nothing: no age, no TTL, no lease tie-in. Executed:

```
marker age 365 days | lease expired 365 days | foreign pid ALIVE
  owner_is_executing=True | reclaim=0 reap=0 | status='claimed'
control: same rows once that pid DIES -> reclaim_orphaned_claims -> 3 rows
```

The realistic production chain reproduced end to end: the drive-exit clear raises `OperationalError: database table is locked` — the exact class `describe_lease_loss` already documents on this row at this moment — so the row goes `FAILED` → `_reopen` → `PENDING` **with the marker intact** → re-claimed by the live worker → the `loops_tick` child's sweeps both return 0. Final state: `status='claimed'` with a 24h-expired lease that neither sweep will ever touch.

The sibling module this borrows its probe from already solved this: `loop_lease_liveness.UNVERIFIABLE_OWNER_GRACE = 180s`, with per-loop slots passing `trust_pid_past_ttl=False` because — its own docstring — *"a number the reader's namespace happened to reuse read as alive and would pin a dead owner's lease past its TTL forever."* This reuses the probe and drops the bound, on a signal that is deliberately never renewed.

## New major — three claim-release sites omit the new columns

`transient_requeue.py`'s `_reopen`, `_corrective_reopen` and `_self_repair_reopen` hand-enumerate claim columns and omit **all three** of `owner_pid`, `owner_pid_namespace`, `owner_driving_since`. `claim()` refreshes `owner_pid` but never clears `owner_driving_since`.

That violates the invariant `CLAIM_FIELDS`' own docstring states — *"a released claim that kept a stale `owner_pid` would report a dead owner as the executor of whoever holds the row next."* `CLAIM_FIELDS` is consumed only inside `task.py` with no drift guard. All three columns are new in this PR, so this is introduced here rather than inherited, and it is the second leg of the wedge chain above.

## Why CI could not have caught F2

None of `claim_liveness.py`, `task_claim.py`, `managers_task_sweeps.py` or `headless.py` is listed in `[tool.teatree.mutation] high_value_modules` (`pyproject.toml:551`), so `mutation-diff` and `mutation-full` give this claim-safety code **zero** coverage. That is why the M1 survivor is invisible to CI as well as to the local suite — worth closing alongside, since it is the reason the gap could exist unnoticed.

## Verified by execution vs read

**Executed:** the F1 cross-process reproduction and its control; F2's mutation at two scopes; ten per-guard mutations; the staleness-bound probe with a dead-pid control; the full production wedge chain; `makemigrations --check`; `ruff`; `ty`; the paginated CI read; the migration-slot survey. **Read only:** the spend work (not re-litigated per the brief), the `tach.toml` / dependency-graph deltas, and the tests outside the F2 scope — though `drive_claim` is grep-confirmed absent from all of them.

---

The narrow fixes: one test that drives `run_headless` and asserts the marker is populated (closing F2); a bound on tier 2 mirroring `UNVERIFIABLE_OWNER_GRACE`; and the three `transient_requeue` sites routed through `CLAIM_FIELDS` rather than hand-enumerated. Adding these modules to `high_value_modules` would stop the next such gap being invisible.

Issue #4188, opened 2026-08-04
The off preset stops the survival loops while db_backup keeps writing — a box under pressure has nothing left to relieve it


souliane:
> **Reworked 2026-08-04** after the preset model was settled. The original text cited `off` and `offline` as two presets; `offline` is being folded into `away` and the set is now five (`present`, `away`, `low-token`, `maintenance`, `off`). The finding survives the rename — the `off` half is unchanged and still live.

## The defect

The `off` preset sets the box's **survival loops** to `False`:

- `resource_pressure` — frees disk and RAM under pressure
- `idle_stack_reaper` — reaps idle stacks so provisioning does not starve
- `local_stack_queue`
- `inbox`
- `housekeeping`

…while leaving **`db_backup = True`**.

So under `off` the box keeps **writing backups** with nothing running that can reclaim space. That combination is incoherent on its own terms, independent of any naming decision: it is the one configuration that can only ever consume disk.

`offline` (being folded into `away`) had the same survival-loop set off, so folding it away removes one instance but not this one.

## Why this box specifically

This deployment has taken two out-of-memory emergencies in a single day, and `resource_pressure` / `idle_stack_reaper` are what recovered it both times. A preset that disables them removes the mechanism that has actually saved the box — at exactly the moment an operator is most likely to reach for a "stop everything" preset, during an incident. There is no recovery path from inside the box, because the loops that would recover it are the ones the preset turned off.

## The owner decision this needs

The settled model says load-bearing loops are always active, **not** dashboard-editable, admin-editable as break-glass, and *"only case where they should be disabled is when we switch to low token preset."*

`off` is defined as *"to put everything off"*. Those two statements are in tension for the survival loops specifically, and the resolution is an owner call, not an implementation detail:

- **(a)** `off` genuinely means everything, survival loops included — accepted as a deliberate "I am shutting the box down" state. Then the only fix here is `db_backup`, which must not keep writing while the reaper is off.
- **(b)** survival loops are un-quietable by *every* preset including `off`, and only `low-token` may quiet them. Then `off` means "all *work* loops off" and the box stays self-maintaining.

**(b) is the safer default** and matches the stated purpose of the load-bearing tier, but (a) is a legitimate choice if `off` is meant as a genuine halt. Either way `db_backup` should not outlive the reaper.

[...]

PR #4318, opened 2026-08-08
fix(loops): the write seam also refuses the backup-without-reclaim shape (#4188)

souliane:
fix(loops): the write seam also refuses the backup-without-reclaim shape (#4188)

refuse_quieted_load_bearing only checked the load-bearing-tier quieting
shape. A preset could still admit db_backup while the reclaim pair was
quiet through the BASE Loop.enabled columns rather than the mask -- the
exact shape a cold review reproduced via apply_entry_edits and
set_preset_entry, leaving BLUEPRINT's write-seam claim untrue. Now checks
backup_without_reclaim too, reading base_enabled from Loop.

## Why


souliane:
Closing: this PR contributes no changes. Verified with a three-dot comparison against the merge base — the diff is empty, so the branch adds nothing on top of where it was cut. The intended content already reached main through an earlier squash merge, which ancestry-based checks cannot see, so the branch still looked live.

Reopen if a diff is expected here; nothing is lost by closing an empty branch.

Issue #4202, opened 2026-08-05
Presets: seven modes still carry the pre-decision names and three dead attributes; collapse to the five decided presets


souliane:
The preset surface still carries **seven** `Mode` rows under the pre-decision names, and three of the
four attributes that were established as dead are still on the model. The decisions were taken but
never written down as work, so nothing has picked them up.

## Current state, read from the live table

```
engaged      entries=27   pauses_self_pump=False  presence_sensitive=True
heads-down   entries=27   pauses_self_pump=False  presence_sensitive=True
low-power    entries=27   pauses_self_pump=False  presence_sensitive=True
maintenance  entries=18   pauses_self_pump=False  presence_sensitive=True
off          entries=27   pauses_self_pump=False  presence_sensitive=True
offline      entries=27   pauses_self_pump=True   presence_sensitive=False
unattended   entries=17   pauses_self_pump=False  presence_sensitive=True
```

`Mode` fields today: `name`, `description`, `entries`, `defers_questions`, `pauses_self_pump`,
`presence_sensitive`, `overlay_scope`.

## Target: five presets, each a pure loop on/off table

| today | becomes |
|---|---|
| `engaged` | `present` |
| `heads-down` | **cut** |
| `low-power` | `low-token` |
| `unattended` + `offline` | `away` (merged) |
| `maintenance` | `maintenance`, redefined |
| `off` | `off` (unchanged — `off` was never the problem) |

**`maintenance` is redefined**: `tickets` / `issue_implementer` OFF (no new intake), **`ship` and
`review` ON** (finish and merge what is in flight), consolidation loops ON. Today's row has
`ship=false` / `review=false`, i.e. the current `maintenance` cannot merge anything. It doubles as
the holiday preset.

**`away` vs `maintenance` is about intake**: `away` keeps taking new work; `maintenance` drains only.

## Three dead attributes to delete

`availability_mode` was already removed (#4190). The rest:

- `pauses_self_pump` — self-pump runs during maintenance, so no preset sets it. Note it is currently
  the *only* thing distinguishing `offline`, which merges into `away`.
- `presence_sensitive` — only `offline` differs, and `offline` is going away.
- `defers_questions` — killed by the #4045 work: questions are asked immediately, none deferred.

The live data corroborates both: the two columns vary on exactly one row each, and that row is one
of the two being merged.

## Open question for whoever takes this

With `defers_questions` gone, `present` and `away` differ only by `followup`. Verify that against the
[...]

Issue #3642, opened 2026-07-23
Route interactive-session owner questions through Slack, like the headless lane already does


souliane:
Owner expectation: questions for the owner should reach them on Slack. The HEADLESS lane was fixed to do this (it may only contact the owner via the Slack egress). The INTERACTIVE lane was not: a question asked during an interactive session surfaces only in that session, so if the owner is away it is invisible until they return to that terminal.

This is the same class of defect as the intent-rot problem: an owner-directed question that nobody sees is silently lost.

## Task
Give the interactive lane the same Slack egress the headless lane uses, so an owner-directed question is delivered to Slack (and answerable from there) regardless of which lane raised it.

## Design notes
- Reuse the existing question/answer plumbing (the deferred-question record + the Slack answer loop) rather than adding a second path.
- Avoid double-delivery: a question answered in the live session should not keep pinging Slack, and vice versa.
- Keep the fast path fast: an interactive question the owner is present for should still resolve inline without waiting on a round-trip.

## Acceptance
- An owner-directed question raised in an interactive session is delivered to Slack and can be answered there.
- No duplicate delivery once answered on either surface.
- Covered by a test asserting the interactive lane reaches the same egress seam as the headless lane.

souliane:
Resolved by #3679 (merged as 498ae24b). Reopen if any part remains outstanding.

PR #4323, opened 2026-08-09
feat(modes): collapse the seven pre-decision modes to the five decided presets (#4202)

souliane:
fix(tests): pin the live-turn predicate the #4202 mode collapse orphaned

TestAttendedTurnMirrorsButDoesNotDeny was rewritten from the deleted
present-mode routing but kept a payload with no session_id, so both
_is_live_user_turn and _session_drives_loop resolved deterministically to
the loop-driven-deny arm — the opposite of what the test asserts. Pin
_is_live_user_turn the same way TestMirrorHandler already does, and give
the mocked Slack post a real string return value so the #3642 DB write
it feeds doesn't 500 and trigger a second, unexpected post.

## Why


souliane:
**Review focus for the cold pass at the new head — the prior verdict no longer binds.**

The head moved `b436de73` → `2881b1c0` with the re-fix, which voids verdict 1007 (a hold binds to a SHA). The conflict is also gone: this PR now reports `mergeable=MERGEABLE`, where it was `DIRTY`.

Whoever takes the reviewing task: establish the live head yourself and bind the verdict to the SHA you actually reviewed. Diff three-dot against the merge base using **distinct named refs** — after fetching two refs a bare `FETCH_HEAD` is ambiguous and silently reviews the wrong range.

Four things worth spending the time on, in priority order:

1. **Verify the operator-intent inversion is discharged by EXECUTION, not by reading the diff.** The blocker in verdict 1007 was that migration 0069 repointed the stored `offline` mode — the holiday hold, every work loop OFF — onto `away`, which *takes* new intake. It was proven by running the migration: a seeded `--hold` holiday override resolved post-migration to `away` with tickets/ship/dispatch all true. Re-run that, don't accept the claim.

2. **Enumerate the class, not the named line.** A recurring pattern on this repo is a fix that closes the one site it was handed and leaves a sibling of the same class in the same module. Having checked migration 0069, look for the other places this mode-collapse mapping is expressed — another migration, a fixture, a default, a doc table — and say whether they agree with each other. Four sources in this PR previously named *different* successors, and only the migration said `away`.

3. **The two GREEN mutants from the prior pass.** Re-derive them. A surviving mutant means the pin does not constrain the behaviour it claims to, and that is exactly the dimension this PR is about.

4. **Run each added test against `main`.** A test that passes on `main` is not pinning this change. Report the RED-without / GREEN-with counts rather than asserting coverage.

A clean verdict is a valid and useful outcome — say so plainly if the work is sound, because this PR is currently the closest thing on the board to a merge.

Posted as a comment rather than as a second reviewing task: the reviewer ticket already carries one (the author-row duplicate I opened by mistake has been cancelled), and two reviewing tasks on one PR produce competing verdicts.

Issue #4216, opened 2026-08-05
The TaskCreated skill-loading gate fires on local task-list entries too, and denies with an empty reason — task tracking dies silently


souliane:
The `TaskCreated` skill-loading gate (#1488, `hooks/scripts/hook_router.py:1266`) exists to close a
real loophole: a fanned-out sub-agent starts blank, cannot inherit the parent's loaded skills, and has
no `Skill` tool — so the dispatch prompt itself has to name the skills. That reasoning is sound for a
fan-out.

But `TaskCreated` does not only fire for a fan-out. It also fires when a session adds an entry to its
own **local task list**, where there is no sub-agent, no dispatch prompt, and nothing that could ever
load a skill. The gate cannot distinguish the two, so it applies the sub-agent demand to a plain todo
entry and denies it.

## Reproduction

A session whose pending demand set is non-empty adds a local task. The gate denies:

```
{"continue": false, "stopReason": "SKILL LOADING ENFORCEMENT (TaskCreated): this fanned-out
sub-agent starts BLANK and will not load teatree skills unless its dispatch prompt tells it to.
The prompt does not reference these required skills: /internals. ..."}
exit 2
```

The demand is **unsatisfiable by construction** on this path. The remedy the message gives — put a
`Read .../skills/internals/SKILL.md` line in the task description — is meaningless for a todo entry:
nothing reads that description as a prompt, so complying changes nothing except padding the task text
with an instruction addressed to a sub-agent that does not exist.

## Why it stayed invisible

The gate's `stopReason` carries the full explanation and the two documented escapes. **The caller
never sees it.** What surfaces at the call site is:

```
TaskCreated hook feedback: [.../hook_router.py --event TaskCreated]: No stderr output
```

An empty failure. The gate is doing exactly what it was written to do, is saying so clearly, and the
message is dropped somewhere between `stopReason` and the caller — leaving a block with no reason, no
remedy, and nothing to distinguish "the gate denied you, here is how to comply" from "the hook
crashed".

That is this repo's own defect class: an unreadable signal presented as an opaque failure. It is what
turned a debatable over-fire into a silent one — the deny fired repeatedly across a long session,
task tracking simply stopped working, and there was no way to tell why without running the hook by
hand against a synthetic payload.

## Impact

Task tracking degrades silently for any session with a non-empty pending demand set. Because the
denial is empty, the natural read is that the task tool is flaky rather than that a gate is refusing —
[...]

Issue #1488, opened 2026-05-29
fix(hooks): close the ultracode Task/Workflow fan-out skill-loading bypass via a TaskCreated gate


adrien-oper:
## Problem

`ultracode` (and any harness Workflow/Task fan-out) spawns sub-agents through the Task/Workflow vehicle, which **bypasses `PreToolUse` hooks** (a known regression from TodoWrite — see `docs/claude-code-internals.md` §9). TeaTree's skill-loading enforcement gate (`handle_enforce_skill_loading`) fires only on the `PreToolUse` matcher `Bash|Edit|Write`, so it is never consulted on the fan-out. Sub-agents therefore skip auto-loading the matching teatree lifecycle skill.

This is the loophole that let a **bespoke review workflow** run instead of `/t3:review` on a fanned-out review task.

## Fix

Wire a `TaskCreated` handler that forces the matching teatree lifecycle skill + its (already-transitive) companions onto the fan-out vehicle — enforcing **skill-loading only**, never capping agent count / token budget / workflow size (ultracode keeps maximal room).

- New `handle_enforce_skill_loading_on_task_create` reading the `TaskCreated` stdin schema (`task_id`, `task_subject`, `task_description`).
- Demand set = (pending − loaded) ∪ companions of `lifecycle_for_task_text(task_description)`, each dropped if unresolvable (fail-open on stale/renamed skills).
- `TaskCreated` deny schema = `{"continue": false, "stopReason": <reason>}` + exit 2.
- Fail-open guarantees: `[teatree] skill_loading_gate_enabled` kill-switch, per-call `[skip-skill-gate: <reason>]` token, `t3 <overlay> gate` self-rescue extended to skill-loading.
- `TEATREE_PLAN_GATE_WINDOW_MINUTES = 0` sentinel disables the plan-gate freshness window for big multi-wave fan-outs.

Empirically verified against the Claude Code 2.1.156 binary: `TaskCreated`/`TaskCompleted` events fire with that exact schema; `continue: false` sets `preventContinuation`.

souliane:
Closing — verified implemented (triage + code-verification 2026-06-07): handle_enforce_skill_loading_on_task_create wired on TaskCreated + kill-switch + [skip-skill-gate]; corpus test. Reopen if a gap remains.

PR #4332, opened 2026-08-10
fix(hooks): scope the TaskCreated skill-loading gate to a real dispatch, and never deny mutely (#4216)

souliane:
fix(hooks): scope the TaskCreated skill-loading gate to a real dispatch, and never deny mutely (#4216)

- `TaskCreated` also fires when a session adds an entry to its OWN task list —
  no sub-agent, no dispatch prompt, nothing that reads the description as one.
  The demand was unsatisfiable by construction there, so task tracking died.
  `subagent_skill_gate.is_subagent_dispatch` keys on the only fields that
  positively identify a fan-out (`teammate_name` / `team_name`); a payload
  carrying neither is out of scope. A fan-out missing its skills still denies.
- The deny reason never reached the caller. The harness's task-creation
  consumer reads only the `blockingError` its runner derives; `continue: false`
  becomes `preventContinuation`, which that consumer ignores — so an exit-2 deny
  carrying only the stop envelope fell through to the runner's fallback and
  surfaced as `[<command>]: No stderr output`. `decision: "block"` is the one
  stdout key the runner turns into a `blockingError` carrying our own text, so
  the deny now emits both.
- New sibling `hooks/scripts/task_created_deny.py` owns the deny surface for all
  three `TaskCreated` gates (the router is shrink-only; it nets -16 LOC).
  `harness_surfaced_deny_text` models the harness read and the emitter enforces
  it: a reason that would not reach the caller fails OPEN and logs.
- `TaskCompleted` shares the schema but teatree registers no handler on it, so
  the mismatch cannot reach it — pinned, along with "every `TaskCreated` handler
  denies through this emitter".

Regression pins observed RED before the fix: a local task-list entry denied
(3 tests), and the emitted payload surfacing no text (3 tests).

Open questions & assumptions:
- assumed: `teammate_name`/`team_name` are the discriminator. Read off the
  installed Claude Code binary — the schema marks both optional, and the sole
  `TaskCreated` emitter fills them from the session's teammate context, which is
  absent for a main session's own task list. No other payload field separates
  the two (`task_description` is present for both).
- assumed: the fail-open note may go to stderr. The binary's own per-event
  contract is "exit 0 — stdout/stderr not shown", and the hook runner derives a
  `blockingError` from stderr only at exit 2; `hooks/CLAUDE.md`'s broader "aborts
  on ANY handler stderr" claim is left untouched, and the deny path stays silent.
- open: the sibling `TaskCreated` gates (#171 dispatch-quote, #4107 admission)
  over-fire on a local task-list entry the same way — the admission arm books an
  `InteractiveDispatch` seat for a todo entry. Out of scope here; they share the
  fixed deny surface but not the discriminator.
- open: if the harness fan-out stops carrying the teammate fields, this gate
  goes quiet rather than wrong. The `PreToolUse` `Agent` matcher is where a
  modern Agent-tool dispatch would be governed instead.

## Why


souliane:
**Live reproduction of this PR's premise, hit by accident just now — plus a residual the current diff does not appear to cover.**

I tried to add a row to my own harness todo list. Not a dispatch: no sub-agent, no fan-out, no `Agent`/`Workflow` call. Just a local task-list write.

It was denied. Twice.

```
TaskCreate(subject="React to a pushed fix and drive it to merge via the keystone", ...)
→ TaskCreated hook feedback: [.../hook_router.py --event TaskCreated]: No stderr output
```

Adding the gate's own escape token to the description made the identical call succeed:

```
[skip-skill-gate: local harness todo write, not a sub-agent dispatch]
→ Task #37 created successfully
```

So the `enforce-skill-loading-on-task-create` gate (#1488) is what fired, on a write that spawns nothing. That is this PR's premise, reproduced from the operator side rather than from a test.

## The residual: the deny reason never arrives

This is the part worth adding to the plan. `emit_task_create_deny` writes its reason as `{"continue": false, "stopReason": reason}` on **stdout**. What the caller actually received, both times, was:

```
No stderr output
```

The reason was composed and then lost. From where I sat the block was indistinguishable from a crashed hook — no reason, no named remedy, no mention that `[skip-skill-gate: …]` exists. I only found the cause by reading `hook_router.py:1224-1300` directly.

That matters more than the misfire itself. A gate that denies with a reason teaches; a gate that denies silently just looks broken, and the natural next move for anyone who does not read the router source is to conclude the hook is failing and work around it. Note the asymmetry with the sibling gates: the `PreToolUse` denies surface their reason fine, so nothing about the operator's experience suggests "there is a documented escape token you are missing."

This is the same shape the repo already tracks as fail-silent: a refusal whose payload is written but never reaches the consumer, rendering "denied for reason X" as "something went wrong."

## Suggested additions to the replan

- Whatever survives of this gate emits its reason on the channel the harness actually surfaces, and names `[skip-skill-gate: <reason>]` in that text. A test that FAILS on current code: trigger the deny and assert the caller receives the reason string, not an empty message.
- A local task-list write is not a dispatch and is out of scope for the gate by construction, rather than by a caller remembering the token. The token should be the escape for a genuine fan-out that legitimately needs it, not the routine cost of using the todo list.
- Worth checking whether the same stdout-vs-stderr mismatch affects the `TaskCompleted` arm and the dispatch-quote gate at `hook_router.py:2192`, since they share the envelope style. If they do, that is one fix rather than three.

Recorded here rather than acted on — this session routes, it does not implement.

